PR #17, opened 2015-02-19
Msproteomicstools

bgruening:


packages/package_searchgui_1_26/tool_dependencies.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<tool_dependency>
+    <package name="searchgui" version="1.26">
+        <install version="1.0">
+            <actions>
+                <action type="download_by_url" target_filename="SearchGUI-1.26.2.tar.gz">http://genesis.ugent.be/maven2/eu/isas/searchgui/SearchGUI/1.26.2/SearchGUI-1.26.2-mac_and_linux.tar.gz </action>
+                <action type="shell_command">chmod -R o+w resources</action>
```
trevor: Does this create a word-writable directory? If so these can be problematic from a security standpoint. Is there another way of getting the desired end result?

bgruening: Yes it will give `others` write access. I think this is a left over from the original installation instruction. The user who is running this installation should also execute the jobs ... we should remove this line and see if it still works if you think this can be problematic. In theory this should work :)

trevor: Sounds good to me, let's remove it and the similar line in package_peptideshaker_0_37

bgruening: Fixed: https://github.com/galaxyproteomics/tools-galaxyp/commit/cdc2de7d061c7c822755d00eb84f050a1ad4ede4

trevor: :+1:

jmchilton: :+1: Shouldn't be needed with the tool shed - only Galaxy will ever run tool shed applications. Probably would still be needed in a multiple user setting unless they have fixed this limitation upstream in SearchGUI and/or PeptideShaker.

PR #18, opened 2015-02-23
update maxquant wrapper

MateusBellomo:
Add option to select protein database like Andromeda using it's regular expressions.
Change option to get maxquant modifications (use from_data_table)
Add option to select several raw inputs.
With the new galaxy configuration, we have to copy 'maxquant_mods.loc' and 'maxquant_proteases.loc' files to tool-data folder and include in 'the tool_data_table_conf.xml' this lines:

    <table name="proteases" comment_char="#">
      <columns>value</columns>
      <file path="tool-data/maxquant_proteases.loc" />
    </table>
    <table name="maxquant_mods" comment_char="#">
      <columns>value</columns>
      <file path="tool-data/maxquant_mods.loc" />
    </table>

### DIFF
--- a/tools/maxquant/maxquant.xml
+++ b/tools/maxquant/maxquant.xml
@@ -311,11 +311,11 @@
     <expand macro="misc_conditional" />
   </inputs>
   <outputs>
-    <data format="csv" name="output_protein_groups" label="MaxQuant Protein Groups for ${on_string}"/>
-    <data format="csv" name="output_peptides" label="MaxQuant Peptides for ${on_string}"/>
-    <data format="csv" name="output_evidence" label="MaxQuant Evidence for ${on_string}"/>
-    <data format="csv" name="output_parameters" label="MaxQuant Tabular Parameters for ${on_string}"/>
-    <data format="csv" name="output_msms" label="MaxQuant MSMS for ${on_string}"/>
+    <data format="tabular" name="output_protein_groups" label="MaxQuant Protein Groups for ${on_string}"/>
+    <data format="tabular" name="output_peptides" label="MaxQuant Peptides for ${on_string}"/>
+    <data format="tabular" name="output_evidence" label="MaxQuant Evidence for ${on_string}"/>
+    <data format="tabular" name="output_parameters" label="MaxQuant Tabular Parameters for ${on_string}"/>
+    <data format="tabular" name="output_msms" label="MaxQuant MSMS for ${on_string}"/>
     <data format="tabular" name="output_mqpar" label="MaxQuant Parameters XML for ${on_string}"/>
   </outputs>
   <help>
--- a/tools/maxquant/maxquant.xml
+++ b/tools/maxquant/maxquant.xml
@@ -1,59 +1,38 @@
 <tool id="maxquant" version="0.1.0" name="MaxQuant">
-  <description>
-  </description>
+  <description></description>
   <requirements>
     <requirement>maxquant</requirement>
     <requirement type="platform">windows</requirement>
   </requirements>
   <configfiles>
     <configfile name="inputs_config">##Describe inputs
-#set $fraction = 0
-#set $type = str($analysis_type.type)
-#if $type.startswith("sg" )
-#set $groups = [$analysis_type]
-#elif $type.startswith("mgs")
-#set $groups = $analysis_type.groups
-#end if
-#for $i, $group in enumerate($groups)
-num:${str(i + 1)}
-#if $type.endswith("samples")
-#for $j, $input in enumerate($group.inputs)
-name:${input.display_name}
-path:${input}
-sample:$j
-#end for
-#else
-#if $type.endswith("fractions")
-#set $samples = [$group]
-#else
-#set $samples = $group.samples
-#end if
-#for $j, $sample in enumerate($samples)
-#for $k, $input in enumerate($sample.inputs)
-name:${input.display_name}
-path:${input}
-sample:$j
-#end for
-#end for
-#end if
-#end for
-</configfile>
+    
+    #for $s in $sample
+       ${s.inputs}
+       ${s.inputs.display_name}
+       ${s.fraction_of_sample}
+       ${s.class_of_sample}
+       ${s.group}
+    #end for
+
+    </configfile>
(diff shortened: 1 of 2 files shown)


tools/maxquant/maxquant_wrapper.py
```diff
@@ -589,63 +587,61 @@ def setup_database(options):
...
+        database_regex = '>([^\|]*)'
+
+
     assert database_destination == os.path.basename(database_destination)
     symlink(database_path, database_destination)
 
     database_conf = get_env_property("MAXQUANT_DATABASE_CONF", None)
     if not database_conf:
-        exe_path = which("MaxQuantCmd.exe")
+        exe_path = "C:\MaxQuant\\bin\MaxQuantCmd.exe"
```
jmchilton: This is not ideal - you should make sure ``MaxQuantCmd.exe`` is on LWR/Pulsar's path instead of changing the tool. If this is the most common place MaxQuant is installed to though - it may make sense to fallback to this default if which cannot determine which MaxQuant to use.

tools/maxquant/maxquant.xml
```diff
@@ -309,51 +285,22 @@ sample:$j
...
-        <expand macro="group_params" />
-      </when>
-      <when value="sg_both">
-        <repeat name="samples" title="Sample">
-          <expand macro="input_param" />
-        </repeat>
-        <expand macro="group_params" />
-      </when>
-    </conditional>
+    <repeat name="sample" title="Sample">
```
jmchilton: So you have eliminated all of the different analysis type options except `Single Group - Multiple Fractions and Samples`. This makes me nervous, I feel like those other options were there for reasons and might be useful in various scenarios. Are you certain the other scenarios aren't useful?

MateusBellomo: I thought the way someone describes the Experiemental Design determines the type of analysis. Isn't work that way?

tools/maxquant/maxquant_wrapper.py
MateusBellomo: I'll work on that.

tools/maxquant/maxquant.xml
MateusBellomo: John, I talked with people in our group and we think that is very rare to someone use more than one group in the analysis. Because of that, I'll come back to do this later if it's not a problem to you. And I'd like to know if you are ok the way we redesign the select of input data.

bgruening: Mateus;
Why removing these options if they are working, even if the use-case is very rare?

tools/maxquant/maxquant.xml
```diff
@@ -364,11 +311,11 @@ sample:$j
     <expand macro="misc_conditional" />
   </inputs>
   <outputs>
-    <data format="tabular" name="output_protein_groups" label="MaxQuant Protein Groups for ${on_string}"/>
-    <data format="tabular" name="output_peptides" label="MaxQuant Peptides for ${on_string}"/>
-    <data format="tabular" name="output_evidence" label="MaxQuant Evidence for ${on_string}"/>
-    <data format="tabular" name="output_parameters" label="MaxQuant Tabular Parameters for ${on_string}"/>
-    <data format="tabular" name="output_msms" label="MaxQuant MSMS for ${on_string}"/>
+    <data format="csv" name="output_protein_groups" label="MaxQuant Protein Groups for ${on_string}"/>
```
bgruening: Is a comma separated files needed for downstream analysis? If not can we produce tabular files as output by simply piping all outputs through `sed`? Keeping everything `tab` separated as a big advantage in Galaxy. You get a ton of tools for free.

tools/maxquant/maxquant_wrapper.py
```diff
@@ -10,6 +10,9 @@
 from xml.sax.saxutils import escape
 import xml.etree.ElementTree as ET
 
+import pprint
+import re
```
bgruening: pprint is not used in this document

tools/maxquant/maxquant.xml
jmchilton: No `sed` on Windows :).

tools/maxquant/maxquant.xml
jmchilton: If no one is using these features and they are hard to test - I guess I would prefer to stick with @MateusBellomo advice here. I don't think MSI is actively using these wrappers anymore (@trevor ?) so I think them getting maintained and upgraded is more important than preserving features past-@jmchilton speculatively thought he might need but didn't understand.

tools/maxquant/maxquant.xml
bgruening: I found this one: 
```
SED-replacement with PowerShell
cat somefile.txt | %{$_ -replace "expression","replace"}
```
As far as I know maxquant can run under wine, isn't it? If so we need to support both worlds or advertise only one.

tools/maxquant/maxquant.xml
bgruening: Yes this is a valid point, if there are hard to test we can remove them, otherwise I would keep them.

bgruening:
Besides my small comments, I like those changes. Looks like you could remove a lot complexity. Thanks!

tools/maxquant/maxquant.xml
trevor: @jmchilton correct, not installed at MSI

bgruening:
Btw. you can also ship a `tool_data_table_conf.xml.sample` file with this repository, so you don't need to edit the `tool_data_table_conf.xml` file manually. See for example here:  https://github.com/bgruening/galaxytools/tree/master/tools/diamond

tools/maxquant/maxquant.xml
MateusBellomo: In fact I did not remove completely the possibilitie of selecting several parameter groups. To do that (and maintaining this interface of select raw data that we changed) I'd have to change the xml so the user could select the parameters for every group, and I'd also have to change the wrapper to take this different parameters for each group. But I am planning to do that later, meanwhile we want to let the usual features working properly.
Sorry but I didn't understand what you mean about MSI. What is it?

PR #24, opened 2015-03-04
Add support for converting the Wiff datatype to the msconvert tool.

jj-umn:
Did the minimal to add conversion of wiff datasets.   
This relies on the Wiff datatype having files named wiff and wiff_scan in the extra_files_path, 
and requires the input_names to have the .wiff and .wiff.scan extensions.
The option --implicit was added to  msconvert_wrapper.py in order to include the .wiff.scan file in the correct directory/filename but not to add it to the mscovert application command line.

### DIFF
--- a/tools/msconvert/msconvert.xml
+++ b/tools/msconvert/msconvert.xml
@@ -8,11 +8,24 @@
   </requirements>
 -->
   <command interpreter="python">
+    #import re
     #set $ext = $input.ext
     msconvert_wrapper.py
-    --input=${input}
-    #if hasattr($input, 'display_name')
-    --input_name='${input.display_name}'
+    #if $type.input_type == 'wiff':
+      #set basename = 'absciex'
+      #if hasattr($input, 'display_name')
+        #set basename = $re.sub('\W','_',$input.display_name)
+      #end if
+      --input=${input.extra_files_path}/wiff
+      --input_name='${basename}.wiff
+      --implicit=${input.extra_files_path}/wiff_scan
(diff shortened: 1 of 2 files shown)


tools/msconvert/msconvert.xml
```diff
@@ -371,7 +388,7 @@ You can view the original documentation here_.
 
 For the underlying tool, please cite `ProteoWizard: Open Source Software for Rapid Proteomics Tools Development. Darren Kessner; Matt Chambers; Robert Burke; David Agus; Parag Mallick. Bioinformatics 2008; doi: 10.1093/bioinformatics/btn323.`
 
-If you use this tool in Galaxy, please cite Chilton J, et al. https://github.com/galaxyproteomics/
+If you use this tool in Galaxy, please cite Chilton J, et al. https://bitbucket.org/galaxyp/msconvert
```
bgruening: The old one was correct I guess.

tools/msconvert/msconvert_wrapper.py
```diff
@@ -249,6 +250,8 @@ def run_script():
             print("4- input_file: %s" % input_file)
         input_file = input_file
         copy_to_working_directory(input, input_file)
+        if input in options.implicits:
+            continue
```
bgruening: Imho, the following is more readable:
```
if input not in options.implicit:
    input_files.append(input_file)
```

jj-umn:
Made changes as per bgruening suggestions.  Thanks.

bgruening:
:+1:

PR #25, opened 2015-03-10
Use multiple temp directories

bgruening:
This should be a safe-guard to not confuse peptide-shaker during different instances.


jj-umn:
+1  Seems like a good idea.

PR #30, opened 2015-04-13
add initial version of spectrast2spectrast_irt wrapper

bgruening:
This wrapper is work in progress. 

Does anyone have a small test input file?
@iracooke are these datatypes already included in `proeomics-datatypes`?

### DIFF
--- a/tools/spectrast2spectrast_irt/.shed.yml
+++ b/tools/spectrast2spectrast_irt/.shed.yml
@@ -1,0 +1,3 @@
+# repository published to https://toolshed.g2.bx.psu.edu/repos/galaxyp/spectrast2spectrast_irt
+owner: galaxyp
+name: spectrast2spectrast_irt
--- a/tools/spectrast2spectrast_irt/spectrast2spectrast_irt.xml
+++ b/tools/spectrast2spectrast_irt/spectrast2spectrast_irt.xml
@@ -1,0 +1,74 @@
+<tool id="gp_spectrast2spectrast_irt" name="SpectraST RT Normalizer" version="0.1.0">
+    <description></description>
+    <requirements>
+        <requirement type="package" version="0.3.2">msproteomicstools</requirement>
+    </requirements>
+    <command>
+<![CDATA[
+        spectrast2spectrast_irt.py
+            --in ${input}
+            --out ./result_file.txt
+            --kit #','.join($kit)#
+            $applychauvenet
+            $precursorlevel
+            $spectralevel
+            $report
+            ##--exclude $exclude
+            ##--surrogate $surrogate
+            ##--linearregression $linearregression
+            --rsq_threshold $rsq_threshold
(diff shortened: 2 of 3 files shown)


bgruening:
Hi Ira,

I would be in favor of using a different tool to replace the functionality, the code of `spectrast2spectrast` is not the best.

iracooke:
OK .. sounds like maybe a protk wrapper?  Naturally that would be my preference ;)

bgruening:
@iracooke should I work on this or do you want to integrate it in protk?

iracooke:
I don't think this needs to be in protk.  I think there is a reasonable separation here, since the protk spectrast tools work with the spectrast binary (and depend on the TPP), whereas this is a converter specifically required for OpenSWATH type workflows right?  I think it makes sense for it to be in its own separate package.

bgruening:
@iracooke do you have a test input file for `spectrast2spectrast_irt`?
https://github.com/msproteomicstools/msproteomicstools/blob/master/test/data/testlib.sptxt results in `Exception("Error: At least one of your runs doesn't contain any peptides from the Biognosys RT-kit!")
`

bgruening:
@iracooke this wrapper should work now, but is missing a test case to verify.

iracooke:
No but  Sarah or Pratik may have one though.

iracooke:
Yeah - I'll send a file in a bit. Out away from my laptop right now.

bgruening:
@iracooke Sarah gave me some test data and I could test this tool. It seems to work. Unfortunately, this is to big to be added so I will not include it in this PR. I think this is ready be to merged.

Do we need to add a splib-text datatype?

iracooke:
@bgruening I suggest depending on the proteomics-datatypes package. I've added an splib datatype to that package which should be what you need.  We can add an splibbin datatype later if needed but it seems that all Sarah's needs are related to the text formatted splib data and this is what my splib datatype represents.

bgruening:
I thought your SPLIB datatype is a composite datatype and needs to have 3 files. As far as one file was enough in this case.
I will add the proteomics datatype as repository dependency. Thanks Ira!

iracooke:
OK.  I see what you are saying.  For convenience maybe it makes sense to also have a non-composite datatype when the indexes are not needed.  Maybe call it "splibnoindex" rather than splib-text because even the indexed files are actually in ASCII

bgruening:
@iracooke sounds like a good idea. Do you want to me just subclass `Text` or a proper datatype with sniffer? The sniffer will make some trouble in both datatypes I assume. Because your current sniffing function will only recognise the one text file, not the index. To be honest, I don't know offhand how to sniff 2 files (composite files) at the same time.

bgruening:
@iracooke I will merge this so we can start testing this wrapper. I think we can add this datatype later, currently text will work.

PR #34, opened 2015-05-23
Update the package for msproteomicstools

bgruening:


jj-umn:
Does packages/package_pyprophet_0_13_2/tool_dependencies.xml
 need to have the set_environment for atlas similar to package_msproteomicstools_0_3_2/tool_dependencies.xml?

bgruening:
It will not hurt to include it, but I improved the numpy packages to export the atlas libs/includes as well. So importing numpy will result in an imported atlas, hopefully.

As soon as all this get's merged I can test with a clean docker container.

PR #45, opened 2015-06-04
Add LC_ALL setting to tool_dependencies for searchgui

iracooke:
@bgruening Can you check this to see if it looks right?


bgruening:
Interesting. It's long time ago were I used this syntax, but it looks correct.

What about adding this line to the searchgui package?

```
 <environment_variable name="LC_ALL" action="set_to">C</environment_variable>
```

iracooke:
Do you mean just add that line and then we would not need anything in tool_dependencies.xml?

That would be nice.  I'll try it later. I need to go to some talks now

bgruening:
Yes, I guess this could work! Enjoy!

iracooke:
I'm merging this now as it seems to work on my test

PR #48, opened 2015-08-07
Use temporary per-job copy of peptideshaker and searchgui packages

iracooke:
This is a fix for cluster environments suggested by Harald and Marc

### DIFF
--- a/tools/peptideshaker/peptide_shaker.xml
+++ b/tools/peptideshaker/peptide_shaker.xml
@@ -16,14 +16,18 @@
         #set $samp_str = "Sample_%s" % datetime.now().strftime("%Y%m%d%H%M%s")
         #set $temp_stderr = "peptideshaker_stderr"
 
+
         mkdir output_reports;
         cwd=`pwd`;
+
+        cp -r "\${PEPTIDESHAKER_JAR_PATH%/*}" bin;
+        tmp_peptideshaker_jar_path=`echo "\$cwd/bin/\${PEPTIDESHAKER_JAR_PATH\#\#/*/}"`;
 
         ln -s '$searchgui_input' searchgui_input.zip &&
         ######################
         ## PeptideShakerCLI ##
         ######################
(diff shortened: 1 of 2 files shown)


iracooke:
@bgruening Can you take a look and merge if you think its OK

tools/peptideshaker/peptide_shaker.xml
```diff
@@ -16,14 +16,18 @@
         #set $samp_str = "Sample_%s" % datetime.now().strftime("%Y%m%d%H%M%s")
         #set $temp_stderr = "peptideshaker_stderr"
 
+
         mkdir output_reports;
         cwd=`pwd`;
```
bgruening: I would rather use Python magic than Bash magic. Simply because I think this is more readable.
```
#import os
#set $PEPTIDE_SHAKER_ROOT_PATH = os.path.dirname( os.environ['PEPTIDESHAKER_JAR_PATH'] )
#set $JAR_NAME = os.path.basename(os.environ['PEPTIDESHAKER_JAR_PATH'])
#set $tmp_peptideshaker_jar_path = os.path.join( $PEPTIDE_SHAKER_ROOT_PATH,  $JAR_NAME)
```

```java -cp $tmp_peptideshaker_jar_path```

bgruening:
I think this looks good! I'm not sure about the bash magic and I would prefer Python code, but if this works let's get this in :)
Thanks @iracooke

iracooke:
Ha ha ... bash is beautiful :)  Is there a PEPTIDESHAKER_ROOT_PATH variable defined though?  
I will merge but feel free to replace with python code if you prefer.

bgruening:
`$PEPTIDESHAKER_ROOT_PATH` was defined in line 1, with the `#set`.
Can you upload this to TTS and TS? I will test on our server if I find time. Thanks!

iracooke:
Oops. Missed that. 

I have already uploaded it to the TS. Do you need TTS as well?

Cheers
Ira

PR #49, opened 2015-08-07
overwride $HOME dir to the cwd

bgruening:
This is needed because msamanda wants to write to `~/.msamanda`
ping @iracooke


bgruening:
Unfortunately, this does not work for /usr/local/galaxy/.compomics/ptmFactory-3.28.24.cus but for the others.
@iracooke do you know what .compomics is?

PR #50, opened 2015-09-09
Bumbershoot updates

chambm:
I renovated the Bumbershoot tools (MyriMatch/idpQonvert) in galaxyp and added the rest of IDPicker tools as well (idpAssemble, idpQuery, idpQonvertEmbedder). I had to change ownership of the packages to test them, but I'm open to changing the ownership back to galaxyp if I get commit access to that on the toolshed. I combined all the bumbershoot tools into a subdirectory in the tools directory and flattened the directory structure (e.g. bumbershoot/myrimatch/myrimatch.xml instead of myrimatch/tools/myrimatch.xml). Each tool still has a test-data folder, but most tools use test data from other tools, e.g. idpAssemble uses out from idpQonvert as its input.


packages/package_bumbershoot_3_0_8738/.shed.yml
```diff
@@ -0,0 +1,10 @@
+categories:
+- Tool Dependency Packages
+description: Contains a tool dependency definition that downloads the Bumbershoot tools.
+long_description: |
+  Bumbershoot
+
+  https://sourceforge.net/p/proteowizard/code/HEAD/tree/trunk/pwiz/pwiz_tools/Bumbershoot/
+name: package_bumbershoot_3_0_8738
```
bgruening: remote_repository_url: https://github.com/galaxyproteomics/tools-galaxyp/packages/package_bumbershoot_3_0_8738/

is missing.

packages/package_bumbershoot_3_0_8738/.shed.yml
```diff
@@ -0,0 +1,10 @@
...
+owner: chambm
```
bgruening: `galaxyp`

packages/package_bumbershoot_3_0_8738/tool_dependencies.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0"?>
+<tool_dependency>
+    <package name="bumbershoot" version="3.0.8738">
+        <install version="1.0">
+            <actions>
+                <action type="download_by_url">https://teamcity.labkey.org/guestAuth/repository/download/ProteoWizard_Bumbershoot_Linux_X86_64/265091:id/bumbershoot-bin-linux-gcc48-release-3_0_8738.tar.bz2</action>
```
bgruening: This link does not look very stable. If you like you can put this tarball into https://github.com/bgruening/download_store.
We try to make these tarballs as sustainable as possible.

packages/package_bumbershoot_3_0_8789/.shed.yml
```diff
@@ -0,0 +1,10 @@
+categories:
+- Tool Dependency Packages
+description: Contains a tool dependency definition that downloads the Bumbershoot tools.
+long_description: |
+  Bumbershoot
+
+  https://sourceforge.net/p/proteowizard/code/HEAD/tree/trunk/pwiz/pwiz_tools/Bumbershoot/
+name: package_bumbershoot_3_0_8789
+owner: chambm
```
bgruening: As above ...

tools/bumbershoot/idpassemble/.shed.yml
```diff
@@ -0,0 +1,6 @@
+# repository published to https://toolshed.g2.bx.psu.edu/repos/galaxyp/idpqonvert
+categories:
+- Proteomics
```
bgruening: Can you please add a long_descriptions and shorten the description. Also please add repository_url.

tools/bumbershoot/idpassemble/idpassemble.xml
```diff
@@ -0,0 +1,122 @@
...
+            -SummarizeSources 1
+            #if len($input) > 1
+                -MergedOutputFilepath output
+                #for $i in $input
+                    '${i.file_name}'
+                #end for
+            #else
+                '${input}'
+                ;
+                cp '${input}' output
```
bgruening: `mv` is usually better for the performance.

tools/bumbershoot/idpassemble/idpassemble.xml
```diff
@@ -0,0 +1,122 @@
...
+            #end if
+    </command>
+    <inputs>
+        <param name="input" type="data" format="idpdb" label="Input File(s)" multiple="true"/>
+        <param name="max_fdr_score" type="float" label="Max FDR Score" value="0.05" />
+        <conditional name="filter_at_gene_level_condition">
+            <param name="filter_at_gene_level" type="boolean" truevalue="1" falsevalue="0" label="Filter at Gene Level" help="Apply filters at the gene level (i.e. 'min distinct peptides per gene group' instead of 'min distinct peptides per protein group')"/>
```
bgruening: As of https://github.com/galaxy-iuc/standards/blob/master/docs/best_practices/tool_xml.rst#booleans
`booleans` are not recommended for conditionals. Would be great if you can change this to a select.

tools/bumbershoot/idpassemble/idpassemble.xml
```diff
@@ -0,0 +1,122 @@
...
+                #end for
+            #else
+                '${input}'
+                ;
+                cp '${input}' output
+            #end if
+    </command>
+    <inputs>
+        <param name="input" type="data" format="idpdb" label="Input File(s)" multiple="true"/>
+        <param name="max_fdr_score" type="float" label="Max FDR Score" value="0.05" />
```
bgruening: I would add the original parameter (-MaxFDRScore) to the help of all `params`. This will enable our users to map the Galaxy parameter to the command line one easily.

tools/bumbershoot/idpassemble/idpassemble.xml
```diff
@@ -0,0 +1,122 @@
...
+        <test>
+            <param name="input" value="../../idpqonvert/test-data/201208-378803-mm.idpDB" />
+            <param name="max_fdr_score" value="0.05" />
+            <param name="filter_at_gene_level_condition.min_distinct_peptides" value="2" />>
+            <param name="filter_at_gene_level_condition.min_spectra" value="2" />
+            <param name="filter_at_gene_level_condition.min_additional_peptides" value="1" />
+            <param name="min_spectra_per_distinct_match" value="1" />
+            <param name="min_spectra_per_distinct_peptide" value="1" />
+            <param name="max_protein_groups_per_peptide" value="10" />
+            <output name="output" file="../../idpassemble/test-data/201208-378803-mm.idpDB" compare="sim_size" delta_size="500000" />
```
bgruening: This will not work in the TS and with running tests with `--installed` isn't it?
Github understands symlinks. So what we usually do is symlinking the files into the test-data repositories and not specifying `../../`. So everything is self-contained.

tools/bumbershoot/idpassemble/idpassemble.xml
```diff
@@ -0,0 +1,122 @@
...
+<![CDATA[
+**What it does**
+
+Merges and filters one or more IDPicker 3 idpDB files into a combined idpDB file. Protein assembly (e.g. parsimony) is conducted on the combined set of proteins.
+]]>
+    </help>
+    <citations>
+        <citation type="doi">10.1021/pr900360j</citation>
+        <citation type="bibtex">@misc{toolsGalaxyP, author = {Chilton, J, Chambers MC, et al.}, title = {Galaxy Proteomics Tools}, publisher = {GitHub}, journal = {GitHub repository},
+                                      year = {2015}, url = {https://github.com/chambm/tools-galaxyp}}</citation> <!-- TODO: fix substitution of commit ", commit = {$sha1$}" -->
```
bgruening: obsolete ToDo?

tools/bumbershoot/idpassemble/idpassemble.xml
```diff
@@ -0,0 +1,122 @@
...
+            <param name="min_spectra_per_distinct_peptide" value="1" />
+            <param name="max_protein_groups_per_peptide" value="10" />
+            <output name="output" file="../../idpassemble/test-data/201208-378803.idpDB" compare="sim_size" delta_size="500000" />
+        </test>
+    </tests>
+    <help>
+<![CDATA[
+**What it does**
+
+Merges and filters one or more IDPicker 3 idpDB files into a combined idpDB file. Protein assembly (e.g. parsimony) is conducted on the combined set of proteins.
```
bgruening: If you have a link to more documentation or some images, detailed explanations ....

tools/bumbershoot/idpqonvert/README.md
```diff
@@ -0,0 +1,51 @@
...
+Contributions to this repository are reviewed through pull requests. If you would like your work acknowledged, please also add yourself to the Authors section. If your pull request is accepted, you will also be acknowledged in <https://github.com/galaxyproteomics/tools-galaxyp/>
+
+
+Authors
+-------
+
+Authors and contributors:
+
+* John Chilton <jmchilton@gmail.com>
+* Minnesota Supercomputing Institute, Univeristy of Minnesota
```
bgruening: Your name is missing here.

tools/bumbershoot/idpqonvert/idpQonvertEmbedder.xml
```diff
@@ -0,0 +1,124 @@
...
+        <requirement type="package" version="3.0.8789">bumbershoot</requirement>
+    </requirements>
+    <stdio>
+        <exit_code range="1:" level="fatal" description="Job Failed" />
+        <regex match="\nError:.*\n" source="both" level="fatal" />
+    </stdio>
+    <command>
+<![CDATA[
+        #set $input_name = $input.display_name
+        cp '$input' output;
```
bgruening: any reason to not link the file, or move it?

packages/package_bumbershoot_3_0_8789/tool_dependencies.xml
```diff
@@ -0,0 +1,48 @@
...
+                <action type="move_file">
+                    <source>tagrecon</source>
+                    <destination>$INSTALL_DIR</destination>
+                </action>
+                <action type="move_file">
+                    <source>pepitome</source>
+                    <destination>$INSTALL_DIR</destination>
+                </action>-->
+                <action type="set_environment">
+                    <environment_variable name="PATH" action="prepend_to">$INSTALL_DIR</environment_variable>
```
bgruening: please include a `BUMBERSHOOT_ROOT_PATH` path here with `set_to`

tools/bumbershoot/idpqonvert/idpQonvertEmbedder.xml
```diff
@@ -0,0 +1,124 @@
...
+        ln -s output output.idpDB;
+        ln -s \$(dirname \$(which idpQonvert))/gene2protein.db3 gene2protein.db3;
```
bgruening: `idpQonvert` is your binary and gene2protein.db3 is directly next to it?
You can simply use `\$BUMBERSHOOT_ROOT_PATH/`

tools/bumbershoot/idpqonvert/idpQonvertEmbedder.xml
```diff
@@ -0,0 +1,124 @@
...
+                -QuantitationMethod $use_raw_data_condition.quantitationMethod
+            #end if
+            output.idpDB;
+]]>
+    </command>
+    <inputs>
+        <param name="input" type="data" format="idpdb" label="Input idpDB file" />
+        <param name="embedGeneMetadata" type="boolean" value="true" label="Embed Gene Metadata?" help="Allows gene-centric analysis in IDPicker. The gene metadata embedding only works for human/mouse proteins from a RefSeq database. For best results, run idpQonvert with RefSeq no matter what database was used for the search: idpQonvert always remap the peptides anyway. Non-human/mouse proteins and any unmappable human/mouse proteins will get gene ids like &quot;Unmapped_&lt;protein accession&gt;&quot;. This option, run by itself, is very fast." />
+        <conditional name="use_raw_data_condition">
+            <param name="use_raw_data" type="boolean" label="Do you have the raw data (spectra)?" help="More embed options are available if the raw spectra are available." />
```
bgruening: select

tools/bumbershoot/idpqonvert/idpqonvert.xml
```diff
@@ -0,0 +1,228 @@
...
+        <exit_code range="1:" level="fatal" description="Job Failed" />
+        <regex match="^Error:.*$" source="both" level="fatal" />
+    </stdio>
+    <command>
+<![CDATA[
+        #set $db_name = $input_database.display_name.replace(".fasta", "") + ".fasta"
+        #set $input_name = $input.display_name
+        #set $output_name = $input_name.split(".")[0] + ".idpDB"
+        ln -s '$input' '${input_name}';
+        ln -s '$input_database' '${db_name}';
```
bgruening: In general you should use && and not ; to concatenate commands. See here why: http://stackoverflow.com/questions/6152659/bash-sh-difference-between-and?rq=1

tools/bumbershoot/idpqonvert/idpqonvert.xml
```diff
@@ -0,0 +1,228 @@
...
+            #end if
+            '${input_name}';
+        mv '$output_name' output
+]]>
+    </command>
+    <inputs>
+        <param name="input" type="data" format="mzid,raw_pepxml" label="Input mzIdentML/pepXML file(s)" />
+        <param name="input_database" format="fasta" type="data" label="Protein Database"/>
+        <param name="decoy_prefix" type="text" label="Decoy Prefix"/>
+        <param name="max_fdr" type="float" label="Max FDR" value="0.05" />
```
bgruening: You could think of to use more macros, `max_fdr` at a few more paramters are shared between different tools and you can have one macro for all of them.

tools/bumbershoot/idpqonvert/idpqonvert.xml
```diff
@@ -0,0 +1,228 @@
...
+                <param name="qonverter_method" label="Qonverter Method" type="select">
+                    <option value="MonteCarlo" selected="true">Monte Carlo</option>
+                    <option value="SingleSVM">SVM (single)</option>
+                    <option value="PartitionSVM">SVM (parition)</option>
+                    <option value="StaticWeighted">Static Weighted</option>
+                </param>
+            </when>
+        </conditional>
+        <conditional name="scoring">
+            <param name="override_scoring" type="boolean" truevalue="true" falsevalue="false" label="Modify Search Engine Scores/Weights" />
```
bgruening: select

tools/bumbershoot/idpqonvert/idpqonvert.xml
```diff
@@ -0,0 +1,228 @@
...
+                <param name="gamma" type="float" label="Gamma" value="5" />
+                <param name="nu" type="float" label="Nu" value="-0.5" />
+                <param name="true_positive_threshold" type="float" label="True Positive Threshold" value="0.01" />
+                <param name="polynomial_degree" type="integer" label="Polynomial Degree" value="3" />
+                <param name="min_partition_size" type="integer" label="Minimum Partition Size" value="10" />
+                <param name="svm_type" label="SVM Type" type="select">
+                    <option value="CSVC" selected="true">CSVC</option>
+                    <option value="NuSVC">Nu SCV</option>
+                    <option value="OneClass">One Class</option>
+                    <option value="NuSVR">Nu SVR</option>
```
bgruening: Maybe a better description of the different types?

tools/bumbershoot/idpqonvert/idpqonvert.xml
```diff
@@ -0,0 +1,228 @@
...
+                </param>
+            </when>
+        </conditional>
+        <conditional name="scoring">
+            <param name="override_scoring" type="boolean" truevalue="true" falsevalue="false" label="Modify Search Engine Scores/Weights" />
+            <when value="false" />
+            <when value="true">
+                <section name="myrimatch_mvh" title="Myrimatch (mvh) Score/Weight" expanded="False">
+                    <param name="myrimatch_mvh_weight" label="Myrimatch (mvh) Weight" type="float" value="1" />
+                    <param name="myrimatch_mvh_type" label="Myrimatch (mvh) Normalization" type="select"><expand macro="normalizationOptions" /></param>
```
bgruening: a line break for the macro would be nice

tools/bumbershoot/idpqonvert/idpqonvert.xml
bgruening: You could think of to use more macros, `max_fdr` at a few more paramters are shared between different tools and you can have one macro for all of them.

tools/bumbershoot/myrimatch/myrimatch.xml
```diff
@@ -0,0 +1,241 @@
+<tool id="myrimatch" version="0.1.0" name="MyriMatch">
+    <description></description>
```
bgruening: short description ... Searchengine for ...

tools/bumbershoot/myrimatch/myrimatch.xml
```diff
@@ -0,0 +1,241 @@
...
+
+        #set $dynamic_mods_str = ""
+        #for $dynamic_mod in $dynamic_mods
+            #set $dynamic_mods_str = $dynamic_mods_str + " " + str($dynamic_mod.motif) + " * " + str($dynamic_mod.mass)
+        #end for
+        #set $dynamic_mods_str = $dynamic_mods_str.lstrip()
+
+        ln -s '$input' '${input_name}';
+        ln -s '$input_database' '${db_name}';
+        nice myrimatch -cpus 1 -DecoyPrefix '${decoy_prefix}'
```
bgruening: nice? I don't think this is a good idea :)

tools/bumbershoot/idpquery/idpquery.xml
```diff
@@ -0,0 +1,138 @@
+<?xml version="1.0"?>
+<tool id="idpquery" name="idpQuery" version="0.1.0">
+    <description>Creates text reports from idpDB files.</description>
+    <macros>
+        <xml name="proteinGroupByColumns">
+            <param name="report_columns" type="select" label="Select report columns" display="checkboxes" multiple="true">
+                <option value="Accession">Accession</option>
+                <option value="GeneId">Gene Id</option>
```
bgruening: Here nothing is selected by default, is this correct?

tools/bumbershoot/myrimatch/tool_dependencies.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0"?>
+<tool_dependency>
```
bgruening: Also these files you could symlink into the separated folders.

bgruening:
@chambm this is awesome!
Very good wrappers and a few nice ideas included. Please see my in line comments.

General remarks:
 * conditionals should use select and not booleans (best practice)
 * `&&` instead of `;`
 * reorganisation of files: I would prefer you create symlinks so that every subfolder has a complete set of files; An other idea would be to put everything in one folder and use the planemo functionality to decompose your folder into 4 different TS repositories
 * more, detailed help text is always good
 * please consider to add `(--parameter)` to the help section of every param (best practice)

If you are ok with it, please move all packages and tools over to the user `galaxyp` I will upload them to the TS and provide write-access to you.

bgruening:
@galaxyproteomics/owners I had the pleasure to work with @chambm the last weeks and he did an incredible good job here.
He also fixed a few Galaxy bugs on his way, learned all the magic behind the Galaxy Docker Image and created a new data type.

I would like to endorse him to become a regular GalaxyP contributor and with a little bit luck he will help us reviewing pull requests and maintaining tools. What's your opinion to grant him commit access?

tools/bumbershoot/idpassemble/idpassemble.xml
chambm: I'm making a copy of input and then modifying it. Don't want to modify the input file directly, although that would be most efficient.

tools/bumbershoot/idpassemble/idpassemble.xml
chambm: VirtualBox doesn't like symlinks, and since I put my code inside the VM and share it with my Windows machine outside for editing, I couldn't go that route (I tried). I'll test it with --installed now. It works with planemo test.

tools/bumbershoot/idpassemble/idpassemble.xml
chambm: No. I was working on a git smudge/clean script that would replace {$sha1$} with the current commit hash so that the citations part would link to the correct github commit. It was going to be really cool, but I couldn't get git to run the script and I couldn't spend time to figure out why. So I commented it out for later work.

tools/bumbershoot/idpqonvert/idpQonvertEmbedder.xml
chambm: Like idpAssemble, when idpQonvert operates on an idpDB input file, it changes the input rather than creating a new output. When idpQonvert operates on pepXML/mzidentML files, it creates idpDB files.

packages/package_bumbershoot_3_0_8789/tool_dependencies.xml
chambm: Set to $INSTALL_DIR?

tools/bumbershoot/idpqonvert/idpQonvertEmbedder.xml
chambm: That would be much clearer. :)

tools/bumbershoot/idpassemble/idpassemble.xml
chambm: Will a select conditional give a nice "Yes/No" button? If not, I'd prefer to keep the boolean (although I'm open to changing to a <section> rather than <conditional>, since I learned about that previously undocumented feature).

tools/bumbershoot/idpassemble/idpassemble.xml
chambm: The command-line parameter names are almost all trivial to map to the param names. A longer prose description of what the parameter means would make more sense.

tools/bumbershoot/idpqonvert/idpqonvert.xml
chambm: Makes sense. I was just following what was there in galaxyp, but I'll change it.

tools/bumbershoot/idpqonvert/idpqonvert.xml
chambm: Hmm, not really. If I add help attributes, they won't be entirely the same between different tools. I would prefer to use macros only when there is a significant savings in space or complexity of updating it.

tools/bumbershoot/idpassemble/idpassemble.xml
trevor: It seems like an appropriate use of a boolean to me. I think the linked text is poorly worded, the conditional they're describing is one that changes the interface dynamically, not a blanket statement to not use conditionals / checkboxes, which certainly have their place.

> Boolean should not be used as conditionals. For conditionals please use a select type as described in Dynamic Options below.
>
>
>[...]
>
> Certain options such as "Advanced Options" should be a select box and not a boolean. A checkbox is not expected to change the content of the mask from a user point of view.
</blockquote>

If it is changing other values in the interface, then I agree with @bgruening a select box should be used.

tools/bumbershoot/idpqonvert/idpqonvert.xml
chambm: Eh, that's ok, MonteCarlo is much better. I should document that though. ;)

tools/bumbershoot/idpquery/idpquery.xml
chambm: Can I set a default with selected="true"? I don't think I've tried it yet. The only default would be Accession.

tools/bumbershoot/myrimatch/myrimatch.xml
chambm: Yeah, probably not necessary. Why is it a bad idea though? If it's in a cgroup, doesn't it get it's own scheduling? I think I put it in there because I do testing running the Galaxy server and the tasks on the same VM, so 'nice'ing this very intensive process (it will easily use 100% of the CPU) keeps the server responsive. Perhaps that should be part of the Galaxy queuing system when using the local launcher?

packages/package_bumbershoot_3_0_8738/tool_dependencies.xml
chambm: How do I add things there?

tools/bumbershoot/idpqonvert/idpQonvertEmbedder.xml
jmchilton: Hmm... disagree on this point. I like what is there better since it will just work with manually installed tool dependencies and doesn't require non portable tool shed magic.

packages/package_bumbershoot_3_0_8738/tool_dependencies.xml
bgruening: as a normal pull request

tools/bumbershoot/idpassemble/idpassemble.xml
bgruening: This text was originally written for the old tool form interface. In the old tool form a boolean was rendered as a checkbox. And imho no user expects that a checkbox triggers a user interface change. This is why we encourage people to use a `select` for cases where we get new options and the interface changes.
I'm fine with changing it to sections as well!

tools/bumbershoot/idpassemble/idpassemble.xml
bgruening: It's a recommendation you don't need to stick to it, but we encourage you to stay consistent with other tools. Moreover, we introduced a new `attribute="--my-param"` param-attribute. If you use this one, you don't need a name-attribute (this will be automatically generated from the attribute) and the `--my-param` will be automatically attached to the help text.

tools/bumbershoot/idpassemble/idpassemble.xml
bgruening: Then maybe the decompose feature of planemo is the correct choice for you.

tools/bumbershoot/idpassemble/idpassemble.xml
bgruening: Git commit hooks are the way to go here, but they are usually not portable.

tools/bumbershoot/idpqonvert/idpQonvertEmbedder.xml
bgruening: Ok, perfect!

tools/bumbershoot/idpqonvert/idpQonvertEmbedder.xml
bgruening: @jmchilton it's not portable either. `which` and `dirname` needs to be installed.
Whatever dependency mechanism you prefer to use you can simply set the `BUMBERSHOOT_ROOT_PATH`. Even in `job_conf.xml` if needed.

packages/package_bumbershoot_3_0_8789/tool_dependencies.xml
bgruening: Yes.

tools/bumbershoot/idpassemble/idpassemble.xml
chambm: Yeah, run_functional_tests didn't work. I can't find any documentation 
for the "decompose" feature. What's that?

tools/bumbershoot/idpquery/idpquery.xml
bgruening: Yes. :)

tools/bumbershoot/myrimatch/myrimatch.xml
bgruening: Exactly, this should be handled by the job scheduler and/or Galaxy. Same goes for memory recommendations or CPU count.

chambm: How do I do that? By default MyriMatch will take all the CPUs it can see. With my test setup, that meant all of them. So if I was running multiple jobs, the docker-galaxy-stable would run 3 of those, each of which would run 8 threads, with default priority.

tools/bumbershoot/idpqonvert/idpqonvert.xml
bgruening: You can have nested macros and inheritance if you like. But I understand if the overhead is to big. Was just a recommendation :)

tools/bumbershoot/idpqonvert/idpqonvert.xml
bgruening: Oh we should change that in all other tools as well.

tools/bumbershoot/idpqonvert/idpQonvertEmbedder.xml
jmchilton: Obviously we disagree on this :crying_cat_face:. Packages should just add things to the ``PATH`` not inject all sorts of environment variable magic that tools depend on. These make the tool more difficult to depend and the packages harder to port to other package management tools.

tools/bumbershoot/idpqonvert/idpqonvert.xml
jmchilton: Yup, I used to use ``;`` a lot - it was a mistake.

tools/bumbershoot/myrimatch/myrimatch.xml
jmchilton: ``myrimatch -cpus \${GALAXY_SLOTS:-1} -DecoyPrefix``

jmchilton:
:+1: on giving @chambm commit access - he is obviously the best person to review these tools and other very important tools such as msconvert.

trevor:
Invited

tools/bumbershoot/idpassemble/idpassemble.xml
bgruening: I understand, but this is at the end of you command. Not before the command?

tools/bumbershoot/idpqonvert/idpQonvertEmbedder.xml
bgruening: All kind of packages rely on additional ENV vars, R_HOME etc ... we need to have a way to do this as well.
In this case the best solution would be a data_manager, but this can be added later I guess.
I don't fell strongly about this, @chambm decide what you like most.

tools/bumbershoot/myrimatch/myrimatch.xml
bgruening: Thanks @jmchilton!

bgruening:
Thanks @trevor!

tools/bumbershoot/idpassemble/idpassemble.xml
chambm: Sorry, this one doesn't make sense as a section. The Advanced Options in myrimatch.xml does (and I'm changing it). This boolean option (rendered as a yes/no switch rather than a checkbox) changes the meaning and text of several parameters below it, but the help text of the button makes that clear. It doesn't add or drop any parameters like an advanced options section does.

tools/bumbershoot/idpassemble/idpassemble.xml
chambm: Ah, reading your comment again, you did say `attribute="--my-param"` rather than `argument="--my-param"`. No wonder I got confused. ;)

tools/bumbershoot/idpassemble/idpassemble.xml
bgruening: @chambm I can only recommend this to you. It's your baby and you need to decide. We tried to make a decision here so that most tools look the same and behave the same. Imho this has a great advantage for the usability of all tools in Galaxy.

packages/package_bumbershoot_3_0_8738/tool_dependencies.xml
chambm: It appears I have to clone the entire repo to do that? Is there a way to checkout just a skeleton of the repo which I can modify? The depth option seems to preclude pushing changes.

bgruening: No, sorry :( I know its getting large. Maybe I can remove, clean the history at some point.

chambm: Doesn't seem scalable. Would it work if tarballs are added and then deleted immediately, so that they're still available in the history, but not in HEAD?

tools/bumbershoot/idpassemble/idpassemble.xml
chambm: Yeah, that was silly. Will be fixed.

packages/package_bumbershoot_3_0_8738/tool_dependencies.xml
bgruening: This will not help. The Galaxy project is working on a more scalable solution. This was my poor many approach and it is serving me well. But I see the end ... maybe a second repo ;)

chambm:
@trevor: Can you also give me upload access on the test toolshed for galaxyp? Or do you actually have to give me the key for that user? I guess that would be ok on TTS?

packages/package_bumbershoot_3_0_8738/tool_dependencies.xml
chambm: It would work for a SVN repository. But it's also just basically a release system. I left it as-is for now. Maybe by the time it's a problem (i.e. LabKey goes down), the Galaxy project's solution will be ready. :)

trevor:
> Can you also give me upload access on the test toolshed for galaxyp? Or do you actually have to give me the key for that user? I guess that would be ok on TTS?

@bgruening @jj-umn are either of you able to do this?

bgruening:
I ask to get a proper TS group for us. If this does not work out I will send @chambm the login details.

bgruening:
Again this is an awesome work! Thanks @chambm and sorry for being so nitpicking, in the end I think these wrappers are good prime examples for proteomics Galaxy tools.

Thanks a lot!
Bjoern

PR #51, opened 2015-10-14
Update peptideshaker to SearchGUI 2.1 and PeptideShaker 1.1

jj-umn:
This relies on a new binary SearchGUIArchive datatype in galaxy that would track versions in metadata.  
Should also remove repository_dependencies.xml  since datatypes are in main galaxy distribution,
We should also determine if searchgui_mods.loc.sample needs any updates.

### DIFF
--- a/tools/peptideshaker/searchgui.xml
+++ b/tools/peptideshaker/searchgui.xml
@@ -1,4 +1,4 @@
-<tool id="search_gui" name="Search GUI" version="2.1.1">
+<tool id="search_gui" name="Search GUI" version="@SEARCHGUI_VERSION@.1">
     <description>
         Perform protein identification using various search engines and prepare results for input to Peptide Shaker
     </description>
@@ -6,7 +6,7 @@
         <import>macros.xml</import>
     </macros>
     <requirements>
-        <requirement type="package" version="2.1">searchgui</requirement>
+        <requirement type="package" version="@SEARCHGUI_VERSION@">searchgui</requirement>
         <environment_variable name="LC_ALL" action="set_to">C</environment_variable>        
     </requirements>
     <expand macro="stdio" />
--- a/tools/peptideshaker/searchgui.xml
+++ b/tools/peptideshaker/searchgui.xml
@@ -1,4 +1,4 @@
-<tool id="search_gui" name="Search GUI" version="1.28.0">
+<tool id="search_gui" name="Search GUI" version="2.1.1">
     <description>
         Perform protein identification using various search engines and prepare results for input to Peptide Shaker
     </description>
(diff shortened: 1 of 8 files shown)


tools/peptideshaker/searchgui.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="search_gui" name="Search GUI" version="1.28.0">
+<tool id="search_gui" name="Search GUI" version="2.1.1">
```
bgruening: Can we reuse the version macro here. Ideally we don't need to update the version number manually.

bgruening:
@jj-umn like it! :+1:

iracooke:
@jj-umn this is great.

jj-umn:
Updated the extensions used in the tool wrappers to reflect the change in extension names:


http://compomics.github.io/projects/peptide-shaker.html 
commit 02763fbff4c2b207d5a80f61196996c039baa5bb
Author: Harald Barsnes <harald.barsnes@gmail.com>
Date:   Tue Aug 25 14:23:15 2015 +0200
    Renamed .cps to .cpsx and .parameters to .par.

jj-umn:
package_peptideshaker_1_1 tool_dependencies.xml 
The download of zip results in a redundant PeptideShaker-1.1.1 dir, changed the source_directory for the move_directory_files action.  

This also requires:
commit 197c7b21b209a7b6fbdb9a8a11c232a2b88523fc
Author: Nicola Soranzo <nicola.soranzo@tgac.ac.uk>
Date:   Thu Sep 3 12:55:28 2015 +0100
    Fix for fix for fix for handling of unix permissions in .zip files.
    Append common_prefix to extraction_path only when returning.

jj-umn:
@iracooke @bgruening 
With SearchGuiArchive datatype in current dev lib/galaxy/datatypes/binary.py 
and changes in this pull request, I was able to run searchgui and peptideshaker on the sample files from our GCC2015 tutorial  at UMN on galaxyp server.

Still need to test if any changes are required for:  searchGUI_mods.xml   searchGUI_usermods.xml

tools/peptideshaker/searchgui.xml
```diff
@@ -168,6 +171,27 @@
                 -myrimatch_max_peak ${myrimatch.myrimatch_max_peak}                    
             #end if
 
+            #*
+            #if $andromeda.andromeda_advanced == "yes"
+                -andromeda_max_pep_mass ${andromeda.andromeda_max_pep_mass}
```
bgruening: @jj-umn here you include the andromeda engine but below it's commented out. I don't think this will work.
Moreover, I don't think we should include Andromeda. The only setup this would be useful is a windwos-only worker and this would mean that all search engines run on windows + SearchGUI. Not sure if this is worth the maintenance and the user confusion on all other installations.
If we could get this bundled with Mono (I made a package for this) this would change the game and make it possible to use this engine on a Linux Cluster as well. What do you think?

jj-umn: I'll remove it.   We only need in the command line generation:  -andromeda 0
Otherwise, the default is 1 (run andromeda).

bgruening: Thanks @jj-umn !

PR #52, opened 2015-11-06
Update peptideshaker tool_dependencies.xml

bgruening:
This is needed because we are inside the folder after downloading.


jj-umn:
Thanks, @bgruening

PR #54, opened 2015-11-10
msconvert/pwiz update

chambm:
- update ProteoWizard to a recent version
- streamline the filters by dividing them into sections: "Data Processing", "Inclusion/Exclusion Filters", and "Output Encoding Settings"
- add functional tests for almost all filters

I kept msconvert and msconvert_raw as separate tools but factored out more of their redundant XML via macros; the only difference should be whether they are run on Windows or Linux, and whether they accept vendor formats as input.

I haven't tested msconvert_raw on Windows yet (I haven't even added the Windows package yet). In fact, because it will double the amount of binary data to download, we may want to make these two separate tools (i.e. separate .shed.yml). That way the user can download just the one they need.


packages/package_proteowizard_3_0_9016/tool_dependencies.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0"?>
+<tool_dependency>
+    <package name="proteowizard" version="3.0.9016">
+        <install version="1.0">
+            <actions>
+                <action type="download_by_url">https://teamcity.labkey.org/guestAuth/repository/download/bt17/279368:id/pwiz-bin-linux-x86_64-gcc48-release-3_0_9016.tar.bz2</action>
```
bgruening: Is this a stable URL, or should we mirror it?

tools/msconvert/msconvert_raw.xml
```diff
@@ -1,396 +1,282 @@
...
+      <output name="output" file="small-zlib-64.mz5" compare="sim_size" delta="100" />
+    </test>
+    <test>
+      <param name="input" value="small-peakpicking-cwt-allMS.mzML" />
+      <param name="output_type" value="mzXML" />
+      <param name="mz_encoding" value="32" />
+      <param name="intensity_encoding" value="32" />
+      <output name="output" file="small-zlib-32.mzXML" />
+    </test>
+    <!-- TODO: how to test gzipped output?
```
bgruening: The question is my do you get gz output. I would unpack it and provide the unpacked file. This can be consumed by many other tools as well.

tools/msconvert/msconvert_macros.xml
```diff
@@ -0,0 +1,390 @@
+<macros>
+  <xml name="msconvertCommand">
+    <command>
+<![CDATA[
+    #import re
```
bgruening: this is not needed anymore, isn't it?

tools/msconvert/msconvert_macros.xml
```diff
@@ -0,0 +1,390 @@
...
+      #end if
+
+      &&
+
+    #if $data_processing.precursor_refinement.use_mzrefinement
+      mv '${input.name.rsplit('.',1)[0]}.mzRefinement.tsv' output.refinement &&
+    #end if
+
+    mv 'outdir/$output_name' output &&
+    rmdir outdir
```
bgruening: rmdir is not needed.

tools/msconvert/msconvert_macros.xml
```diff
@@ -0,0 +1,390 @@
...
+      #else if $binary_compression == "numpressLinear"
+      --numpressLinear
+      #else if $binary_compression == "numpressPic"
+      --numpressPic
+      #else if $binary_compression == "numpressSlof"
+      --numpressSlof
+      #end if
+
+      #if $settings.gzip_compression
+      --gzip
```
bgruening: Do we want this? See below.

bgruening:
@chambm this is an awesome cleanup! Thanks a lot!
I wouldn't split this up into 2 directories, I think I would like to see plenemo demulitplexing here. I can patch this in as soon as the gets merged.

packages/package_proteowizard_3_0_9016/tool_dependencies.xml
chambm: It's not stable. Needs a mirror. I can't think of a better approach than your github one, although I don't like having to clone the huge repo to add files. :/

tools/msconvert/msconvert_raw.xml
chambm: We can save a lot of space if we keep it compressed. Some tools can operate on it in the native compressed form (pwiz based tools, at least, like msconvert, myrimatch). I don't think MSGF/xtandem/omssa can, but we could unzip it on the fly for those.

tools/msconvert/msconvert_macros.xml
chambm: It seems not!

tools/msconvert/msconvert_raw.xml
bgruening: We should define than a new filetype for it.

packages/package_proteowizard_3_0_9016/tool_dependencies.xml
bgruening: You you go: https://github.com/bgruening/download_store/tree/master/pwiz

tools/msconvert/msconvert_raw.xml
chambm: How do I go about that? I'd add ones for mzML, mzXML, mzIdentML, pepXML, and maybe MGF as well.

bgruening: We could get this working even with a associated converter. So that every tool that can consume mzML, can also use bgzip_mZML, but this would involve a little bit of work. Can we release this version for now without the compressed file output?

Most of the Galaxy servers have file compression enabled and this is not really an issue, just increases complexity.

chambm: Yes, I'm fine with the way it is (commented out). I'd prefer to leave the option to compress in though, in case someone just wants to use it to convert a bunch of data into a reasonably compact form. Pwiz-based tools can consume the file without decompressing first (which is what I assume the "associated converter" would do).

bgruening: ok, yes the converter would unpack it. The example is the VCF datatype and vcf_bzip. But I have no time this week to work on it, sorry.

jj-umn:
I liked to review this at the Galaxy-P meeting on Friday with the grant principal investigators.  
I think this is a good approach factoring out common elements from msconvert.xml and msconvert_raw.xml  It will be good to have a tool_dependency for the linux binary.   Is it even possible to do so for the Windows code?   I have presumed that the Windows side of this would have to be set up manually by an admin.   
We would definitely want to bump the tool version numbers.  I can anticipate our researchers complaining of broken workflows unless we can retain the current versions.  
Including the wiff format for msconvert_raw.xml will be important at Minnesota.

jj-umn:
@chambm @bgruening  @jmchilton @iracooke 

The new tool form layout was well received at Galaxy-P project meeting.
 - Let's have tool versions reflect proteowizard version: 3.0.0 to start with?

I think we should break this into 2 tool repositories:

  - msconvert - refactored, but geared toward Windows deployment with vendor libs, requires admin to setup Windows and pulsar

  - msconvert_public - includes tool_dependencies.xml and <requirements>, can be automatically installed on *nix platforms

github layout:
  msconvert/
    .shed.yml
    COPYING
    msconvert_macros.xml
    msconvert_wrapper.py
    test-data/
    msconvert/
      msconvert.xml (With vendor formats)
      msconvert_wrapper.py -> ../msconvert_wrapper.py
      msconvert_macros.xml -> ../msconvert_macros.xml
      test-data  -> ../test-data
    msconvert_public/
      msconvert.xml (No vendor formats)
      msconvert_wrapper.py -> ../msconvert_wrapper.py
      msconvert_macros.xml -> ../msconvert_macros.xml
      test-data  -> ../test-data
      tool_dependencies.xml

msconvert_wrapper.py

I would still use the msconvert_wrapper.py inorder to abstract the differences for executing on a *nix host vs Windows.

Many of the flag options could be passed to msconvert_wrapper.py as args after '--' rather than options, simplifying the msconvert_wrapper.py options:
msconvert_wrapper.py --input="dataset_111.dat" --input_name="my.mzML" --fromextension=mzml --toextension=mgf --output="dataset_222.dat" -- --64 --mz64 --inten32

bgruening:
Fully agree! 
For windows you need to install your deps manually.

iracooke:
I have not had time to check out Matt's changes but I definitely agree with the idea of splitting into two repositories and having the linux one with tool_dependencies.

jj-umn:
@chambm   Can you provide me with a copy of Rpal_01.pepXML.gz so I can try testing the mzRefine filter?   Thanks.

chambm:
@jj-umn Added to repo. It was supposed to be there already, but it .gz files are .git-ignored, so I had to override that.

jj-umn:
@chambm   Yes, "msconvert_public" is an awful name.  I wanted to keep the msconvert as the windows version for historical consistency.   How about something like "msconvert_x" for uniX-derived systems eXcluding the windows-only vendor libs?

chambm:
@jj-umn Unless I'm missing something, the current version has msconvert_raw as the Windows version; msconvert is the non-Windows version. I think msconvert_x makes sense, but not intuitively. It has to be explained. the windows/linux dichotomy is clearer, but could also be _win/_nix.

bgruening:
:+1: for `_win/_nix`

bgruening:
@chambm do you plan to work on this further?

chambm:
Yes, but it will have to wait a few more weeks. Just had a baby a couple weeks ago and working on a master's thesis. :)

bgruening:
Wonderful new @chambm - enjoy! Work can wait!

PR #55, opened 2015-11-10
- add standalone msgfplus wrapper that doesn't require protk

chambm:
- add package_msgfplus_v10089
- add example output (created by searching symlinks to the Myrimatch input files)
- add common and custom mod parameters
- add functional tests

I recognize this PR may controversial as it is functionally quite redundant with Ira's MSGF wrapper. I created a standalone wrapper in galaxyp rather than forking Ira's protk version because I am doing (relatively) frequent docker deployments. Doing enough of those, it got to the point where downloading and building ruby/protk was an unnecessary delay. It also lets me make changes without juggling another repository, and it's decoupled from ProteoWizard: why not have idconvert in a separate tool? I guess because it's expected to feed the output to the TPP in the protk suite. But then why not also have TPP as a tool dependency?

The advantage of the protk version is it has some built-in databases. But there has to be a way to handle that more universally in Galaxy, like is done for data managers. Not that I've looked into that... I've just been uploading my own FASTAs.

The only advantage of my version other than not requiring protk is that it allows specifying custom mods in addition to the common mods. I opted to just put them in the XML rather than reading them from a file; it's only marginally harder to edit, and unless we're going to get the mods file from Unimod I don't see much point in the file approach. What would be really cool is a text search that looks up the mod in Unimod by name, e.g. "phospho" or "carbam". :)

I plan on porting the common mods feature to MyriMatch as well (right now it just has custom mods); would it make sense at that point to factor out the duplicate XML into macros, so that different tools are using the same masses and "common mods"? The only problem is the mod syntax is different for each engine. Hah, at that point it could make sense to go back to using a file, with one column being the name and the other columns being the specification for different engines.

### DIFF
--- a/packages/package_msgfplus_v10089/.shed.yml
+++ b/packages/package_msgfplus_v10089/.shed.yml
@@ -1,6 +1,6 @@
 categories:
 - Tool Dependency Packages
-description: Contains a tool dependency definition that downloads the Bumbershoot tools.
+description: Contains a tool dependency definition that downloads the MS-GF+ MS/MS search engine.
 long_description: |
   MSGF+
 
--- a/packages/package_msgfplus_v10089/.shed.yml
+++ b/packages/package_msgfplus_v10089/.shed.yml
@@ -1,0 +1,11 @@
+categories:
+- Tool Dependency Packages
+description: Contains a tool dependency definition that downloads the Bumbershoot tools.
+long_description: |
+  MSGF+
+
+  http://proteomics.ucsd.edu/software-tools/ms-gf/
+name: package_msgfplus_v10089
+remote_repository_url: https://github.com/galaxyproteomics/tools-galaxyp/packages/package_msgfplus_v10089/
+owner: galaxyp
+type: tool_dependency_definition
--- a/packages/package_msgfplus_v10089/tool_dependencies.xml
+++ b/packages/package_msgfplus_v10089/tool_dependencies.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0"?>
+<tool_dependency>
+    <package name="msgfplus" version="v10089">
+        <install version="1.0">
+            <actions>
+                <action type="download_by_url">http://omics.pnl.gov/sites/default/files/software/MSGFPlus.20140716.zip</action>
+                <action type="move_file">
+                    <source>MSGFPlus.jar</source>
+                    <destination>$INSTALL_DIR</destination>
+                </action>
+                <action type="set_environment">
+                    <environment_variable name="PATH" action="prepend_to">$INSTALL_DIR</environment_variable>
+                </action>
+            </actions>
+        </install>
+        <readme>
+            This package downloads and installs MSGF+.
+        </readme>
+    </package>
+</tool_dependency>
--- a/tools/msgfplus/.shed.yml
+++ b/tools/msgfplus/.shed.yml
@@ -1,0 +1,11 @@
+owner: galaxyp
+name: msgfplus
+categories:
+- Proteomics
+description: MSGF+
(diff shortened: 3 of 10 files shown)


bgruening:
+1 for a common file with all mods :)
This file should be updated with a script automatically.
@iracooke can you review this one and give your opinion on it?

iracooke:
@chambm @bgruening Sorry I only just saw this.  Not sure how I missed it. 
I'm totally in favour of making this independent of protk.  As it stands this wrapper looks great.  I also agree that there is no need for a separate mods file.  

Since this is a replacement for the protk wrapper we would also need another tool to convert mzid to pepXML using idconvert. This involves some minor hacks to both the input mzid file (to make it work with idconvert) and then to the pepXML output. In the protk wrapper at the moment this is done with a "convert to pepXML check box". My usual workflow is to run through peptide prophet and the other TPP tools.

Come to think of it, it would be great to have an idconvert tool anyway so I am certainly in favour of splitting this up.  I just need some time to test things out.

bgruening:
@iracooke thanks for your comments!
@chambm I'm ready for merging this. Sorry that it took me too long. Are you volunteering for the `convert mzid to pepXML using idconvert` wrapper? :)

Happy new year GalaxyP community!

chambm:
@iracooke Can you give me some test criteria for idconvert's pepXML output so that it will work as output to TPP? I can integrate that into the pepXML output unit test and fix it from pwiz rather than hacking the wrapper.

iracooke:
Hi Matt, 


Thanks for offering to do this. I believe it would be a nice improvement. And as Pratik says it would be really useful to have a separate idconvert wrapper.  Effectively my original msgfplus tool would then be replaced by your new wrapper plus a new general purpose idconvert tool.

As far as I can see the main issues are;

1. If any of the database sequences contain ambiguous residues ms-gf will denote these with a ? in the pre / post tags. For example

    <PeptideEvidence id="PepEv_3452745_1_63" peptide_ref="Pep1" dBSequence_ref="DBSeq3452683" start="63" end="91" pre=“?" post="L" isDecoy="false"/>

In my wrapper I substitute these ? symbols with an X and idconvert is happy.

2. Within the pepXML output the attribute “base_name” occurs in several places and in all cases it should be a full path to the original mass spec input file minus its extension (.mzML).  At least this is how it is defined in the schema.  I cant remember the specifics but I believe that some TPP tools will fail if this isn’t set properly.  idconvert changes this value to a relative path and strips off the .dat extension which it shouldn’t do.

http://sashimi.sourceforge.net/schema_revision/pepXML/pepXML_v117.xsd <http://sashimi.sourceforge.net/schema_revision/pepXML/pepXML_v117.xsd>

For example here is an mzid file converted to pepXML by idconvert

<msms_run_summary base_name="dataset_582" raw_data_type="" raw_data=“">
<search_summary base_name="dataset_582" search_engine="MS-GF+" precursor_mass_type="monoisotopic" fragment_mass_type="monoisotopic" out_data_type="" out_data=“">

Note that the relevant part of the original mzid file is 

<SpectraData id="SID_1" name="dataset_582.dat.mzML" location="/mnt/galaxy/files/000/dataset_582.dat”>

And as you can see the original input file had both a .dat and a .mzML extension.

And here are is how I believe the output pepXML should look

<msms_run_summary base_name="/mnt/galaxy/files/000/dataset_582.dat" raw_data_type="" raw_data=“">
<search_summary base_name="/mnt/galaxy/files/000/dataset_582.dat" search_engine="MS-GF+" precursor_mass_type="monoisotopic" fragment_mass_type="monoisotopic" out_data_type="" out_data="">

Hope that makes sense
Ira

PR #60, opened 2016-04-28
WIP: Hardklör and Krönik

glormph:
Hardklör and Krönik are 2 tools from I believe Hoopmann/MacCoss labs. I have made Galaxy wrappers for the toolshed on this branch.

Hardklör identifies peptide or protein-like features in mass spectra, while Krönik filters, validates, and summarizes peptide features identified with Hardklör. Our lab uses them for determining peptide quantities from MS1 spectra.

The toolshed install works when testing with planemo, but I guess it needs to be looked at. I will continue to push commits here (citations need formatting updates, etc


packages/package_hardklor_2_30/tool_dependencies.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<tool_dependency>
+    <package name="hardklor" version="2.30">
+        <install version="1.0">
+            <actions>
+		<action type="download_by_url" target_filename="hardklor.zip">https://drive.google.com/uc?export=download&amp;id=0ByyfOdPaY1xkX0h6RTBnQWNhb3c</action>
```
bgruening: Can you use cargo-port instead? https://github.com/galaxyproject/cargo-port

packages/package_kronik_2_02/tool_dependencies.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0"?>
+<tool_dependency>
+    <package name="kronik" version="2.02">
+        <install version="1.0">
+            <actions>
+		<action type="download_by_url" target_filename="kronik.zip">http://proteome.gs.washington.edu/software/hardklor/public/kronik.zip</action>
```
bgruening: Does not seem to be versioned, maybe also cargo-port? https://github.com/galaxyproject/cargo-port

tools/hardklor/COPYING
```diff
@@ -0,0 +1,121 @@
+Creative Commons Legal Code
```
bgruening: You don't need this strictly as part of the wrappers.

tools/hardklor/hardklor.xml
```diff
@@ -0,0 +1,75 @@
+<tool id="hardklor" name="Hardklor" version="2.30">
+    <description>identification of features from mass spectra</description>
+    <requirements>
+        <requirement type="package" version="2.30">hardklor</requirement>
+    </requirements>
+    <stdio>
+        <exit_code range="1:" level="fatal"/>
+    </stdio>
+    <command>python hardklor.py $ms1_in.extension $ms1_in
```
bgruening: I guess you need to use `$__tool_directory__/hardklor.py`

All params from type `data` and `text` should be quoted.

tools/hardklor/hardklor.xml
```diff
@@ -0,0 +1,75 @@
...
+          <option value="2">MS2</option>
+          <option value="3">MS3</option>
+        </param>
+        <param name="algorithm" type="select" label="Algorithm">
+            <option value="Basic">Basic</option>
+            <option value="Version1">Version 1</option>
+            <option value="Version2" selected="true">Version 2</option>
+        </param>
+        <param name="cdm" type="select" label="Charge determination method">
+            <option value="None">None</option>
```
bgruening: Do you need none, or would `optional=True` be sufficient for you?

tools/hardklor/hardklor.xml
```diff
@@ -0,0 +1,75 @@
...
+        <param name="maxfeat" type="integer" value="12" label="Max nr. of features for mz-window to solve" />
+
+<!--        <param name="sn" type="float" value="1" label="Signal to noise threshold" help="Must be >=0"  />
+        <param name="snWin" type="float" optional="true" label="Dynamic signal to noise window length (m/z)" 
+                help="For spectra with varying noise levels (not for Thermo data)"/>
+        <param name="win" type="float" value="5.25" label="Max width of any set of peaks in a spectrum" /> -->
+    </inputs>
+    
+    <outputs>
+        <data format="hk" name="output" />
```
bgruening: Is there any reason you subcall the tabular output? Simply tabular would be more portable I guess. More recent Galaxy installations do also support to attach column metadata to a tabular output.

tools/hardklor/kronik.xml
```diff
@@ -0,0 +1,35 @@
+<tool id="kronik" name="Kronik" version="2.02">
+    <description>processes Hardklör features to find peptides by chromatographic profiling</description>
+    <requirements>
+        <requirement type="package" version="2.02">kronik</requirement>
+    </requirements>
+    <stdio>
+        <exit_code range="1:" level="fatal" description="Error occurred while running Kronik"/>
+    </stdio>
+    <command>kronik -c $c -d $d -g $g -m $m -n $n -p $p $inputfile $output</command>
```
bgruening: quoting inputfile and outputfile

tools/hardklor/kronik.xml
```diff
@@ -0,0 +1,35 @@
...
+    <inputs>
+	<param name="inputfile" type="data" format="hk" label="Hardklör isotope distributions" />
+        <param name="c" type="integer" value="5" label="Contaminant threshold in minutes" help="Peptides persisting for longer than this value are excluded" />
```
bgruening: identation of params

tools/hardklor/kronik.xml
```diff
@@ -0,0 +1,35 @@
...
+        <param name="d" type="integer" value="3" label="Match tolerance" help="Minimum consecutive scans peptide must be found in" />
+        <param name="g" type="integer" value="1" label="Gap tolerance" help="Maximum amount of scans a peptide can skip to be considered in results" />
+
+        <param name="m" type="integer" value="8000" label="Max mass (Da)" help="Maximum mass of peptide isotope distribution" />
+        <param name="n" type="integer" value="600" label="Min mass (Da)" help="Minimum mass of peptide isotope distribution" />
+        <param name="p" type="integer" value="10" label="Mass accuracy (ppm)" help="Peptides in adjacent scans must differ by less than this value to be called the same." />
+    </inputs>
+    <outputs>
+        <data format="kr" name="output" />
```
bgruening: same as above

bgruening:
@glormph sorry for taking so long to review! This is a very good work!
I added a couple of things, but overall this looks pretty much merge-ready!

Thanks!

jj-umn:
@bgruening   Thanks for the careful review.
@glormph    Did you want to address bgruening last suggestions before merging?   Many thanks for the contribution.

tools/hardklor/hardklor.xml
glormph: Apparently Hardklor accepts a value "None" for the charge determination in its config file.

tools/hardklor/hardklor.xml
glormph: I have only used it for making sure kronik (or other tools) get the right input format, esp when using large histories with a lot of tabular stuff when the dropdowns get very long. But technically it is absolutely tabular. Not sure what the preferred way is for Galaxy though, but I have nothing against using just tabular.

glormph:
Hi and thanks for looking! I will await the cargo-port PR merge and download so that the tests pass when they actually download the files there. Considering the datatype I have no problems with "tabular" instead of specific hardklor or kronik datatypes, but not sure what best practice is there.

Don't worry about the long time, I have plenty to do. :)

tools/hardklor/hardklor.xml
jj-umn: @glormph @bgruening   I tend to agree with Jorrit with subclassing tabular in this case.   There will be fewer help requests when users are guided to make sensible input choices rather than getting the long list of anything tabular.

bgruening: I'm fine with this as well!

bgruening:
@glormph @jj-umn thanks a lot! This is a great addition!

PR #61, opened 2016-05-22
remove proteomic datatypes

bgruening:
These should be included in newer Galaxy releases.

### DIFF
--- a/tools/blast_plus_remote_blastp/tools/repository_dependencies.xml
+++ b/tools/blast_plus_remote_blastp/tools/repository_dependencies.xml
@@ -1,5 +1,4 @@
 <?xml version="1.0"?>
 <repositories>
-  <repository name="proteomics_datatypes" owner="iracooke" />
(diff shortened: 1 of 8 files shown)


bgruening:
@jj-umn can we remove the proteomics datatypes? They should be included in Galaxy now, isn't it?

bgruening:
ping @jj-umn @iracooke

iracooke:
Hi @bgruening this is most likely fine but I am not the maintainer of those repos so I cant be 100% sure.  I believe almost everything in the old proteomics datatypes is now in the main galaxy distribution but in theory it is possible this might break stuff.

bgruening:
I would like enforcing the removal of all datatypes and if anything is missing we should fix it asap in Galaxy.
Just to stick to the best-practice-guidelines.

bgruening:
Let's merge it. All datatypes should be in Galaxy and we should enforce this as community.

PR #62, opened 2016-05-23
Let's go without ö.

bgruening:
This will probably couse some trouble down-stream. Like not rendering in the TS :)

### DIFF
--- a/tools/hardklor/kronik.xml
+++ b/tools/hardklor/kronik.xml
@@ -1,5 +1,5 @@
 <tool id="kronik" name="Kronik" version="2.02">
-    <description>processes Hardkloer features to find peptides by chromatographic profiling</description>
+    <description>processes Hardklor features to find peptides by chromatographic profiling</description>
     <requirements>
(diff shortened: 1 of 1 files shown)


bgruening:
ping @jj-umn

glormph:
Is it only necessary in the <description> tag? Otherwise there are some more umlauts in the help texts. Also, I know +e is the way to go in German, but I think these are supposed to be "hardrock-umlauts" which are not pronounced. Don't ask me why, but what the command line interface does is to just strip them:
http://cruxtoolkit.sourceforge.net/hardklor.html

bgruening:
@glormph I have changed it thanks. I hope this is enough. Help text can be rendered afaik. It's just the name that ends up in all places were I'm not sure about the consequences :)

jj-umn:
@glormph @bgruening   I didn't want to question the alphabet, having been raised with a diacritically-challenged language.

bgruening:
@jj-umn thanks for the merge!

PR #58, opened 2016-04-04
Travis

bgruening:
This PR will introduce enhanced travis testing. As soon as we can get conda packages for our dependencies we can use planemo to actually test our tools. Assuming tests are passing and we merge a PR, we can automatically push these tools to the TTS.

xref: https://github.com/galaxyproject/tools-iuc/issues/700

PR #64, opened 2016-06-07
Fix erroneous module name for Tabular datatype

glormph:
This is a small fix to address the incorrectly specified datatype modules in hardklor's datatypes_conf.xml.

### DIFF
--- a/tools/hardklor/datatypes_conf.xml
+++ b/tools/hardklor/datatypes_conf.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <datatypes>
   <registration>
-    <datatype extension="hk" type="galaxy.datatypes.text:Tabular" subclass="true" />
-    <datatype extension="kr" type="galaxy.datatypes.text:Tabular" subclass="true" />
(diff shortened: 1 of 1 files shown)


bgruening:
@glormph we should move these datatypes into Galaxy core.

bgruening:
Thanks!

glormph:
Merging into core is probably a good idea, although there are currently no tools yet that use them (I guess). Thanks for the merge!

glormph:
Is the toolshed-updating from this repo automatic?

bgruening:
@glormph not yet. We need some consensus one https://github.com/galaxyproteomics/tools-galaxyp/pull/58. I feel very unconformable to push tools automatically to the TS if they are not travis tested. So if we agree on:

- using conda dependencies
- using travis testing
- testing pass

I'm for automatically upload it to the TS. We do this in the IUC repo for example since a few weeks.

bgruening:
I pushed it manually to TTS and TS.

glormph:
Cheers and thanks for the info!

PR #74, opened 2016-11-05
Test data for kronik 2.20 instead of 2.02 which was version in toolshed

glormph:
Hope it fixes the broken test!


bgruening:
Awesome! Thanks!

PR #76, opened 2016-11-06
Mgfformatter conda & testing

bgruening:


bgruening:
This tool is now under constant testing and conda only.

PR #83, opened 2016-11-08
Msgf+ on mzml/fasta pairs

glormph:
Hi, our lab does quite a bit of MSGF+ searches using prefractionated databases matched with prefractionated spectra (by isoelectric point). We use our own MSGF+ tool XML but it would be nice if this could be in the toolshed instead. The tool takes as input either a spectra and a fasta dataset, or a dataset_collection of a pair of spectra/fasta.

There are 2 problems with this PR:
 + Breaks workflows/automation by moving the `params` inside a `conditional`
 + I cannot get the planemo tests to work (but the tool works inside galaxy)

### DIFF
--- a/tools/msgfplus/msgfplus.xml
+++ b/tools/msgfplus/msgfplus.xml
@@ -13,11 +13,18 @@
     </stdio>
     <command>
 <![CDATA[
-        #set $db_name = $d.display_name.replace(".fasta", "") + ".fasta"
-        #set $input_name = $s.display_name
+        #if $msgf_input.intype_selector == "single"
+        #set $db_name = $msgf_input.d.display_name.replace(".fasta", "") + ".fasta"
+        #set $input_name = $msgf_input.s.display_name
+        ln -s '$msgf_input.s' '${input_name}' &&
+        ln -s '$msgf_input.d' '${db_name}' &&
+        #else if $msgf_input.intype_selector == "fractions"
+        #set $db_name = $msgf_input.db_spectra.reverse.display_name.replace(".fasta", "") + ".fasta"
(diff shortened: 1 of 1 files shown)


bgruening:
This is great! Can you elaborate one the two problems? The first one seems to be OK to me. You can use conditionals in workflows.
Do you have a error of the planemo failure?

glormph:
I guess the first one is more if someone has a workflow/API automation and the tool is updated, the workflow will also need to be updated because the param names change.
Second point may depend on my setup. I am unfortunately not so used to `planemo test` yet. I get:

    RunToolException: Error creating a job for these tool inputs - History does not include a dataset of the required format / build, History does not include a dataset of the required format / build

I am not sure if this is actually message that comes after the real error/fail though.

jj-umn:
UMN does have anybody using the existing MS-GF+ tool or workflows depending on it.   Currently, MS-GF+ is being used via SeaarchGUI.  
Unless someone else has a strong objection to this enhancement, I'd vote to merge it.

tjgriff1:
It seems reasonable to merge and use MS-GF+ via SearchGUI

glormph:
It is unlikely we are going to find a lot of people using this tool in a github PR. But yeah, it may break someone's workflow, but updates are at your own risk, test it first, etc I guess.
I see I have to condaify the msgfplus.xml also if I want to run `planemo test` using conda installed msgfplus.

bgruening:
Yes it is possible to have both of all with a bash trick. You ask if this python wrapper exists, if not you use java -jar ... we have this in tools-iuc. Let me know if you need it.

PR #84, opened 2016-11-08
Made msgfplus.xml play with conda

glormph:
Questions:
 + Conda command line is not `java -jar MSGFPlus.jar ...`, but rather `msgf_plus ...` So I guess when installing this you MUST run conda. Is that a deal breaker?
 + Version is yet to be accepted by bioconda PR [#2866](https://github.com/bioconda/bioconda-recipes/pull/2866)


bgruening:
I vote for conda only, we need to get our tools under testing. We simply do not have manpower in galaxyp to maintain TS and conda next to each other.
Thanks @glormph !

glormph:
I would not vote for having it in TS and conda, but if there are people running their own install for some reason there might be incentive to use the bash trick you referred to in #83.

glormph:
Tests pass now, and it is conda only ATM (calls msgf_plus). I do not know how to do the bash trick @bgruening talked about inside Cheetah though, so please elaborate on how to be able to run both `java -jar MSGFPlus.jar` and `msgf_plus`. :)

bgruening:
Let the code speak: https://github.com/galaxyproject/tools-iuc/blob/master/tool_collections/snpeff/snpEff_macros.xml#L13

(I'm fine with conda only.)

bgruening:
You need to rebase this branch against upstream master and remove this line from the blacklist to trigger a real test: https://github.com/galaxyproteomics/tools-galaxyp/blob/master/.tt_blacklist#L10

glormph:
Well that is weird, the one of the tests fail (by diffing the result, I can't inspect the full diff but there are some peptides that have moved around the file). It runs fine on my OSX box.

I don't understand why MSGF+ on Travis yields a different result than here. Shouldn't the run be independent of environment (java in conda)?

bgruening:
Can you test locally with the conda package? Maybe there is some randomness in this search engine?

glormph:
I'll try another box (Ubuntu). Maybe randomness from some system component that differs or something.

glormph:
Meh, now planemo complains when running test:

`Collecting PyYAML==3.11 (from -r requirements.txt (line 4))
  Downloading https://wheels.galaxyproject.org/packages/PyYAML-3.11.tar.gz (138kB)
    100% |████████████████████████████████| 143kB 792kB/s
    Complete output from command python setup.py egg_info:
    running egg_info
    creating pip-egg-info/PyYAML.egg-info
    writing pip-egg-info/PyYAML.egg-info/PKG-INFO
    writing top-level names to pip-egg-info/PyYAML.egg-info/top_level.txt
    writing dependency_links to pip-egg-info/PyYAML.egg-info/dependency_links.txt
    writing manifest file 'pip-egg-info/PyYAML.egg-info/SOURCES.txt'
    warning: manifest_maker: standard file '-c' not found

    error: package directory 'lib3/yaml' does not exist`

glormph:
Ok, I tested again, fresh Ubuntu 16 VM, git checkout my repo (glormph/tools-galaxyp), install msgf_plus from conda (py2.7), planemo in a virtenv (py2.7). Both tests pass. Same if I checkout the galaxyproteomics/tools-galaxyp one and fetch this PR.

What is travis running the tests on?

bgruening:
Some Ubuntu afaik.
You can use planemo for installing the correct dependency for you `planemo conda_install path_to/tool.xml`.

glormph:
I tried that on my dev laptop, used the same Galaxy release, etc. Still passed.

Restarting the test didn't do it. I'll go and try it on a Ubuntu/precise box somewhere I think. Looks like the travis planemo runs on that. That is 12.04, I've only tested on 16.04. I doubt that is the problem though but we'll see.

bgruening:
@glormph are you using always the conda packages? These should return the same results on every machine!
You can simply fix the test with `compare="contains"`.

glormph:
We could do that, but I want to know why I get different outputs on travis vs other machines.

bgruening:
It needs to be the changed lines * 2. One remove, one added.
Or you can use `contains` as comparison method and just check if a certain element is included.
Maybe `sim_size` does also work, this just compares the filesize of both files.

glormph:
Thanks, there is also a SNIP in the test output so I am not sure how many lines differ. Could be a lot. I'll try with sim_size.

bgruening:
But why simsize with a large delta? We expect the same content, just in a different order isn't it?

glormph:
Not sure, I cannot see the full diff in travis. I speced 1000 (assume it's bytes). The output file is supposed to be around 240kB.

glormph:
Alright, 20bytes seems to be enough.

glormph:
Tests pass on the previous commit, but when running it the galaxy log says somewhere "dependency msgf_plus not found". Is that a problem?

bgruening:
Let me see the final result, I could not find the error you mentioned.

glormph:
In the travis test it is here: https://travis-ci.org/galaxyproteomics/tools-galaxyp/jobs/181891494#L2501

But in contrast to my local planemo, it does not complain and finds the dep, so I guess we are in the clear! I did NOT specify --conda_dependency_resolution on my local planemo, assume that was the problem.

bgruening:
More and more tools under testing this is awesome! Thanks a bunch!

PR #93, opened 2016-12-07
Percolator toolshed wrapper with conda

glormph:
This PR:
  + moves to conda dependency management from toolshed installs (is automagic)
  + removes the .py wrapper scripts
  + removes fido (not installed in conda and supposedly not used a lot anymore, but if someone wants to use it, we could possibly put it back)

Do we keep the package_percolator_2_10 for backwards compatibility? AFAIAC we can remove it.


bgruening:
Please remove the tool_dependecy, than it's clear for the user that he should enable conda. Thanks!

tools/percolator/percolator.xml
```diff
@@ -1,84 +1,83 @@
...
-		</change_format>
-	</data>
+		    <change_format>
+			    <when input="output_type" value="tsv" format="tsv" />
+		    </change_format>
+	    </data>
     </outputs>
     <tests>
         <test>
             <param name="input" value="percolatorTab" />
```
bgruening: This test does not have an output?

bgruening:
@glormph this looks really nice. Do you have time to clean them a little bit up and fix the tests?

tools/percolator/msgf2pin.xml
```diff
@@ -1,14 +1,29 @@
...
+    <command>
+             msgf2pin -o $pinout -e $enzyme
+             #if $matches
+             -m $matches
+             #end if
+             #if $aafreq
+             -a
+             #end if
+             #if $ptm
+             -b
```
bgruening: @glormph there is a better way for booleans in Galaxy.
please add `truevalue="-b" falsevalue=""` to your param and here in the command just $ptm.

This will do the trick, no need to the if/else stuff.

glormph:
Oh neat trick! Im kind of hopeless with the indents in XML/Cheetah though. Do we have a best practice? 4 spaces like python?

bgruening:
Yes 4 spaces like in python. But it is not that important at the moment, I will just fix them when I see it. The IUC recommends 4 spaces.

glormph:
I'll fix the tests too

bgruening:
Wow, nice one! Thanks a lot!

glormph:
Awesome, thanks again for the help and merges Björn!

bgruening:
Sure thing! If you like and want to have the full GalaxyP exposure, the last step is to add your tool to the GalaxyP Docker flavor: https://github.com/bgruening/docker-galaxyp/blob/master/galaxyp.yaml

:)

glormph:
No problem https://github.com/bgruening/docker-galaxyp/pull/7

PR #94, opened 2016-12-12
Update Bumbershoot

chambm:
Update Bumbershoot to r10185
Switch Bumbershoot tools to use Conda for dependency resolution
Delete old Bumbershoot-related packages


chambm:
@bgruening Is it ok to abandon and delete the TS package resolution system?

packages/package_bumbershoot_3_0_8738/.shed.yml
```diff
@@ -1,11 +0,0 @@
-categories:
-- Tool Dependency Packages
-description: Contains a tool dependency definition that downloads the Bumbershoot tools.
-long_description: |
```
bgruening: @chambm can you leave the packages please. Maybe we need to fix some older version. It is perfectly fine to remove the tool_dependencyies.xml file and completely go with conda-only, but I would leave the package for the moment.

chambm: In the case of Bumbershoot, I don't think these packages were even working for dependency resolution. They never created the proper env.sh for me, and now the tarball they pointed to is gone since I neglected to pin it on TeamCity. The Conda tarball is pinned.

bgruening:
Moreover, please remove these lines to trigger the travis testng: https://github.com/galaxyproteomics/tools-galaxyp/blob/master/.tt_blacklist#L1

bgruening:
Thanks!!!!

packages/package_bumbershoot_3_0_8738/.shed.yml
bgruening: Oo, don't tell anybody ...

chambm: :+1: So OK to delete the packages? I think that actually applies for package_proteowizard too which will be a separate PR.

bgruening: I would keep them for historical reasons. But if you are sure they never worked ... you decide ;)

bgruening: Don't forget to increase the tool version number. Need to go to bed, see you tm.

chambm:
Re: increasing tool version. You mean the version in the <tool> tag for each wrapper? I was planning to fix those when I move these over to the main toolshed. Although now that they don't use tool_dependencies.xml, I wouldn't need to change anything in this repo to do that move. So yeah, I'll set a version.

chambm:
Doh, I've always done `planemo lint`, not `planemo shed_lint`. Learn something new every day...

chambm:
@bgruening Is the Galaxy that Travis uses configured to resolve with Conda? Because It seems the Conda package downloads, but Galaxy can't find it. For my local Galaxy instance, the Conda resolver works. I didn't have to change the default configs at all (for my 16.07 instance on Jetstream)

bgruening:
Yes it is: https://github.com/galaxyproteomics/tools-galaxyp/blob/master/.travis.yml#L43
You are not running latest planemo correct? Maybe it is something with the latest version from yesterday. I need to look at it.

chambm:
I wasn't testing with planemo, I was testing a real TS installation.

tools/bumbershoot/macros.xml
```diff
@@ -2,7 +2,7 @@
     <token name="@VERSION@">3.0.10185</token>
     <xml name="requirements">
         <requirements>
-            <requirement type="package" version="3.0.10185">bumbershoot</requirement>
+            <requirement type="package" version="3_0_10185">bumbershoot</requirement>
```
bgruening: nice catch!

bgruening: Now just fix the version number and everything will work :) ` 3_0_10158`

chambm: DOH! And even with that typo, my toolshed install worked with a clean instance of Jetstream Galaxy... I guess it doesn't require exact version matching. :)

bgruening:
Do you know what this is? 
```
[UnimodData::ctor] error parsing term "UNIMOD:454": [UnimodData::getBrickFormula] unknown element or brick "HexN"
|  Error: boost::filesystem::current_path: No such file or directory
```

bgruening:
This tool is working great: https://travis-ci.org/galaxyproteomics/tools-galaxyp/jobs/183642593#L785

chambm:
> Do you know what this is?

Working on reproducing it locally. I expect it has something to do with the execution context for the test. I have been able to run these tests with Planemo before, but that was not in a container or on Travis.

bgruening:
Yeah! What was it Matt?

chambm:
There was a bug with running idpQonvert when it was run via the shell's PATH resolution rather than an absolute or relative path from the current directory.

bgruening:
Now we have it under testing and will avoid this in the future!
Thanks again!

PR #107, opened 2016-12-16
uniprotxml_downloader retrieve XML or fasta

jj-umn:


tools/uniprotxml_downloader/uniprotxml_downloader.xml
```diff
@@ -101,6 +101,7 @@ python $__tool_directory__/uniprotxml_downloader.py -v
         <test>
             <param name="input_choice" value="history"/>
             <param name="taxon_file" value="Helicobacter_strains.tsv"/>
+            <param name="column" value="1"/>
```
jj-umn: I'll try that, but I looked at galaxy tools and test/functional/tools/column_param.xml and they all use a numeral for the test value of a data_column param.

PR #111, opened 2017-01-11
Added tsv building functionality to msgf plus tool

glormph:
This PR aims to add an extra option to the MSGF+ tool.

It is already capable of creating mzIdentML output, but when checking the box for tsv output, the command line also runs (effectively via conda):
`java -cp MSGFPlus.jar edu.ucsd.msjava.ui.MzIDToTsv -i infile.mzid -o outfile.tsv`

This should result in a new dataset which is a tab separated file alongside the mzIdentML containing largely the same information. It is not of mzTab format or a new tool, the conversion to tsv is done by MSGFPlus itself.

I am having some problems testing locally in planemo (seem to get a HTTP400 BAD when planemo tries to upload the input files), so I am hoping the tests are actually run here and don't expect them to pass immediately.


glormph:
Alright, the linter does not like the `<environment variable name="LC_ALL">` element. I have not added this and don't know if it is necessary but it has been in there for a while nonetheless. I'll try removing it but am not sure if it is required. Also we use conda now, maybe the element has something to do with the old install package.

glormph:
I get the exact same error in local testing as displayed here.
https://travis-ci.org/galaxyproteomics/tools-galaxyp/jobs/190880744#L2211

I'm a bit confused, because the error persists also when running tests on master branch, but we did not encounter it when I had a previous PR with spectra/db pairs (which has been merged into master).

bgruening:
Not sure if I can look at it today, but tm morning for sure. Thanks @glormph

glormph:
No worries, I'm not in a hurry. :)

bgruening:
The problem is that something happend during the upload of the test data. I'm not sure what, but what over is uploaded there is no mzml file in the history Galaxy could use for the test :(

glormph:
Aha, the tests have incorrect defintion of the nested field names for the input files. They should use a `|` instead of a `.` -- I'll fix.

glormph:
By the way, there is still a `tool_dependencies.xml` file in this repo. That should be removed since it is conda now, correct? I can do that in this PR or open a new one, just tell me.

bgruening:
@glormph can you remove the `msgf_input` in the test, you don't need the nesting there, afaik.
Removing the tool_dependencies.xml file sounds good.

glormph:
Really? But what if there is another conditional with the same inner names but another outer (say `blabla_input|s` )?

bgruening:
Have a look at more complicated tools: https://github.com/galaxyproject/tools-devteam/blob/master/tools/bowtie_wrappers/bowtie_wrapper.xml#L572

glormph:
Ah yes, I saw this now:
https://github.com/galaxyproject/galaxy/blob/dev/test/functional/tools/collection_paired_test.xml

tools/msgfplus/msgfplus.xml
```diff
@@ -56,6 +55,13 @@
             -maxCharge $advanced.maxCharge
             -n $advanced.n
             -addFeatures $advanced.addFeatures
+	#if $tsvcheck
+	    &&
+	    msgf_plus
+	        edu.ucsd.msjava.ui.MzIDToTsv
+	        -i "$output_name"
```
bgruening: single-quotes are preferred

glormph: What is this, python? ;)

bgruening: No, nothing to do with Python. The reational behind it is, that on bash the `"` are interpreted. So assuming someone bad includes some strange characters in your text-param or in the file-system path via filenames or what ever, bash will try to interpret it. Single quotes don't do this and hence are more safe.
This should work for everything in cheetah, it's just a bashism-

glormph: Ok got it and makes sense! Bash always confuses me though, not using it enough so I forget the details. (the python part was a joke)

bgruening:
Ha, I missed this! It works awesome! It's under testing!

PR #114, opened 2017-01-13
Feature/customprodb

chambm:
Add customProDB and its data annotation manager. Still waiting on bioconda's R support so I can make this a bioconda package!


bgruening:
@chambm what are you waiting for in BioConda? R3.3.2?

chambm:
@bgruening Waiting on a resolution to https://github.com/bioconda/bioconda-recipes/pull/2434

bgruening:
new one here: https://github.com/bioconda/bioconda-recipes/pull/3477

tools/bumbershoot/custom_pro_db/customProDB.R
```diff
@@ -0,0 +1,113 @@
...
+
+options('useFancyQuotes' = FALSE)
+
+suppressPackageStartupMessages(library("optparse"))
+suppressPackageStartupMessages(library("RGalaxy"))
+
+
+option_list <- list()
+
+option_list$bam <- make_option('--bam', type='character')
```
bgruening: @chambm if you use <configfile> in Galaxy you can just use this script inside the XML, without the need to care about parameter parsing. This will reduce the code a lot.

bgruening:
@chambm why is this under bumbershoot? It should be in its own directory under tools, isn't it?

tools/bumbershoot/custom_pro_db_annotation_data_manager/data_manager/customProDB_annotation.xml
```diff
@@ -0,0 +1,38 @@
+<tool id="custom_pro_db_annotation_data_manager" name="CustomProDB Annotation" tool_type="manage_data" version="0.0.1">
+    <description>builder</description>
+    <command interpreter="Rscript --vanilla">customProDB_annotation.R
```
bgruening: The interpreter tag is deprecated, just put this into your command section.

tools/bumbershoot/custom_pro_db_annotation_data_manager/data_manager/customProDB_annotation.xml
```diff
@@ -0,0 +1,38 @@
...
+      --outputFile "${out_file}"
```
bgruening: single quotes are preferred for everything

tools/bumbershoot/custom_pro_db_annotation_data_manager/data_manager/customProDB_annotation.xml
```diff
@@ -0,0 +1,38 @@
...
+      2&gt;&amp;1
+    </command>
+    <stdio>
+        <exit_code range=":-1" />
+        <exit_code range="1:" />
+    </stdio>
+    <inputs>
+        <param type="genomebuild" name="dbkey" value="" label="UCSC dbKey for reference genome" />
+        <param type="text" name="dbsnp" value="" label="dbSNP identifier currently available from UCSC" help="e.g. 'snp142'" />
+        <param type="boolean" name="cosmic" value="" label="Annotate somatic SNPs from COSMIC (human only)" />
```
bgruening: use `truevalie` and `falsevalue` to save code in the command section.

tools/bumbershoot/custom_pro_db_annotation_data_manager/data_manager/customProDB_annotation.xml
```diff
@@ -0,0 +1,38 @@
...
+    <command interpreter="Rscript --vanilla">customProDB_annotation.R
+      --outputFile "${out_file}"
+      --dbkey "${dbkey}"
+      --dbsnp "${dbsnp}"
+      #if $cosmic
+            --cosmic true
+      #end if
+      --dbkey_description ${ dbkey.get_display_text() }
+      2&gt;&amp;1
+    </command>
```
bgruening: CDATA is also missing

tools/bumbershoot/custom_pro_db_annotation_data_manager/data_manager_conf.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0"?>
+<data_managers>
```
bgruening: This one should be in a general data_manager directory in the root folder.

tools/bumbershoot/custom_pro_db/COPYING
```diff
@@ -0,0 +1,121 @@
+Creative Commons Legal Code
+
+CC0 1.0 Universal
```
bgruening: is this really needed?

bgruening:
@chambm here you go: https://anaconda.org/bioconda/bioconductor-customprodb

tools/bumbershoot/custom_pro_db_annotation_data_manager/data_manager_conf.xml
chambm: What is the root folder? tools/bumbershoot/../..?

tools/bumbershoot/custom_pro_db_annotation_data_manager/data_manager/customProDB_annotation.xml
chambm: I'm pretty sure I tried that a while ago and it broke because it couldn't find Rscript. That was before conda though, I'll try again with conda.

tools/bumbershoot/custom_pro_db/customProDB.R
chambm: It's easier to test the R wrapper in RStudio with its own parameter parsing. Also, RGalaxy can autogenerate these wrappers from roxygen2 comments. Although I did have to customize it quite a bit.

chambm:
It's under bumbershoot because I couldn't get the author to come up with a name for this suite of tools: PSM2SAM (proBAMr), CustomProDB, and the annotation that is shared between them. :) PSM2SAM will come as a separate commit. I could put them in the main tools directory, but in my mind they do logically go together (especially the annotation data manager and the two tools that use it).

tools/bumbershoot/custom_pro_db_annotation_data_manager/data_manager_conf.xml
bgruening: Like here: https://github.com/galaxyproject/tools-iuc/tree/master/data_managers

chambm:
@bgruening Any idea why this is failing? It works for me locally (`planemo test --conda_dependency_resolution`)

bgruening:
No but lets upgrade the test environment.

bgruening:
All I see is that this `+    <exit_code range="1:" level="fatal" description="Job Failed" />
` get trigered as it seems, but not sure why.

bgruening:
This error is unrelated and will hopefully be fixed soon upstream.

bgruening:
Still failing due to the unknown error.

jj-umn:
@chambm @bgruening   
Do we need a bioconda recipe for RGalaxy and a requirement for it in  customProDB.xml  since RGalaxy is used in the Rscript: customProDB.R

bgruening:
Afaik this is a converter from R to Galaxy-XML and this script is just needed for boostrapping the XML. @chambm am I right?

chambm:
JJ may be right. I use these commands to install on my instance:

bgruening:
@chambm and I thought this RGalaxy will generate an XML and is only a build-time dep. It this this is a lot of overhead compared to traditional R Galaxy tools.

chambm:
Compared to the dozens of other bioconductor packages that customProDB depends on, RGalaxy seems like a pretty minor addition. This particular package only has one entry point, but for a package which needed to expose many functions, RGalaxy would save someone a lot of handwritten wrapper code, right?

bgruening:
In contrast du a R wrapper that uses configfiles, like this one: https://github.com/galaxyproject/tools-iuc/blob/master/tools/vegan/vegan_rarefaction/vegan_rarefaction.xml#L19

RGalaxy adds a lot of code. But maybe I'm missing something.

chambm:
I find it more difficult to read (and thus maintain) that configfiles script. The RGalaxy code is autogenerated, so the amount of code seems not so important to me, as long as it is maintainable.

bgruening:
In general I would agree if we would talk about a language with a focus on readability, we talk about R ;)
You add a new layer of complexity, during runtime, which is bringing in new risks, a lot of new code (not only the auto-generated, also the package) and a conda package we need to create and maintain ;). Maintenance also means that I need to learn RGalaxy to update these wrappers. I don't see a benefit that outweighs the small disadvantage about readability.

I'm fine if you want to merge it that way, as a general best-practice I would argue against it.

Do I see it correct that this package is not really updated since 2014?

jj-umn:
@chambm @bgruening   I added a bioconda recipe for bioconductor-rgalaxy.  My local customProDB recipe I added bioconductor-rgalaxy,r-optparse,r-json

chambm:
@bgruening We got further?
```
UnsatisfiableError: The following specifications were found to be in conflict:
  - bioconductor-biobase 2.32.0 0
  - bioconductor-summarizedexperiment -> r-matrix 1.2_6
```

chambm:
Hmm, so should this have been automatically uploaded to the toolshed?

PR #120, opened 2017-01-19
Update blast_plus_remote_blastp to 2.6.0

jj-umn:
Need newer version to use -remote option to NCBI.  blast must be
compiled with GNUTLS to connect via https.


bgruening:
@jj-umn can we add a shed.yml and tests to this tool if we want to maintain this?

PR #122, opened 2017-01-24
SearchGUI fix isotope params

jj-umn:
SearchGUI  isotope options moved to compomics identification params
“Spectrum matching parameters”


bgruening:
@jj-umn we need to increase the version number, aren't we?

PR #123, opened 2017-01-31
Add uniqueness criterion parameter to "Merge FASTA" tool

chambm:
To allow merging on unique accession and allowing unique sequences to remain (e.g. identical protein isoforms, gene paralogs/homologs); also enforce accession (header) uniqueness in both modes


chambm:
Does anyone see a problem with enforcing uniqueness of the header lines even in the "unique sequences" mode? I wasn't able to think of a legitimate case for when the header lines would be the same but the sequences different. But my FASTA experience is almost entirely from the proteomics world, so I could be missing a genomics case.

bgruening:
A header should be unique I think. But I'm worried if you create yourself a FASTA file, give it some dummy names, merge it and loose sequences just because your name was not unique. Can/Should we add an option to rename the header automatically - with a suffix?

bgruening:
What about the following, we leave this PR as it is but we include more output to the stdout, so a user can ready what was happening. In the help section we encourage people to use fasta-to-tabular and tabular-to-fasta for more complicated use-cases?

iracooke:
@bgruening you read my mind.  I was just writing a comment to that effect.  Write a summary of what the tool did to stdout. Also my 2c is that I think uniqueness should on accession not header.  Many tools require unique accessions and completely ignore the headers.

jj-umn:
Ira makes a good point about the "uniqueness should on accession not header".  Compomics goes to great lengths to pick the accession out of the header: [https://github.com/compomics/compomics-utilities/blob/master/src/main/java/com/compomics/util/protein/Header.java](https://github.com/compomics/compomics-utilities/blob/master/src/main/java/com/compomics/util/protein/Header.java)

chambm:
Accession is part of the header, of course. The FASTA format only defines that there is a header line, it doesn't say how it must be delimited into an accession, description, and god knows what else. That's why tools like Compomics and Mascot have to go to great lengths to parse them to get just the accession. However, my understanding is that that parsing is not really intended for uniqueness detection, but for better presentation of the data and cross-referencing to other resources.

@iracooke , @jj-umn For the most part, unique accessions and unique headers should be the same thing. There are very few cases where the non-accession part will change while the accession part stays the same (and both entries should be kept in the output). Unfortunately, the first and only case that comes to mind is customProDB, which inserts the RPKM value of the protein's transcript into the FASTA header. That makes the header different between different samples even for identical variant proteins. 

@PratikDJagtap , I guess what you're asking for is to accommodate the customProDB case where we have multiple variant proteins that may be identical in sequence, but we suffix their accessions with the sample they were found in. You want to keep all those sample names in the header but only keep one instance of the protein sequence, right? I don't think that's necessary except to optimize searching speed. The protein assemblers should take care of that. For example, if there are identical variant proteins from pre-pro-B and pro-B, we should see a protein group like: `ABC123_pre-pro-B, ABC123_pro-B`. In any case, doing what you ask would require 2 scans through all of the input FASTAs and would be a bigger restructuring of the current script. The current modification is minimal. But if that behavior is really desired, we can talk about it for another PR.

@bgruening It makes perfect sense to add duplicate notifications to stdout. I will do that. Can I test stdout in the tool wrapper _without_ making it a separate output (e.g. mpileup log, tophat log)?

SamGG:
Hi! Just a question concerning memory. When requiring sequence uniqueness, the dictionary stores the entire set of merged sequences. I don't know if this could lead to a memory issue.

chambm:
Eh, you're right. We should store a hash instead.

SamGG:
This was just a question. If the FASTA output fits in memory, your current code is OK.
Hash is typically used in UniPARC. Computing hash sounds time consuming but might prevent memory failure. May be you should benchmark how much time hash computing will take. For example a loop of 10k hash on a defined-once string of 500 random characters.

iracooke:
@chambm @jj-umn Sorry I raised the ugly head of accessions.  As you say a unique header line works almost all the time.  If you were to "parse" accessions I'd say the only sensible approach is to treat everything up to the first whitespace as an accession.  Almost every fasta file I've seen has this structure.  What comp-omics does is to try and sub-parse what I would consider the fasta accession into other database specific accessions (uniprot, genbank etc) which is definitely not a good idea for this tool.

SamGG:
@iracooke Totally agree. Accession stops at whitespace or line end if there is no description.

PR #124, opened 2017-02-01
fixed incorrect parameter format for mzIdentML output

jvolkening:
Calling as -OutputFormat 'mzid' throws an error (should be 'mzIdentML')

### DIFF
--- a/tools/bumbershoot/myrimatch/myrimatch.xml
+++ b/tools/bumbershoot/myrimatch/myrimatch.xml
@@ -13,7 +13,7 @@
     <command>
 <![CDATA[
         #set $db_name = $ProteinDatabase.display_name.replace(".fasta", "") + ".fasta"
-        #if $OutputFormat.value == "mzid"
+        #if $OutputFormat.value == "mzIdentML"
             #set $output_ext="mzid"
         #else
             #set $output_ext="pepXML"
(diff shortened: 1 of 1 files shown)


jj-umn:
@jvolkening  Looks like the tests failed due to slight unavoidable differences in the outputs.  You might want to try using assert_contents rather than trying to match a file: [https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-tests-test-output-assert-contents]

jvolkening:
I updated the tests (apparently unrelated to the changes I made). It seems that "<is_valid_xml>" would be appropriate but there seems to be a bug in core galaxy (galaxyproject/galaxy/#3530) so I commented that check out for now.

I also added an option to change the filter string. By default myrimatch tries to run peak-picking on the input, which may not be appropriate in some cases (e.g. if the data is already centroided). The changes keep the default but allow the user to modify. Wasn't sure if this should be in a separate PR or not.

chambm:
@jvolkening If a spectrum is already centroided, peakPicking will leave it alone. In fact even if you take out peakPicking, IIRC MyriMatch will add it back anyway to ensure that no profile spectra sneak through.

chambm:
FYI those (pepXML) tests were pretty good the way they were. They just didn't anticipate the change in MyriMatch in version. You can just change the lines_diff from 12 to 14 and that should cover changes to version. And still make sure that the rest of the file is unchanged. I.e. no parameter specifications change between versions (of either MM or the wrapper).

jvolkening:
@chambm Thanks - I wasn't aware of the peakPicking behavior. I just saw the warnings myrimatch was giving (about not being on Windows) and was concerned it would mangle my peaks. I'm happy to revert 9bb77d9f8658983d5ab26425504c8234d54e0c3c in the PR.

Regarding the testing, I was just trying to respond to @jj-umn above - it wasn't related to my original bugfix. I can modify to whatever the consensus opinion is.

jj-umn:
@jvolkening  You can adjust your test as you like.  I just like using assertions to focus on the actual output differences based on parameter settings.  @chambm When you are satisfied, do you want to merge?

bgruening:
Thanks all!

PR #126, opened 2017-02-03
Accession parser regex and memory usage improvements

chambm:
Add accession parser regex, which defaults to counting everything in the header before the first space as the accession
Reduce memory usage (by about a factor of four) by hashing seen_sequences; the reduced memory allocation also improves speed a bit


chambm:
@SamGG Hashing actually speeds up large merges (at least on my Windows tests).
@iracooke The default is now to stop parsing accession at the first space, but it's pretty to easy to customize (that's why I used [^ ] instead of \S+, although I suppose adding \t in there wouldn't hurt).

SamGG:
Well done! Thanks for your feedback.

PR #127, opened 2017-02-06
tweaked Travis config to not try Galaxy upload for forks

jvolkening:
The last step in the Travis CI config automatically tries to update the Galaxy TS if a build passes, but this isn't possible for forked repos and results in a failed build. This complicates testing of bug fixes, etc prior to submitting a PR.

This PR adds an additional check to make sure the build is under the main repo before trying the Galaxy update. Obviously I can't check that it works for you -- just that it prevents CI failure in my fork.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,9 @@ script:
 
 after_success:
   - |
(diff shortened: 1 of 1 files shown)


bgruening:
@jvolkening never work on the master branch :)
If you work on feature branches you can simply avoid this problem.

jvolkening:
Fair enough, thanks. Will close PR.

PR #130, opened 2017-02-08
Update Unipept with batched API requests and D3 JSON

jj-umn:
Relies on d3_hierarchy Json subclass being added to
config/datatypes_conf.xml.sample for the Json output
https://github.com/galaxyproject/galaxy/pull/3580

### DIFF
--- a/tools/unipept/unipept.xml
+++ b/tools/unipept/unipept.xml
@@ -1,27 +1,31 @@
-<tool id="unipept" name="Unipept" version="1.1.0">
+<tool id="unipept" name="Unipept" version="2.0.0">
     <description>retrieve taxonomy for peptides</description>
     <macros>
         <xml name="equate_il">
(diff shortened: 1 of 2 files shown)


tools/unipept/unipept.xml
```diff
@@ -1,27 +1,31 @@
...
-            </param >
+            </param>
             <param name="allfields" type="boolean" truevalue="-A" falsevalue="" checked="false" label="allfields" >
                 <help>include fields for most specific taxonomic classification: taxon_rank,taxon_id,taxon_name before lineage</help>
-            </param >
+            </param>
         </xml>
     </macros>
     <requirements>
-        <requirement type="package" version="2.7.12">python</requirement>
```
bgruening: @jj-umn why this? I think we should annotate it this way.

PR #132, opened 2017-02-15
Remove deprecated command interpreter attribute

jj-umn:


jj-umn:
@bgruening This didn't get pushed to the toolshed.  Can you do that?  
planemo toolshed updates are still not working for me.

Issue #131, opened 2017-02-13
OpenMS status


jvolkening:
Can anyone comment on the status of OpenMS in Galaxy? The TS package is owned by galaxyp but is several years old and I don't see it in the github repo. @bgruening has it in [his own tool repo](https://github.com/bgruening/galaxytools) and it looks active, but it is in the `.tt_blacklist`. Are either of these options suitable for a production server?

bgruening:
The TTS package is based on the wrappers in https://github.com/bgruening/galaxytools and there is also the latest version. Which should work, it is for us - at least we are testing it intensively, right now.

These wrappers are auto generated and do not have tests, if we create tests, which I guess would be good, these are manual adjustments that needs to be redone after each upgrade ... so much work. I guess 2-4 tests would be great to at least tests the package and so one.

So in a summery, I have a conda package, a Docker container and the wrappers should work. I will migrate them into the GalaxyP repo as soon as our inhouse testing is done (help appreciated) - but they will be special - auto generated.

jvolkening:
Thanks for the explanation. I'd like to install the latest version from your github repo -- is there a way to install it within the framework of the Galaxy tool manager interface, or do I need to manually copy the files in?

I agree that (1) writing tests for every single tool (and maintaining them) would be a large undertaking, and that (2) it would be good to include at least a couple of tests to check for general problems with the installation. Could this be done in a "quasi-tool" whose only function was to perform the tests and with the "hidden" attribute set? This might avoid problems during updates based on the autogeneration.

blankclemens:
Why a quasi tool? I guess it would be reasonable to just choose one of the tools and add a test case, making sure the OpenMS installation is in general working.

Right now you would have to copy the files manually, but it is planned to move the wrappers to Galaxy-P in near future. There is an example `tool.conf` included to ease the pain.

jvolkening:
> Why a quasi tool? I guess it would be reasonable to just choose one of the tools and add a test case, making sure the OpenMS installation is in general working.

My thinking was that if you put the tests into one of the tools, you would need to manually insert the test block into the wrapper XML each time it was re-generated from the CTD2Galaxy script. If you put it in an independent XML file it wouldn't get clobbered each time CTD2Galaxy was run but could exist side-by-side.

blankclemens:
Good point, thanks!

bgruening:
[...]

PR #134, opened 2017-02-21
Add OpenMS Wrappers

blankclemens:
Adding the OpenMS wrappers from https://github.com/bgruening/galaxytools as mentioned in #131.

Some minor changes:

- Removed PepNovoAdapter since it needs additional care (cexpects a folder with `-model_directory`)
- added a small test case for `DecoyDatabase`

These wrappers are automatically generated with some manual adjustments, further information can be found in the `readme.md`.

All tools that I know of not working without further work are listed in `SKIP_TOOLS_FILE.txt`.
There might be additional wrappers that do not work and testing would be highly appreciated.


bgruening:
@blankclemens we need a shed.yml file here. Preferably with the suite stuff.

tools/openms/AccurateMassSearch.xml
```diff
@@ -0,0 +1,107 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!--This is a configuration file for the integration of a tools into Galaxy (https://galaxyproject.org/). This file was automatically generated using CTD2Galaxy.-->
+<!--Proposed Tool Section: [Utilities]-->
+<tool id="AccurateMassSearch" name="AccurateMassSearch" version="2.1.0">
```
bgruening: should we prefix all ids, with openms_ ?

tools/openms/.shed.yml
```diff
@@ -0,0 +1,12 @@
+name: openms
+owner: galaxyp
```
bgruening: More like this: https://github.com/galaxyproject/tools-iuc/blob/master/tools/gemini/.shed.yml
So that every tool ends up in its own toolshed package.

bgruening: Also here is something wrong with the lineendings.

blankclemens: Hmm, I thought about that but I couldn't see a point why we would want that? Fine for me though.

bgruening: We want to give people the opportunity to just pick a few tools put of 120. We still get one meta-package that will contains a link to all others if you want to get all.

tools/openms/.shed.yml
```diff
@@ -1,12 +1,22 @@
...
+long_description: |
+  OpenMS is an open-source software C++ library for LC/MS data management and analyses. It offers an infrastructure for the rapid development of mass spectrometry related software.
+
+  https://www.openms.de/
+remote_repository_url: https://github.com/galaxyproteomics/tools-galaxyp/tree/master/tools/openms
+type: unrestricted
+categories:
+- Proteomics
+auto_tool_repositories:
+  name_template: "{{ tool_id }}"
```
bgruening: should we call this `openms_{{ tool_id }}`. Or maybe change the toolids ... a few of them are very general.

tools/openms/.shed.yml
```diff
@@ -0,0 +1,22 @@
...
+  description_template: "Wrapper for the OpenMS suite tool: {{ tool_name }}"
+suite:
+  name: "suite_openms"
+  description: OpenMS Suite for LC/MS data management and analyses
+  long_description: |
+  OpenMS is an open-source software C++ library for LC/MS data management and analyses. It offers an infrastructure for the rapid development of mass spectrometry related software.
```
bgruening: You need an indentation here.

tools/openms/.shed.yml
blankclemens: Do Tool IDs have to be unique? I would agree in changing the package names, but adjusting the tool IDs seems like a pretty random intervention in the automatic wrapper generation.

bgruening: They are always unique, so we don't need to take care about this. But this ID appears in log files and so on ... as more information is included as better. But ok, let's leave it like this. Just mention that we could change the generator to include this prefix.

bgruening:
Ok! Let's merge this. Thanks @TorHou, @blankclemens and @jmchilton for working on this during the years. It will finally hit the TS with conda package and container.

PR #141, opened 2017-03-09
Percolator 3.1

glormph:
This PR does:
 + Upgrade percolator to version 3.1
 + Include option for target decoy method (mixmax v td-competition)
 + Add Xtandem and SQT input converters
 + Subclass percolator out datatype which was encoded in py file earlier
 + In nested-collection tool the user-specified pool-id names are included in the output `display_name`


tools/percolator/datatypes_conf.xml
```diff
@@ -5,7 +5,7 @@
   </datatype_files>
   <registration>
     <datatype extension="percin" type="galaxy.datatypes.data:Text" subclass="true" />
-    <datatype extension="percout" type="galaxy.datatypes.percolator:PercolatorOutXml" />
+    <datatype extension="percout" type="galaxy.datatypes.xml:GenericXml" subclass="true" />
```
bgruening: This should go into the main Galaxy.

bgruening: Can you create a PR and get this datatype into Galaxy?

glormph: Yes that should be no problem. Thanks for merging!

PR #142, opened 2017-03-14
Delete obsolete tool wrapper files

chambm:


chambm:
@bgruening It seems removing .shed.yml from the main msconvert directory, which doesn't have a current tool in it, makes planemo fail to find the tests. Or maybe it's the lack of a tool wrapper that stops it from recursing.

bgruening:
Mh, it can also be that I only changed modified and added files and only test in this case. I do not remember how I implemented it. Let's modify a file.

chambm:
You're right. It was due to `planemo ci_find_repos` not finding any changes. Which I guess is strictly speaking true!

bgruening:
Thanks @chambm!

PR #145, opened 2017-04-06
Future proofed hardklor/kronik to datatypes name change from short to…

glormph:
… more descriptive

Only changed the datatypes. Future work would be to remove the datatypes_conf.xml file but only after the datatypes have landed in a galaxy core release (17.05 probably).


bgruening:
:)!

PR #146, opened 2017-04-07
Percolator collection tools update

glormph:
This PR does 2 things:
  + Updates the nested-collection-creator to accept multiple filetypes by selecting a filetype. This is until discover_datasets collection can get the datatype from the input files.
  + Introduce a new tool which _unrolls_ a nested list (`list:list`) into a single `list`.


tools/percolator/rollout_nested_collection.xml
```diff
@@ -0,0 +1,52 @@
+<tool id="rollout_nested_collection" name="Rollout nested list" version="3.1">
+    <description>into a single long list</description>
```
bgruening: I haven't followed the discussion on gitter, but is this really needed and if so would this not a better tool-collection tool directly in Galaxy?

jmchilton: Yeah... I think this is what flatten does - or pretty close https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/flatten_collection.xml. Can you try to use flatten and if it isn't quite right clarify how and maybe we can options to make it work the way you are hoping to.

glormph: Aha there is a tool! Thought it was going to be in a future version in javascript. I'll have a try with flatten then! Brb!

jmchilton: Opps - sorry if I indicated this required "the missing operations" - I was probably mistaken.

glormph: Nah, think that was about the nested-list creator in this repo. But I didn't think to look further when I suddenly figured I also needed this yesterday evening. Flatten seems to do exactly what I need so I will remove my tool!

bgruening:
Thanks @glormph!

PR #147, opened 2017-04-08
Bugfix for openMS

glormph:
Added quote marks to activation methods in tool cheetah since this contains values with whitespace in almost all options. Without this, the tool crashes.


bgruening:
@blankclemens can we add this as well to the converter? I added quotation once to the converters to all text and data fields, but maybe this is lost now?

However, let's merge this. Thanks @glormph!

blankclemens:
I'ma bit surprised as well, I'll fix that this week.

blankclemens:
@bgruening It wasn't lost, it was just not present in select lists if the default parameter is predefined. Fixed that in the converter and will update the wrappers after Easter.

PR #148, opened 2017-04-18
Add whitespace validation to select lists

blankclemens:
Using the latest CTD2Galaxy to fix #147 for all other tools.


bgruening:
If " " in str($foobar): looks wrong.

blankclemens:
Indeed, but that's how it is done in CTD2Galaxy and used on other places - I just added the expected behavior to select lists as well. It's the result of changing as less as possible to make sure everything works as before.

One might think about changing the converter, but I would postpone such actions for a time when they are necessary anyways - e.g. maybe with a new OpenMS version.

bgruening:
Thanks!

PR #152, opened 2017-04-27
Fix IDFileConverter

blankclemens:
Unfortunately, IDFileConverter detects the input type according to the file extension, which is not working with Galaxy. This is a work around which should work in common cases (e.g. mzid, idxml).


bgruening:
Makes sense! Thanks!

PR #155, opened 2017-04-28
update to proBAMconvert v 1.01

vladie0:


tools/probamconvert/probamconvert.xml
```diff
@@ -46,6 +47,7 @@
                     <option value="drosophila_melanogaster">drosophila_melanogaster</option>
                 </param>
                 <param name="version" type="select" optional="true" label="ensembl release version">
+                	<option value="88">88</option>
```
bgruening: the indentation is wrong here

vladie0: thank you for fixing !

PR #156, opened 2017-04-28
Better sample tracking nested collections

glormph:
This PR:
 + Changes to continuous numbering of the inner datasets in nested lists, for all files belonging to the same "pool" (or sample set, biological set), as determined by regex. So not `inputfn0..2` in each batch, but the second batch will have `inputfn3..5`, if it belongs to the same pool.
 + Adds full display name of datasets to the those `inputfn` datasets, making them `inputfn1_your_spectra.mzML` for easier sample tracking
 + Adds zero-padding to file numberings so discover datasets uses the correct order when discovering the lists. Otherwise `inputfn10` will come before `inputfn5`.

Planemo tests pass locally, but I want to give it a try on my galaxy installation as well via the testtoolshed, will notify here when it has been tested.

### DIFF
--- a/tools/percolator/nested_collection.py
+++ b/tools/percolator/nested_collection.py
@@ -42,10 +42,18 @@ def main():
     parser.add_argument('--galaxy-files', dest='galaxyfiles', nargs='+')
     parser.add_argument('--pool-ids', dest='poolids', nargs='+', default=False)
     args = parser.parse_args()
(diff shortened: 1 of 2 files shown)


bgruening:
Travis is also happy :)
Thanks @glormph!

bgruening:
Let me know when your testing is done.

glormph:
Think it works as I intended now! Thanks for the check!

PR #165, opened 2017-05-19
Allow dash/hyphen/minus in a filename

glormph:
Nested collection creator has a regex to get nested list identifiers. Except that regex doesnt like names with characters other than `[\w.]`, or in other words `[a-zA-Z0-9_.]`. Typically filenames are used, and I added here the `-` hyphen/dash/minus character which is not uncommon and allowed in filenames.

If you can think of any other character which may show up in a filename, please tell or commit.

### DIFF
--- a/tools/percolator/nested_collection.xml
+++ b/tools/percolator/nested_collection.xml
@@ -1,4 +1,4 @@
-<tool id="batched_set_list_creator" name="Create nested list" version="3.2">
+<tool id="batched_set_list_creator" name="Create nested list" version="3.3">
     <description>based on filenames and batch sizes</description>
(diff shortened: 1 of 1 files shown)


bgruening:
Can we not simply take `.`? E.g. `.*`?

glormph:
Maybe that is better but then you will allow all kinds of weird chars too. Now that I think about it, the script creates symlinks with the `display_name`, is that a problem in case certain characters are used?

bgruening:
If you have unicode characters it would probably fail ...  not sure.

bgruening:
Thanks!

PR #166, opened 2017-05-19
Pi db tools

glormph:
Our lab runs a lot of isoelectric point (pI) based sample prefractionation and I've some scripts here that do prefractionation on a pI-predicted file of peptides and produces FASTA output that can be used in database searches.


tools/pi_db_tools/delta_pi_calc.xml
```diff
@@ -0,0 +1,121 @@
+<tool id="calc_delta_pi" name="Add delta pI" version="1.0">
+    <description>to peptide table</description>
+    <command>
+	    python3 $__tool_directory__/peptide_pi_annotator.py -i '$trainingpi' -p '$peptable'
```
bgruening: quotes for the script ...

tools/pi_db_tools/delta_pi_calc.xml
```diff
@@ -0,0 +1,121 @@
...
+	     <param name="fr_width" type="text" label="fraction widths" />
+     </repeat>
+    </inputs>
+    
+    <outputs>
+        <data format="tabular" name="output"/>
+    </outputs>
+    <tests>
+	    <test>
+		    <param name="trainingpi" value="predicted_peptides.txt" />
```
bgruening: indentation is wired here.

tools/pi_db_tools/delta_pi_calc.xml
```diff
@@ -0,0 +1,121 @@
...
+		    '$mod.regex'
+		#end for
+	    #end if
+    </command>
+    
+    <inputs>
+      <param name="trainingpi" type="data" format="tabular" label="Known peptides with annotation of pI" help="First column is sequence, second pI" />
+      <param name="peptable" type="data" format="tabular" label="Peptide table to determine pI shift from" />
+      <repeat name="ignoremods" title="Peptide modification weights to ignore">
+	      <param name="regex" label="Regex to strip modification from peptide table before looking up sequence in predicted pI" type="text" help="Enter e.g. 15.994915 for oxidation, * to ignore all modifications, or a proper regex (however brackets are stripped by galaxy)" />
```
bgruening: Maybe you can provide a link or some examples in the help what a regex is?

tools/pi_db_tools/delta_pi_calc.xml
```diff
@@ -0,0 +1,121 @@
...
+    
+    <outputs>
+        <data format="tabular" name="output"/>
+    </outputs>
+    <tests>
+	    <test>
+		    <param name="trainingpi" value="predicted_peptides.txt" />
+		    <param name="peptable" value="peptable.txt" />
+		    <repeat name="ignoremods">
+			    <param name="regex" value="*" />
```
bgruening: Should we sanitize this a little bit? No white spaces allowed for example?

tools/pi_db_tools/peptide_pi_annotator.py
```diff
@@ -0,0 +1,108 @@
...
+import sys
+import argparse
+
+
+def main():
+    if sys.argv[1:] == []:
+        sys.argv.append('-h')
+    args = parse_commandline()
+    strips = {strip: {'intercept': args.intercepts[i],
+                      'fr_width': args.fr_width[i]}
```
bgruening: can we write this in a more readable way without the comprehension?

tools/pi_db_tools/pi_database_splitter.py
```diff
@@ -0,0 +1,179 @@
+# => python3.3 required!
```
bgruening: Shebang: !/usr/bin/env python

tools/pi_db_tools/peptide_pi_annotator.py
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/python3
```
bgruening: Shebang: !/usr/bin/env python

tools/pi_db_tools/pi_db_split.xml
```diff
@@ -0,0 +1,80 @@
+<tool id="pi_db_split" name="Split peptide database" version="1.0">
+    <description>into pI separated fractions</description>
+    <requirements>
+        <requirement type="package">numpy</requirement>
```
bgruening: You probably want python=3.6 here as well.

tools/pi_db_tools/pi_db_split.xml
```diff
@@ -0,0 +1,80 @@
...
+    </requirements>
+    <command>
+	    mkdir pi_fr_out; cd pi_fr_out;
+	    python3 $__tool_directory__/pi_database_splitter.py -i $pipeptides -p $peptable
```
bgruening: quotes and please use && instead of ;

tools/pi_db_tools/pi_database_splitter.py
glormph: Do I need a shebang when we call it with `python3 script.py` anyway?

bgruening: I would not call it with python3 ... this is up the the environment to provide you the correct python version ... there might be no python3 bin, but still python is using python3.

glormph: Aha, but then we should also just write `$__tool_directory/script.py` instead of with `python` or `python3`.

tools/pi_db_tools/delta_pi_calc.xml
glormph: Oh my, just discovering the `<sanitizer>` tag.

tools/pi_db_tools/delta_pi_calc.xml
```diff
@@ -1,7 +1,10 @@
 <tool id="calc_delta_pi" name="Add delta pI" version="1.0">
+    <requirements>
+        <requirement type="package" version="3.6">python</requirement>
+    </requirements>
     <description>to peptide table</description>
     <command>
-	    python3 $__tool_directory__/peptide_pi_annotator.py -i '$trainingpi' -p '$peptable'
+	    '$__tool_directory__/peptide_pi_annotator.py' -i '$trainingpi' -p '$peptable'
```
bgruening: You need python here as the file is not guaranteed to be executable. Sorry.

tools/pi_db_tools/delta_pi_calc.xml
glormph: So I guess I should add also the brackets and backslash for sanitation of regexes. And the caret and..
Would it be ok to do the same as grep.xml does? I.e. take all from string.printable and remove `&`. And also whitespace then.

glormph:
In answer to your general question, I had not thought about bioconda packaging because in my brain this was "just some small scripts for our pipeline in Galaxy", but that wouldn't actually be too bad.

tools/pi_db_tools/delta_pi_calc.xml
bgruening: Think so yes. I'm not a cracker, I don't have enough evil skills to think of all evil hacks that people can do with this ;)

glormph: No, same here. I guess that's why we have autosanitizing.

bgruening:
@glormph ready for merge or do you want to explore the conda way?

glormph:
Merge I think. Conda will have to wait until later. :)

bgruening:
And merged! Thanks!

glormph:
Awesome, thanks once again for the rigorous checkup!

PR #169, opened 2017-07-07
OpenMS 2.2

blankclemens:
Updating the wrapper for OpenMS 2.2.

As always, there might be new issues since this is mostly automatically generated, feel free to complain or fix. In the later case please add the necessary fixes to the readme.


blankclemens:
The failed test comes from wrong documentation links, see https://github.com/OpenMS/OpenMS/issues/2433

bgruening:
Thanks a lot.

PR #171, opened 2017-07-13
A fasta database validation tool

caleb-easterly:
Uses the Compomics FASTA header schema (from [https://github.com/compomics/compomics-utilities/blob/master/src/main/java/com/compomics/util/protein/Header.java](url)) and a series of other checks to filter out the bad parts of any FASTA database.


tools/validate_fasta_database/.idea/compiler.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bgruening: this file is not needed?

tools/validate_fasta_database/.idea/libraries/Maven__junit_junit_4_12.xml
```diff
@@ -0,0 +1,13 @@
+<component name="libraryTable">
+  <library name="Maven: junit:junit:4.12">
```
bgruening: the folder more or less :)

caleb-easterly:
@bgruening how much squashing do you want? I could go down to my first commit so it's all in one commit but I'm not sure what the best practice is.

bgruening:
If you don't mind I can also squash during merge, I just want to get rid of the .idea files from the history.

caleb-easterly:
I finally got the .idea folder completely out with the latest commit. Would that be sufficient with the squash?

tools/validate_fasta_database/FastaHeader.iml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module org.jetbrains.idea.maven.project.MavenProjectsManager.isMavenModule="true" type="JAVA_MODULE" version="4">
+  <component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_1_7">
```
bgruening: Here is still something wired :(

bgruening:
Not sure, you added still more than 80 files :(

caleb-easterly:
ah, I think a lot of that is the javadoc. should I ignore that too?

bgruening:
You only  need the tool.xml files and the test-data ... + a .shed.yml

caleb-easterly:
ok, I see. is there a current best practice on where to keep the source code? if i just put in the .jar it would be totally impenetrable aside from decompiling. Should I just keep the code in my own repo?

bgruening:
keeping the source in a separate repo is completely fine.
And then the next step would be to create a BioConda package for it.

caleb-easterly:
Should be cleaned up now! The code itself is at <https://github.com/caleb-easterly/validate_fasta_database>. Let me know if there's anything else I should do.

bgruening:
@caleb-easterly is this ready for the next review?

caleb-easterly:
Yes, but I wasn't sure about a couple of things - this commit includes the UniPept viewer, and I wasn't sure if that should be hosted here or deleted now that the most updated version is in Galaxy's dev branch. 

Also, I tried squashing, but there was some upstream stuff that I wasn't sure about so I'd thought I'd leave it to someone with more experience.

bgruening:
@caleb-easterly yeah the unipept visualization should be removed from this PR. The source is now in Galaxy and we should not duplicate it here. Can you do this before I review?

caleb-easterly:
done!

bgruening:
@caleb-easterly can we put `validate_fasta_database-0.2.jar` into conda?

caleb-easterly:
@bgruening Yes, I can work on that. Then, does the resulting conda package get pulled in by a <requirements> tag?

bgruening:
Yes, if you define the conda package in the `<requirement>` tag, it gets automatically pulled by Galaxy and we can keep this repo clean :)

caleb-easterly:
@bgruening Will do! I'm going on vacation until September but I'll work on it then.

bgruening:
Awesome thanks and enjoy your vacation!

PR #175, opened 2017-07-27
use GALAXY_SLOTS variable, default 12 threads

caleb-easterly:


bgruening:
@caleb-easterly default 12 seems to be really much, can we go for 4 to run these tools on normal computers as well?

caleb-easterly:
Anything is fine by me! I just picked 12 because that's what's in the diamond wrapper :)

bgruening:
Narf, we should fix this in the next Diamond update! :)

PR #177, opened 2017-08-07
remove datatype conf file from distribution

bgruening:
@blankclemens can you make sure that this file will not pop-up again.
All datatypes are now maintained in Galaxy core.


blankclemens:
Sure thing! Thanks!

PR #178, opened 2017-08-08
Fix format tag for text

blankclemens:
The proper `format` tag for text files in Galaxy is `txt` instead of `text` (not to be confused with the `type` tag).

Bump version number for affected wrappers.


bgruening:
Cool, danke vielmals!

PR #181, opened 2017-08-18
Fix issues with FileMerger

blankclemens:
Fix `rep_param_raw_rt_custom`.
Remove `$rep_param_rt_concat_trafo_out`.


bgruening:
`idxml` needs to be allowed as input I guess.

blankclemens:
Not according to the tool description. Can you confirm that it works?

PR #187, opened 2017-08-23
Small update to msgfplus

glormph:
This PR updates the msgfplus.xml tool.

1 - help text was wrong about forward and reverse input
2 - usage of `display_name` is not recommended, marius told me on gitter to use `element_identifier` mostly
3 - Add `.mzML` to the input file name because msgfplus crashes when the extension is wrong and sometimes element identifiers do not contain the mzML extension.


glormph:
One test fails, but that  may have something to do with travis switching to trusty. It seems to run MSGFplus quite ok. Tests fail locally as well on my OSX, but they both do before and after this PR.

I can make the size delta for test output bigger.

bgruening:
Thanks @glormph!

PR #190, opened 2017-08-24
Naming output of MSGFplus paired spectra/db better

glormph:
The old output schema resulted in output named "forward.mzML" since that is the unchangeable name of the paired element that was the spectra dataset.

I changed to use the `name` attribute of the element containing the dataset pair, which results in proper spectra file names. I use `name` and not `element_identifier` since the element itself is a collection and will raise a KeyError if `element_identifier` is used.

A bit annoying this didnt end up in yesterdays PR.

### DIFF
--- a/tools/msgfplus/msgfplus.xml
+++ b/tools/msgfplus/msgfplus.xml
@@ -1,4 +1,4 @@
-<tool id="msgfplus" name="MS-GF+" version="0.2">
+<tool id="msgfplus" name="MS-GF+" version="0.3">
     <description>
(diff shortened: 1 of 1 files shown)


glormph:
I don't understand the Travis output, should we retry the test?

bgruening:
Can you close and reopen this should trigger it. Sorry on my mobile.

glormph:
Test was canceled? Reopening again.

glormph:
Is anyone but @bgruening able to merge stuff? (Also Björn are you alright?)

PR #192, opened 2017-08-25
Tool to align fractionated DBs and spectra files

glormph:
I ran into a problem pairing fractionated DBs with fractionated spectra where some spectra fractions where pooled (low peptide fractions, which saves time on MS) and some where reruns (e.g fr2 and fr2b).

This tool aligns the databases that result from pI database split tool nicely with the spectra.


glormph:
Closing this PR until it also works in my WF. Tool so far only works stand alone for some reason.

glormph:
Reopened and it was apparently the workflow which was at fault.

tools/pi_db_tools/delta_pi_calc.xml
```diff
@@ -1,8 +1,11 @@
 <tool id="calc_delta_pi" name="Add delta pI" version="1.1">
+    <description>to peptide table</description>
     <requirements>
         <requirement type="package" version="3.6">python</requirement>
     </requirements>
-    <description>to peptide table</description>
+    <stdio>
```
bgruening: Can you please use https://docs.galaxyproject.org/en/latest/dev/schema.html#id5 instead of stdio. `exit_code` should be the one which you need.

glormph: Thanks, I had no idea this could be spec'ed on the command element!

tools/pi_db_tools/align_dbspec.py
```diff
@@ -0,0 +1,127 @@
...
+    {'fr01', 'fr04'} # Normal filename set
+    and
+    {'fr03': ['fr02', 'fr03']}  # pool definition
+    and
+    {'fr04': 'fr04', 'fr04b': 'fr04'}  # rerun fraction, rerun may also be pool
+    """
+    specrange = get_fn_fractionmap(specfiles, frregex)
+    to_pool = []
+    poolmap, rerun_map, normal_fns = {}, [], set()
+    for i in range(0, len(dbfiles)):
```
bgruening: You could use `for i,dbfile in enumerate(dbfiles)`

glormph: Yeah but my pythong linter will complain because I a don't use the `dbfile` variable then.

bgruening:
Thanks @glormph!

PR #196, opened 2017-09-08
Added error handling to delta pI calculation 

glormph:
Sometimes a fraction number is NA or no pI strip can be detected in filename. In both cases the output will be NA for experimental pI and delta pI


bgruening:
Nice improvement!
Thanks!

PR #199, opened 2017-09-18
[fix] FidoAdapter.xml in OpenMS

blankclemens:
Names of executables fixed.


bgruening:
Is this a larger problem in the generators?

blankclemens:
Nope, that's manually added anyways.

PR #201, opened 2017-09-25
(re)added zip as requirement of SearchGUI

bernt-matthias:
SearchGUI needs `zip`, see: https://github.com/bernt-matthias/tools-galaxyp/blob/0a0c1b53f934a0edfc7f18e9c401ad353a44092a/tools/peptideshaker/searchgui.xml#L419. This has been added before: 

https://github.com/galaxyproteomics/tools-galaxyp/commit/87d3dffcc9ef41e1f7c8488d1939165056f19413#diff-ec5346afe3992b199241a92092c2e047

but unfortunately removed later

https://github.com/galaxyproteomics/tools-galaxyp/commit/a2bd191e342173a297b0d60caa09bd4826239ad5#diff-ec5346afe3992b199241a92092c2e047

### DIFF
--- a/tools/peptideshaker/searchgui.xml
+++ b/tools/peptideshaker/searchgui.xml
@@ -7,7 +7,7 @@
     </macros>
     <requirements>
 	<requirement type="package" version="@SEARCHGUI_VERSION@">searchgui</requirement>
-	<requirement type="package">zip</requirement>
+	<requirement type="package" version="3.0">zip</requirement>
     </requirements>
     <expand macro="stdio" />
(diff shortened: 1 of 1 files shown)


bgruening:
@bernt-matthias this should be either added to the bioconda recipe or we should include here a version of the zip-tool.

bernt-matthias:
@bgruening I think it should be here, because zip is called from the script in the xml file (and not searchGUI).

bgruening:
Thanks!

PR #202, opened 2017-10-06
Consolidate Bumbershoot tools into single directory

chambm:
* with multi-repository shed.yml
* this way, when the Bumbershoot version macro is updated, all tools will be updated in the toolshed


chambm:
@bgruening Is my multi-repository shed.yml correct? Does it get tested by Travis? And with this consolidation, do I still need to modify the individual tool XML files to get Travis to test them?

bgruening:
@chambm this does not look like it should be :)
Please have a look at this one for inspiration: https://github.com/galaxyproject/tools-iuc/blob/master/tools/datamash/.shed.yml

chambm:
I was following this example: https://github.com/galaxyproject/planemo/blob/master/tests/data/repos/multi_repos_flat_configured/.shed.yml

I intentionally didn't want to make it a suite, since IDPicker and MyriMatch don't really belong in a suite together. They each work without the other. But the IDPicker tools would make sense as a suite.

bgruening:
A sorry, I thought you want to make a suite out of it.
As a test, you could `shed_upload` this to the TTS under your account and make sure the same repos are created. We need to make sure that we push to the same old repos.

PR #205, opened 2017-10-13
add moff for Caleb

bgruening:
ping @caleb-easterly 

* test files are too big
* check all licenses
* proper conda packages are missing


bgruening:
@caleb-easterly anything I can do here. Or is this ok?

caleb-easterly:
@bgruening I'm working on the conda recipe now, and I've needed to make some changes upstream to the moFF code so it can work in bioconda (if you want to take a look at the changes, they're at https://github.com/compomics/moFF/pull/22). moFF is already in bioconda, but I missed a couple things while writing the recipe, so it needs updates before it actually works. 

If the moFF team is willing to accept the changes, I can update the conda recipe and then start testing the Galaxy wrapper - I was able to track down some small data, so I think I can get automated testing involved. I'll work on this early next week.

bgruening:
Changes looks good to me, but you should be able to get this running even without these changes. tix should be on PATH, this way you don't need this one change. In worst case you could copy the binaries into Galaxy temp dir and execute in from within ...

Thanks a lot for your work, looking forward to this tool!

caleb-easterly:
Yes, as I had it in the beginning, I had copied all the necessary binaries (and one text file) and symlinked them. But I thought it was a bit strange to have `moff_setting.properties` and the like on the PATH, so the changes were intended to keep things clean (and not expose moFF stuff to the environment when it doesn't need to be exposed). I was thinking about conda usage outside of Galaxy, as well, when we wouldn't want to have stuff cluttering up the conda environment. But I'm totally open to suggestions about how to do it otherwise.

bgruening:
This is totally the right way to do! Up streaming as much as possible and improving things on the way!
Just saying don't let these things stop you :)

bgruening:
@caleb-easterly new release here: https://github.com/compomics/moFF/releases/tag/v1.2

bgruening:
Closing in favor of: https://github.com/galaxyproteomics/tools-galaxyp/pull/234

PR #206, opened 2017-10-16
Dbbuilder metabolomics

Stortebecker:
Hi. 

I wrote a small tool to download metabolite databases. The tool was based on the "Protein Database Downloader" (dbbuilder). I was not sure, if I really need the `#set $type = "direct"`, but left it like in dbbuilder.

The data are correctly downloaded and the preview in the history seems to work correctly. However, when I click `View data` it shows only rubbish. Can somebody tell me why?

Cheers, Flo

P.S.: My first tool!


bgruening:
@lecorguille and @RJMW could you look at this please?

bgruening:
@Stortebecker I cleaned the file a little bit up, mostly 4 spaces rule and adjusted the requirement tag.
I also pinged my favorite metabolic heros - let's see if we can get testers and a review.

How have you tested the tool?

Stortebecker:
@bgruening Thanks! I did not test every single possible output, but many of them. Everything worked for me. I would need some help to implement an automated test, like you did for the dbbuilder.

lecorguille:
What @bgruening ask to you is to add some functional tests using the tags `<tests> and <test>`
See this tutorial from at least [here](https://galaxyproject.github.io/training-material/topics/dev/tutorials/tool-integration/slides.html#36). The functional test part if from [here](https://galaxyproject.github.io/training-material/topics/dev/tutorials/tool-integration/slides.html#42).
Functional test will help us to validate your tools and any further modifications.

I think that there is a misunderstood between `tool-data` and `test-data`.

So, that make me wondering if your tool more like a [Data Manager](https://galaxyproject.org/admin/tools/data-managers/) than a tool?

What is the downloaded files size?
Because for databank, Galaxy usually caches the files for all users ([ref](https://galaxyproject.org/admin/data-integration/)) instead of ask them to download N copies of the same files.

Stortebecker:
@lecorguille Thanks for the links. I guess the tool really is more like a Data Manager. Though I am not too happy with the thought that I would have to rebuild it complete for it to work like a Data Manager. As I said, I took the Proteomics Database Downloaded (`dbbuilder`) as a template and worked it up from that. 

The files downloaded are about 100-200 MB in size.

bgruening:
@lecorguille I'm not sure about the frequency the datasets are updated. For the proteomic counter part @jj-umn decided that they are so frequently updated that a tool is more appropriate. It just saves you time by choosing a database instead of visiting the DB, searching for the download link and copying this into Galaxy. 

If the datasets are frequently updated I think there is a need for such a tool.

@Stortebecker a few tests would be nice to include. You can use "assert" for the tests for example.

lecorguille:
Ping @fgiacomoni and @yguitton
Because honestly, I don’t know :)

fgiacomoni:
Hi. just to have a better idea: HMDB V1.0: 2007, V02: 2009, V03: 2013 and last version is V04.
Usually we don't manage databases as in proteomics and genomics where db managers are crucial. We also currently use webservices provided by reference banks.
But happy to have feedback from @RJMW and @yguitton ;-)

Stortebecker:
@lecorguille @bgruening I added automatic unzipping of the downloads and tool tests. In accordance with @bgruening, I did not rewrite the tool as a data manager.

bgruening:
@Stortebecker please add a .shed.yml file. See here for an example: https://github.com/galaxyproteomics/tools-galaxyp/blob/master/suites/suite_bumbershoot/.shed.yml

Stortebecker:
@bgruening Can you help with the second test? I do not get why it fails.

bgruening:
I fear this is a larger one, as the tests includes a loc file. I will try to work on it in the next weeks and hope that @lecorguille beats me to it :)

Thanks for your efforts, this is a great start!

Stortebecker:
@bgruening @lecorguille If it is to complicated to fix this test, I could think of another one. Where can I find this loc file you were referring to? Maybe I can fix the test myself.

lecorguille:
@Stortebecker Usually, the .loc are used to indicate the file locations on the localhost filesystem like on [this example](https://github.com/galaxyproject/tools-iuc/blob/master/tools/hisat2/tool-data/hisat2_indexes.loc.sample#L36)

The idea is to just display to the user a combo list with the name of the ressource/databank. Afterward, Galaxy uses the .loc to complete the command line with the file location.

In your case, maybe, you should use a classic `<param type="select"` unless you want the instance administrators set the list the tool will present to the end-user.

Wahouw, I guess I'm not really clear.

Stortebecker:
@lecorguille Thanks, got it.

Issue #209, opened 2017-10-25
Custom Mods in MSGFPlusAdapter


Stortebecker:
Hi.

I did not find a possibility to add custom mods to the MSGFPlusAdapter. In OpenMS, users are expected to manually modify the mods.txt. However, I cannot find the file to modify in Galaxy. Our own MSGF+-Wrapper features a nice way to add custom mods, which I now also requested for MSGFPlusAdapter (see this [OpenMS Issue](https://github.com/OpenMS/OpenMS/issues/2977)).

It is not really high priority, as our own wrapper works fine, but it pretty much cripples the MSGFPlusAdapter for use-cases with stable isotope labelling. The latter are not listed in unimod.

bgruening:
Does peptideshaker offers this?

Stortebecker:
nope.

glormph:
Custom mods work (I think) in the standalone MSGF+. Can one import mzIdentMLs to openMS from there?

Stortebecker:
> It is not really high priority, as our own wrapper works fine, but it pretty much cripples the MSGFPlusAdapter for use-cases with stable isotope labelling. The latter are not listed in unimod.

**Correction:** Stable isotope labels **are** listed in unimod, but only for one aa each. Example: `Label 13C(6) (K)` and `Label 13C(6) (R)`, but not `Label 13C(6) (KR)` (a standard Silac label). I tested both by running MSGF+ directly and manually adding the corresponding entries to `Mods.txt`. There seems to be only minor differences in MSGFPlus runtime and output.

**But:** This does not solve our problem, as our wrapper for MSGFPlusAdapter only allows for one variable mod. I fixed this by deleting this restriction: https://github.com/galaxyproteomics/tools-galaxyp/pull/210

**Conclusion:** There is still no way to add custom mods in MSGFPlusAdapter, but the tool works for common labelling strategies, now. However, we can use our own MSGF+ wrapper for custom mods.

PR #210, opened 2017-10-26
Allow for multiple variable modifications in MSGFPlusAdapter

Stortebecker:
The wrapper for MSGFPlusAdapter allowed only one variable modification, although the tool can take an process several. 

This fix solves part of https://github.com/galaxyproteomics/tools-galaxyp/issues/209

### DIFF
--- a/tools/openms/MSGFPlusAdapter.xml
+++ b/tools/openms/MSGFPlusAdapter.xml
@@ -1,7 +1,7 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <!--This is a configuration file for the integration of a tools into Galaxy (https://galaxyproject.org/). This file was automatically generated using CTD2Galaxy.-->
 <!--Proposed Tool Section: [Identification]-->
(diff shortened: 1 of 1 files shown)


tools/openms/MSGFPlusAdapter.xml
```diff
@@ -2738,7 +2738,7 @@
         <option value="trifluoro (L)">trifluoro (L)</option>
       </param>
     </repeat>
-    <repeat name="rep_param_variable_modifications" min="0" max="1" title="param_variable_modifications">
```
bgruening: This also seems to be a bug in the CTD to Galaxy converter. Why should we using a repeat if there is only one allowed case.

Thanks @Stortebecker!

blankclemens: It's the other way round: Many are allowed, no need for max, this should fix it: https://github.com/WorkflowConversion/CTDConverter/pull/34

bgruening:
Merci!

PR #211, opened 2017-10-26
Update readme.md

blankclemens:
Just a small note so I wont forget it in future. Remove the param_id_pool parameter from IsobaricAnalyzer since it is not used.


bgruening:
Thanks.

PR #204, opened 2017-10-13
New version for msgf+: 20170721

glormph:
I saw that @blankclemens had updated bioconda which meant I didnt have to do that myself 👍 Thanks and here comes the galaxy tool update.

PR #212, opened 2017-10-27
Fixed Msgf custom modification with multiple aa_specificities

Stortebecker:
Hi.
Our MSGF+ wrapper had a bug when using custom mods with more than one `aa_specificity`. I fixed this and adapted the last test accordingly. Instead of comparing the diff, the test looks for the identification of the custom mods now.
Cheers, Flo

P.S.: My first steps into Python/Cheetah!

### DIFF
--- a/tools/msgfplus/msgfplus.xml
+++ b/tools/msgfplus/msgfplus.xml
@@ -1,4 +1,4 @@
-<tool id="msgfplus" name="MS-GF+" version="0.3">
+<tool id="msgfplus" name="MS-GF+" version="0.3.1">
     <description>
         Identifies peptides in tandem mass spectra using the MS-GF+ search engine.
     </description>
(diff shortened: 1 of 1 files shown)


Stortebecker:
Why did the Travis-test fail? I cannot see the error. Can somebody please help?

glormph:
I think it may be the same problem as in #204. @bgruening has pinged @chambm about incorrect test-data I think.

Otherwise, the PR looks good to me!

chambm:
The change to the test to only check for the mod strings is quite a bit less stringent. If, for some strange reason, everything else about the file broke but the mods were right it would still pass. :) Can we assert both unchanged (+/- 6 lines) and also assert contents?

bgruening:
restarted the tests.

bgruening:
Cool, thanks a lot @Stortebecker!

bgruening:
@chambm upps just saw your comment now.
@Stortebecker would be nice if you can address this in a follow up PR.

PR #214, opened 2017-10-27
Improved msgf custom mod testing

Stortebecker:
Improves the test on the custom mods as suggested here: https://github.com/galaxyproteomics/tools-galaxyp/pull/212

I had to replace the file tested against, as the cGMP mod is working now and occuring 66 times. @chambm Could you check, if `lines_diff="6"` is sufficient?


chambm:
You forgot to change the output filename. I also suggest that you don't rename the file so we can compare the old one to the new one. The commit history will suffice to tell us when it was created (not sure why I put that datestamp in there to be honest).

Stortebecker:
Hmpf. What did I forget now? Ah, I see. One minute.

bgruening:
@Stortebecker can you please update the test data. We have now a new msgfplus version.

Stortebecker:
@bgruening All right, schon fertig.

bgruening:
Danke!

PR #221, opened 2017-11-02
One way to address failure to update multi-repository directories

chambm:


bgruening:
@chambm if you really want to work on travis, please take over all changes from tools-iuc. We have a pretty advanced travis setup, which I wanted to port over long time ago.

chambm:
I don't really want to work on travis, I just want this one thing fixed which I only noticed after consolidating the Bumbershoot tools in a single directory. Does this approach to a fix have a serious side effect that I don't realize?

tools/bumbershoot/idpassemble.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<tool id="idpassemble" name="idpAssemble" version="@VERSION@.1">
+<tool id="idpassemble" name="idpAssemble" version="@VERSION@.2">
```
bgruening: can you revert this, its not needed right? I will upload by hand

chambm: It is needed to test that the other commit actually has the intended effect. But it didn't work for the PR because PRs don't trigger uploads. :/ I figured a wrapper number bump was pretty harmless, but I guess it could be an extra space newline too.

bgruening:
These things are hard to test, but I would not abuse tools for this. I will migrate the tools-iuc stuff over ok and we can see if this gets better. I will work on it, promised.

I updated your tools on both TS!
Thanks @chambm!

PR #224, opened 2017-11-06
Fixed XTandemAdapter: deleted restriction to only one variable mod

Stortebecker:
Like in the MSGFPlusAdapter, we had an incorrect restriction that allowed for only one variable modification. Compare https://github.com/galaxyproteomics/tools-galaxyp/pull/210


bgruening:
ping @blankclemens

blankclemens:
Same issue, should be automatically fixed with https://github.com/WorkflowConversion/CTDConverter/pull/34 in the next OpenMS release.

PR #225, opened 2017-11-07
Update idpassemble and idpQonvertEmbedder

chambm:
* Add ReporterIonMzTolerance and NormalizeReporterIons as isobaric-quantitation-specific options in idpQonvertEmbedder.xml
* Fix documentation formatting for idpassemble.xml
* Hide assembly and mapping inputs behind a boolean conditional in idpassemble.xml so that they don't expand workflow steps when unset


chambm:
@bgruening The master build says it successfully updated the idpassemble and idpqonvert repositories, but I don't see a new version on the toolshed.

PR #226, opened 2017-11-09
fix column headers in eggnog-mapper

caleb-easterly:
The column headers in `annotations` changed in 1.0. and I missed it in the original wrapper.


bgruening:
Thanks!

PR #232, opened 2017-12-04
Msgfplus tsv showDecoy

glormph:
This implements the `-showDecoy 1` parameter for creating TSV files from MSGF+ mzids. Enabling the parameter makes the resulting TSV file include decoy PSMs.


tools/msgfplus/msgfplus.xml
```diff
@@ -82,7 +85,13 @@
                       label="Collection: Pairs of spectra (forward) and FASTA database (reverse)"/>
               </when>
         </conditional>
-        <param name="tsvcheck" type="boolean" label="Output TSV as well?" />
+        <conditional name="tsv">
```
bgruening: Why do you use a conditional here, it could be a single boolean isn't it?

glormph: It's to make the showDecoy boolean appear only if produce TSV is checked. Wrong UI?

bgruening: Ok got it, I think this is a little bit confusing. Why not a simple select, "TSV outpout", "TSV ooutput with decoys" ...

bgruening:
Thanks!

Issue #135, opened 2017-02-23
SearchGUI -prec_ppm / "Precursor Ion Tolerance Units" issue


yvanlebras:
Dear Galaxy-P colleagues,

Testing the galaxy-p galaxy flavour, using search GUI Version 2.9.0 tools, I have changed the default "Precursor Ion Tolerance Units" to Da and have the following error: `Error parsing the prec_ppm option: Found 2 where 0 or 1 was expected.`

Cheers,

Yvan

bgruening:
mh, ping @jj-umn

chambm:
@yvanlebras Can you update your PeptideShaker/SearchGUI package? That bug should be fixed.

PR #233, opened 2017-12-12
*tol_units changed form 1/2 to 1/0

bernt-matthias:
fixes https://github.com/galaxyproteomics/tools-galaxyp/issues/135

According to https://github.com/compomics/searchgui/wiki/ReleaseNotes this changed back in SearchGUI 2.8.0

Please note that also with this fix the GTN metaproteomics tutorial does in the peptideshaker step.


bgruening:
We should probably increase the version number of this tool?

bernt-matthias:
I think that it would be a good idea to add the GTN tutorial data as a test case, since it would be really great if the GTN tutorials would just work :)

bernt-matthias:
Jep we could, but this would require quite a bit of modifications to the tool, e.g.:

Changes in SearchGUI 3.2.15 (May 8. 2017):
    FEATURE IMPROVEMENT: Added "Dimethylation of K 2H4" to the standard list of PTMs.

Probably there are more. 

I'm sure that I'm not experienced enough with this tool to do such changes.

bgruening:
We will soon run all tutorials and there workflows weekly to test them, and catch errors early one.

@bernt-matthias what I mean is not increasing the requirement, just bump the tool version. So this line https://github.com/bernt-matthias/tools-galaxyp/blob/8b0e5d96f02d44e191cad12d15455b2ba46613b6/tools/peptideshaker/searchgui.xml#L1 to .2 and the peptideshaker one as well.

bernt-matthias:
@bgruening got it..  and excellent idea to run regular test of the tutorials

bgruening:
Damn, forgot about this one. Will fix the test tomorrow and merge. 
Sorry Mathias!

PR #234, opened 2017-12-15
A first pass at the moff wrapper. 

caleb-easterly:
The biggest question for me right now is the handling of multiple output files. To start, I used discovered datasets. A potential for the future is using a paired dataset collection. In the `moff_all.py` MBR followed by quantitation, each identification file must be matched with a spectrum file. The number of outputs is the (number of inputs) times 2, with another 1 if the peptide summary is enabled. It would be nice to have the logs paired with the outputs. I don't know how user-friendly it is to require dataset collections at this point. 

Also, the test files are small, but they mostly verify that the code runs without error, not that the correct answers are obtained. moFF has a problem with the mini mzML, so it just returns all -1's


jj-umn:
@caleb-easterly   I think it might be easier to use this in a workflow if the output_table and output_logs were in a dataset collection.   

It might also be useful allow the inputs to be specified as a Dataset Pair or as a List of Dataset Pairs.  Bowtie2 and HISAT2 have options like that.

caleb-easterly:
Is it useful to have the dataset collection option (for input, output, or both) for a single pair of datasets? Or should it be available only when the user has multiple pairs of datasets? I haven't used dataset collections, so I'm not sure what is best.

tools/moFF/moff.xml
```diff
@@ -0,0 +1,351 @@
...
+        <token name="@WRANGLE_RAW_INPUT_MULTIPLE@"><![CDATA[
+            mkdir ./raws &&
+            ## for files, need to softlink the display name to the history item
+            #for $counter,$filename in enumerate($task.msms_input.inputraw):
+                ln -s '$filename' './raws/$filename.display_name' &&
+            #end for
+        ]]></token>
+    </macros>
+    <requirements>
+        <requirement type="package" version="1.2">moff</requirement>
```
bgruening: We could also macrofy the version here and reuse it as tool-version as well.

tools/moFF/moff.xml
```diff
@@ -0,0 +1,351 @@
...
+        <token name="@WRANGLE_IDENT_INPUT_SINGLE@"><![CDATA[
+            mkdir ./ident_inputs &&
+            #if ($task.ident_input.input_type_selector == "ps"):
+                ln -s '$task.ident_input.psm_input' './ident_inputs/$task.ident_input.psm_input.display_name' &&
+            #else
+                ## optionally remove first line
+                #if ($task.ident_input.remove_header):
+                    sed -i '1d' '$task.ident_input.generic_input' &&
+                #end if
+                ## header row with correct names: "peptide", "prot", "mod_peptide", "rt", "mz", "mass", and "charge"
```
bgruening: Is this strictly needed or can we just use metadata to describe the columns?

tools/moFF/moff.xml
```diff
@@ -0,0 +1,351 @@
...
+                awk 'BEGIN{OFS="\t"; FS="\t"}{print \$pep,\$prot,\$mod,\$rt,\$mz,\$mass,\$charge}' pep="${task.ident_input.peptide}" prot="$task.ident_input.prot" mod="$task.ident_input.mod_peptide" rt="$task.ident_input.rt" mz="$task.ident_input.mz" mass="$task.ident_input.mass" charge="$task.ident_input.charge" '$task.ident_input.generic_input' >> tempfile.tab &&
+                mv tempfile.tab '$task.ident_input.generic_input' &&
+                ln -s '$task.ident_input.generic_input' './ident_inputs/$task.ident_input.generic_input.display_name' &&
+            #end if
+        ]]></token>
+        <token name="@WRANGLE_IDENT_INPUT_MULTIPLE@"><![CDATA[
+            mkdir ./ident_inputs &&
+            #if ($task.ident_input.input_type_selector == "ps"):
+                #for $counter,$filename in enumerate($task.ident_input.psm_input):
+                    ln -s '$filename' './ident_inputs/$filename.display_name' &&
```
bgruening: You don't use the counter here and could potentially overwrite the link, or fail because of a link already exists.

tools/moFF/moff.xml
```diff
@@ -0,0 +1,351 @@
...
+                #end for
+            #else
+                #for $counter,$filename in enumerate($task.ident_input.generic_input):
+                    ## remove first line (add param for this later on!)
+                    #if ($task.ident_input.remove_header):
```
bgruening: brackets are not needed.

tools/moFF/moff.xml
```diff
@@ -0,0 +1,351 @@
...
+                ## header row with correct names: "peptide", "prot", "mod_peptide", "rt", "mz", "mass", and "charge"
+                echo -e "peptide\tprot\tmod_peptide\trt\tmz\tmass\tcharge" > tempfile.tab &&
+                awk 'BEGIN{OFS="\t"; FS="\t"}{print \$pep,\$prot,\$mod,\$rt,\$mz,\$mass,\$charge}' pep="${task.ident_input.peptide}" prot="$task.ident_input.prot" mod="$task.ident_input.mod_peptide" rt="$task.ident_input.rt" mz="$task.ident_input.mz" mass="$task.ident_input.mass" charge="$task.ident_input.charge" '$task.ident_input.generic_input' >> tempfile.tab &&
+                mv tempfile.tab '$task.ident_input.generic_input' &&
+                ln -s '$task.ident_input.generic_input' './ident_inputs/$task.ident_input.generic_input.display_name' &&
+            #end if
+        ]]></token>
+        <token name="@WRANGLE_IDENT_INPUT_MULTIPLE@"><![CDATA[
+            mkdir ./ident_inputs &&
+            #if ($task.ident_input.input_type_selector == "ps"):
```
bgruening: remove brackets

tools/moFF/moff.xml
```diff
@@ -0,0 +1,351 @@
...
+                    ln -s '$filename' './ident_inputs/$filename.display_name' &&
+                #end for
+            #else
+                #for $counter,$filename in enumerate($task.ident_input.generic_input):
+                    ## remove first line (add param for this later on!)
+                    #if ($task.ident_input.remove_header):
+                        sed -i '1d' '$filename' &&
+                    #end if
+                    ## header row with correct names: "peptide", "prot", "mod_peptide", "rt", "mz", "mass", and "charge"
+                    echo -e "peptide\tprot\tmod_peptide\trt\tmz\tmass\tcharge" > tempfile.tab &&
```
bgruening: Better with metadata?

tools/moFF/moff.xml
```diff
@@ -0,0 +1,351 @@
...
+                #end for
+            #end if
+        ]]></token>
+        <token name="@RAW_INPUT_ARG_SINGLE@"><![CDATA[
+            --inputraw './raws/$task.msms_input.inputraw.display_name'
+        ]]></token>
+        <token name="@RAW_INPUT_ARG_MULTIPLE@"><![CDATA[
+            --inputraw
+            #for $counter,$filename in enumerate($task.msms_input.inputraw):
+                './raws/$filename.display_name'
```
bgruening: counter?

tools/moFF/moff.xml
```diff
@@ -0,0 +1,351 @@
...
+                ln -s '$filename' './raws/$filename.display_name' &&
+            #end for
+        ]]></token>
+    </macros>
+    <requirements>
+        <requirement type="package" version="1.2">moff</requirement>
+    </requirements>
+    <command detect_errors="aggressive"><![CDATA[
+        mkdir ./out &&
+        #if ($task.task_selector == "moff"):
```
bgruening: brackets

tools/moFF/moff.xml
```diff
@@ -0,0 +1,351 @@
...
+                    help="Specify rt window for xic in minutes." />
+                <param argument="--rt_p" type="float" value="1" label="Time window for the peak"
+                    help="Specify the time windows for the peak in minutes." />
+                <param argument="--rt_p_match" type="float" value="1.5" label="Time window for the matched peak"
+                    help="Specify the time windows for the matched peak in minutes." />
+                <param argument="--peptide_summary" type="boolean" value="true" label="Output the peptide summary?"/>
+            </when>
+            <when value="mbr">
+                <expand macro="ident_input_macro" allow_multiple="true"/>
+                <param argument="--ext" type="text" size="30" value="tab" label="Provide the extension used in the display file name (without the period)"/>
```
bgruening: the size attribute is deprecated and can be removed from all params.

tools/moFF/moff.xml
```diff
@@ -0,0 +1,351 @@
+<tool id="proteomics_moff" name="moFF: A modest Feature Finder" version="1.2">
+    <description>extracts MS1 intensities from Thermo raw files</description>
+    <macros>
+        <!-- xml macros, used for shared Galaxy parameter inputs -->
+        <xml name="ident_input_macro" token_allow_multiple="true">
+        <!-- this is exactly the same across all three, except for allowing multiple in MBR and all but not in moff -->
+            <conditional name="ident_input">
+                    <param name="input_type_selector" type="select" label="Choose the format for the identification file:">
+                        <option value="ps">Peptide Shaker PSM report (standard, not extended)</option>
+                        <option value="generic">Another tabular identification file</option>
```
bgruening: remove "Another"?

tools/moFF/moff.xml
caleb-easterly: Not sure on this point - I don't have much experience with metadata. But the reason I did this is because moFF requires the exact column names. Is that possible with metadata?

tools/moFF/moff.xml
caleb-easterly: What do you suggest to do about the link? That would happen if 2 or more datasets had the same display name, right? That would pretty much be an unresolvable problem because the only way that moFF associates an ID file with spectrum file is by name.

tools/moFF/moff.xml
bgruening: Metadata is not part of the file, so if this is needed your approach is good!

tools/moFF/moff.xml
bgruening: You could do something like this:
```
#for $counter,$filename in enumerate($task.ident_input.psm_input):
     ln -s '$filename' './ident_inputs/$counter_$filename.display_name' &&
#end for
```

bgruening: >  file is by name

You mean by filename on the hard-drive? If so you could use the counter, no?

caleb-easterly: I don't think we could use the counter, because if the ID files and spectrum files have the same names but are provided in a different order there will be unmatched files.

moFF requires files to have the same name on the hard drive, which is the reason that we have to rename them by the display name.

bgruening:
@caleb-easterly I will leave this to @jj-umn, but relying on display names is not good in the long run. The also this here: https://github.com/galaxyproject/galaxy/issues/5239#issuecomment-353440859

I'm ok with merging now, but we should work on this in the near future.

caleb-easterly:
I'm going to work on using dataset collections to take out the need to discover datasets. As far as I can tell from the moFF code, the only way to not rely on display names would be to use a paired dataset collection.

caleb-easterly:
@bgruening, I'm having trouble with the dataset collections. I just pushed what I have so far, but am getting the following error: 

```
Traceback (most recent call last):
  File "/tmp/tmpUfGUGC/galaxy-dev/lib/galaxy/tools/__init__.py", line 1296, in handle_single_execution
    job, out_data = self.execute( trans, incoming=params, history=history, rerun_remap_job_id=rerun_remap_job_id, mapping_over_collection=mapping_over_collection, execution_cache=execution_cache )
  File "/tmp/tmpUfGUGC/galaxy-dev/lib/galaxy/tools/__init__.py", line 1376, in execute
    return self.tool_action.execute( self, trans, incoming=incoming, set_output_hid=set_output_hid, history=history, **kwargs )
  File "/tmp/tmpUfGUGC/galaxy-dev/lib/galaxy/tools/actions/__init__.py", line 355, in execute
    known_outputs = output.known_outputs( input_collections, collections_manager.type_registry )
  File "/tmp/tmpUfGUGC/galaxy-dev/lib/galaxy/tools/parser/output_objects.py", line 123, in known_outputs
    collection_prototype = inputs[ self.structure.structured_like ].collection
KeyError: 'ident_input_file'
2017-12-27 15:15:56,551 WARNI [galaxy.tools.execute] There was a failure executing a job for tool [proteomics_moff] - Error executing tool: 'ident_input_file'
ERROR
```
It seems like a problem in the `structured_like="ident_input_file"`, but I can't find anything I've done wrong. 

I'm pretty new to dataset collections, so if you have other tips that would also help.

caleb-easterly:
Also, I'm not sure what's happening with the Travis-CI build - the recipe is currently broken, so it shouldn't be passing

bgruening:
@caleb-easterly how can I help here?

caleb-easterly:
I actually feel pretty good about the current state of it (after the latest push). I never did figure out the `structured_like` attribute, but it's working well with `<discover_datasets>` inside the output collections. If I'm not mistaken, those collections can then be passed to other tools in a workflow, right? 

Could you just look it over and let me know what you think? I'm ready to merge if you are.

tools/moFF/test-data/output1/moff_mbr_test_pepsum.tab
```diff
@@ -0,0 +1 @@
+peptide	prot	sumIntensity_mbr_test2	sumIntensity_mbr_test1
```
bgruening: This file is now not used anymore, right?

tools/moFF/moff.xml
```diff
@@ -0,0 +1,382 @@
...
+        </data>
+        <data format="txt" name="output_logs" label="${tool.name} log: ${on_string}">
+            <filter>task['task_selector']=='moff'</filter>
+        </data>
+        <collection name="ident_output" type="list" label="${tool.name} quantification: ${on_string}">
+            <filter>task['task_selector']=='all' or task['task_selector']=='mbr'</filter>
+            <!--discover datasets method -->
+            <discover_datasets pattern="(?P&lt;designation&gt;.*)\.txt" directory="out" format="tabular"/>
+        </collection>
+        <collection name="log_output" type="list" label="${tool.name} logs: ${on_string}">
```
bgruening: We should add a test for the logs and the other outputs if possible.

tools/moFF/test-data/output1/moff_mbr_test_pepsum.tab
caleb-easterly: true! removing now

tools/moFF/moff.xml
```diff
@@ -0,0 +1,382 @@
...
+            <filter>task['task_selector']=='all' or task['task_selector']=='mbr'</filter>
+            <discover_datasets pattern="(?P&lt;designation&gt;.*)\.log" directory="out" format="txt"/>
+        </collection>
+        <data format="tabular" name="output_peptide_summary" label="${tool.name} peptide summary: ${on_string}">
+            <filter>task['peptide_summary']</filter>
+        </data>
+    </outputs>
+    <tests>
+        <!-- test moff_all -->
```
bgruening: A test with the peptideshaker inputs would be also nice I think.

tools/moFF/moff.xml
caleb-easterly: I can work on that!

tools/moFF/moff.xml
caleb-easterly: tests 1, 3, and 4 use the PS inputs - only test 2 uses the generic input function.

caleb-easterly:
Yes, I agree that using display names isn't ideal. Does anything else come to mind as a way around it? It seems like at the early proteomics stages (raw files, SG/PS), the only file associations are by file names.

bgruening:
Something like this will not work for proteomics data? https://github.com/galaxyproject/tools-iuc/blob/dfa5a68cb20842407941c7ffda9ef956a0e86a04/tools/hicexplorer/hicPlotDistVsCounts.xml#L12

caleb-easterly:
I think the complexity is that whatever is done has to maintain the association between the RAW/mzML and the SG/PS output. moFF itself uses the file names to associate the two types of file (i.e., they must be named `file1.raw` and `file1.tab`. So the only ways I can think of to pair those two files are by order that they are provided to the tool or by display name, both of which seem pretty unstable, though display name seems less unstable than order. A paired dataset collection would mean that we could just iterate through the pairs, but I don't see a way to make the paired collection without relying on names or order. 

We don't need this very urgently! I think we're just trying to get it working and to see what we can do to make it easier to use. Let me know if the freiburg testing reveals any needed changes.

bgruening:
> itself uses the file names to associate the two types of file (i.e., they must be named file1.raw and file1.tab

How many of these raw files do you need to associate to file1.tab, or the other way around can it happen that file1.tab can map to multiple RAW files? 
We could use a repeat tag, isn't it? In each repeat are to data params for RAW and PS/SG. We then iterate over the repeat and name it `file_$i.raw`. This way it should stay together, or do I miss something?

caleb-easterly:
Currently, moff needs a one-to-one pairing of RAW/mzML files and id files. 

I'm not too familiar with repeat tags, but my concern with relying on the "order" of the files is that the order might get out of whack. For example, if the files are uploaded and then are put into collections `collect_raw = [file1.raw, file2.raw]` and `collect_id = [file2.tab, file1.tab]`, then it seems like iterating over the collections in tandem would give 
```
file1_1.raw <--> file2_1.tab
file2_2.raw <--> file1_2.tab
```
and then moFF would fail because none of the files have the same name. I might be misunderstanding the suggestion, though.

bgruening:
This is true, I was talking about single inputs. Lets go with this for the moment.
Thanks @caleb-easterly!

bgruening:
This is true, I was talking about single inputs. Lets go with this for the moment.
Thanks @caleb-easterly!

PR #237, opened 2017-12-19
add pep pointer tool

caleb-easterly:
This is @pravs3683's PepPointer tool to classify novel peptides based on their genomic coordinates.


tools/pep_pointer/pep_pointer.xml
```diff
@@ -0,0 +1,36 @@
+<tool id="pep_pointer" name="PepPointer: classify genomic location of peptides" version="0.1.0">
+    <requirements>
+        <requirement type="package" version="2.7.9">python</requirement>
+    </requirements>
+    <command detect_errors="aggressive"><![CDATA[
+        python $__tool_directory__/pep_pointer.py "$gtf" "$bed" "$classified"
```
bgruening: single quotes, and `$__tool_directory__/pep_pointer.py` should also be single quoted

caleb-easterly: what's the difference between single quotes and double quotes in...Cheetah? I'm used to shell, where `echo '$gtf'` gives `$gtf`

caleb-easterly: (I'm just the wrapper, so I'm not sure about BEDtools)

bgruening: Its more a bash'ism. Single quotes are not escaped/evaluated in bash, double quotes are. So double quotes do have a potential security impact if a user has ENV vars in there text params or filenames etc ...

bgruening: ping @jj-umn for the BEDtools Q.

caleb-easterly:
We're using PepPointer in a publication, so it would be useful to have it released (even if we determine later on that it duplicates some BEDtools functionality)

PR #243, opened 2018-01-11
update moff to 1.2.1

caleb-easterly:


caleb-easterly:
@bgruening , I'm wondering what could cause the failure here. I just updated the version and made a few other changes. The tests are passing on planemo on my computer.

bgruening:
@caleb-easterly it looks to me that the diff looks really different. I see the old files had `Doubtful` included and filenames. The new one has removed this as it seems :(

caleb-easterly:
yes, it doesn't really make sense! the history item on my local computer doesn't look like that at all

bgruening:
Can you run planemo with `planemo t --update_test_data` ?

caleb-easterly:
yep, I did that and it's still passing on my computer.

caleb-easterly:
ok, I finally got the error on my laptop, so i can try to fix it locally.

caleb-easterly:
I'm trying to just use `has_text` instead of comparing files. I'm not sure why this was happening but it looks like the diff was showing the history item with the rows out of order - the rows that seem to be missing all of the columns were in fact the bottom rows of the history file. 

If this passes, I could remove the test files that were being compared against. What do you think about using `assert_contents` instead of diff?

bgruening:
Its okish, I would prefer test files, as these could be used in Tool Describing Tours.
Do you know about the sort feature of tests? https://docs.galaxyproject.org/en/latest/dev/schema.html#id84

bgruening:
This latest error is not yours, I will care about this and let you know.

bgruening:
No it is failing again, even with sort :(
Lets go for assert then ... Thanks @caleb-easterly for trying!

bgruening:
Cool, test data can now be removed.

caleb-easterly:
yep, i'll clean it up and ping you when I'm done. thanks for the help!

caleb-easterly:
@bgruening done!

bgruening:
Super! Thanks a lot!

PR #247, opened 2018-01-25
FlashLFQ can not handle path dirs with dots

jj-umn:
change command to symlink spectrum files into relative dir
spectrum_dir.
FlashLFQ —rep spectrum_dir


jj-umn:
@bgruening I had to change this since one of installations has 'galaxpy.msi.umn.edu' as a dir in the files path  and FlashLFQ choked on that.

bgruening:
@jj-umn anything I can help with?

PR #250, opened 2018-01-26
Upgrade flashlfq

bgruening:
@jj-umn we can keep this open a little bit and collect improvements if you like. 
Otherwise feel free to merge.


jj-umn:
@bgruening Thanks.  I hadn't realized that the bioconda recipe had already been updated.

PR #203, opened 2017-10-12
Msstitch

glormph:
About time I PRed this. Has been reworked a number of times.

Msstitch is a tool for postprocessing of ID and quant data, and generates PSM, peptide and protein/gene tables. It does not fit everything right now, but is currently leaning on Percolator, MSGFplus TSV, OpenMS consensusXML and Hardklor/kronik output.

Allows for flexible modular pipelines since the tools are fairly small.

PR #252, opened 2018-02-05
Retrying an earlier PR on new branch

glormph:
What should have been done in #203


glormph:
What no tests? All the green stuff makes me suspicious.

bgruening:
;) you are missing the .shed.yml file, isn't it?

glormph:
D'oh! There goes my one-commit PR 😄

bgruening:
And the green color is also gone :)
I only see a few failures, most of them are working. I guess for the binaries you need to use sim_size.

tools/msstitch/create_protein_table.xml
```diff
@@ -0,0 +1,104 @@
+<tool id="create_protein_table" name="Create protein table" version="1.2">
```
bgruening: can we prefix them with msstitch_?

tools/msstitch/create_protein_table.xml
```diff
@@ -0,0 +1,104 @@
...
+                <param name="denompatterns" value="^tmt10plex_126" />
+            </repeat>
+            <param name="isoquant|minint" value="0.2" />
+            <param name="isoquant|normalize|type" value="medianother" />
+            <param name="isoquant|normalize|norm_ratio_table" value="protein_ratios.txt" />
+            <output name="output" value="protein_table.txt" />
+        </test>
+    </tests>
+    <help>
+        Creates a protein table from PSM and peptide tables. Choose for each entry
```
bgruening: An example of a PSM table would be nice. In which column is what ...

tools/msstitch/msslookup_quant.xml
```diff
@@ -0,0 +1,118 @@
+<tool id="msslookup_quant" name="Create lookup table with quant data" version="1.1">
+    <description>of openMS isobaric quant and Kronik MS1 quant</description>
```
bgruening: OpenMS

tools/msstitch/msslookup_quant.xml
```diff
@@ -0,0 +1,118 @@
+<tool id="msslookup_quant" name="Create lookup table with quant data" version="1.1">
```
bgruening: do we want to prefix all of these tools?

tools/msstitch/msslookup_seqspace.xml
```diff
@@ -0,0 +1,52 @@
+<tool id="msslookup_seqspace" name="Create lookup DB with sequences" version="1.1">
+    <description>from FASTA for filtering</description>
+    <requirements>
+        <requirement type="package" version="2.5">msstitch</requirement>
```
bgruening: You should probably use macros for all of these repetitive elements, citation, requirements a few params ...

tools/msstitch/msspercolator_merge.xml
```diff
@@ -0,0 +1,66 @@
+<tool id="msspercolator_merge" name="Merge Percolator" version="1.1">
```
bgruening: tool version is changing every now and then :) Maybe we should macrofy them as well and use 2.5 as the msstich version is doing?

glormph:
I pushed to fix tests but will do some work on the macros before merging then.

bgruening:
Thanks @glormph!

PR #253, opened 2018-02-05
tool to split file to collection

caleb-easterly:
This is probably overkill (big tool for a small task), but it's a version of the tools that @bgruening and @jj-umn already wrote, but with some more features. It can split FASTA, FASTQ, MGF, and tabular files, and allows for splitting by line (randomly or in sequential order) or on column values.

When splitting by column, the names for the new files can be a regex-modified version of the column value. This was the initial motivation for the tool - to be able to split a PSM report into multiple files based on the spectrum file the ID came from.


tools/split_file_to_collection/test-data/make_test_data.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
```
bgruening: @caleb-easterly do you know that planemo will generate you test data with `--update_test_data`?

bgruening:
@caleb-easterly I would rather see this as an extension to the already existing tools. We should not create multiple tools with the same logical task but different parameters. Rather keep one tool and hiding some special parameters under an advanced section.

tools/split_file_to_collection/test-data/make_test_data.sh
caleb-easterly: I didn't! that's a good tip, thanks

caleb-easterly:
Yes, I agree that we shouldn't have multiple tools with the same parameters, but I thought expanding to different file types and splitting methods was best accomplished with a Python script - I didn't see an easy way to expand your tool to fit the task. So I was hoping that this could be the go-to splitting tool :) 

It wouldn't be too hard to add new file types, also.

bgruening:
You are referring to this tool? https://github.com/bgruening/galaxytools/blob/master/tools/splitfasta/splitFasta.xml

It is using already a Python file.
The problem with tools is that it is really hard to deprecate tools, because they could potentially be used in older workflows. So if you want to enhance a tool you can change everything as you like as long as the tool_id stays and the repo owner it will be a new version and not a new tool.

caleb-easterly:
Ah, I didn't see that one. I was actually referring to this one: https://github.com/bgruening/galaxytools/tree/master/tools/text_processing/split_file_on_column 

Yes, I see that it could be better to have fewer tools to maintain...but I'm not sure which one these modifications would go in. It sort of combines the splitFasta and split_file_by_column tools, but wouldn't fit completely in either one.

PR #256, opened 2018-02-12
bug fix - modified inputs

caleb-easterly:
This fixes a pretty nasty bug in which the tool inputs were modified on disk. I'm now copying all inputs to temporary files to be extra safe.

### DIFF
--- a/tools/moFF/moff.xml
+++ b/tools/moFF/moff.xml
@@ -77,7 +77,7 @@
         <token name="@WRANGLE_IDENT_INPUT_SINGLE@"><![CDATA[
             mkdir ./ident_inputs &&
             #if $task.ident_input.input_type_selector == "ps":
-                ln -s '$task.ident_input.ident_input_file' './ident_inputs/$task.ident_input.ident_input_file.display_name' &&
+                cp '$task.ident_input.ident_input_file' './ident_inputs/$task.ident_input.ident_input_file.display_name' &&
             #else
                 cp '$task.ident_input.ident_input_file' ./tempfile1.tab &&
                 ## optionally remove first line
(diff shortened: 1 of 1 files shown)


bgruening:
Please bump the tool version number.
We should also report this upstream. Modifying inputs is a no-go.

caleb-easterly:
Re: version number. I've wondered about this situation before - the version of moff is the same (1.2.1), but the wrapper has changed. Should I call the new version 1.2.1.1 or something like that? 

Yes, this makes me wonder if it's possible to make history items read-only, or something like that.

bgruening:
👍  to 1.2.1.1. You can make dataset inputs readonly, but this needs to be done by the admin of your instance by e.g. using Docker/Singularity ...

tools/moFF/moff.xml
```diff
@@ -421,6 +421,8 @@ It currently uses a Go library to directly extract data from Thermo Raw spectrum
 eliminating the need for conversions from other formats.
 Moreover, moFF also allows one to work directly with mzML files.
 
+This is moFF version @VERSION@.
```
bgruening: Don't do it like this. Use https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-version-command

tools/moFF/moff.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="proteomics_moff" name="moFF" version="@VERSION@">
+<tool id="proteomics_moff" name="moFF" version="1.2.1.1">
```
bgruening: here you can keep `@VERSION@.1`

caleb-easterly: ah, good point.

tools/moFF/moff.xml
caleb-easterly: I thought about this, but I don't think moff has a `moff.py --version`, as far as I can tell. Is this way okay or is there a conda workaround?

bgruening: No conda workaround. But you could echo "@VERSION@". Which is also bad but better then nothing.
moff should implement the version ... :)

PR #263, opened 2018-03-01
OpenMS improvements (MetaProSIP, etc)

bernt-matthias:
Adding the memory is necessary on our cluster and should not hurt otherwise. 

The prefix is important for downstream analyses and should be known by the user.


bgruening:
@bernt-matthias you can also just export the ENV var: ` _JAVA_OPTIONS: -XX:MaxPermSize=2G -Xmx6G -Xms1G -Djava.io.tmpdir=/data`
That does the trick in our case and do not involve tool-hacking.

bernt-matthias:
- Added missing r-gplots requirement to openms_MSGFPlusAdapter. 
- Extended output label of some tools to include the file type. 

@bgruening do you agree that the default output label should be  `${tool.name} on ${on_string}: ${FORMAT}` if the tool has multiple outputs? 

@bgruening just checking the java suggestion .. thanks.

bernt-matthias:
The focus of the PR suddenly changed to MetaProSIP. The program creates a variable number of images and score files which are currently ignored by Galaxy. Our users need this data: so I would implement this. 

I'm wondering if this should be implemented as data collection or just as a zip file... any preferences? Our users spontaneously voted for the zip file. 

A data collection would be fancy if one could download them as zip.

tools/openms/MetaProSIP.xml
```diff
@@ -11,6 +11,7 @@
   <expand macro="stdio"/>
   <expand macro="requirements">
```
bgruening: please increase the tool version here

tools/openms/MetaProSIP.xml
```diff
@@ -182,8 +183,8 @@
     </expand>
   </inputs>
   <outputs>
-    <data name="param_out_csv" format="tabular"/>
-    <data name="param_out_peptide_centric_csv" format="tabular"/>
+    <data name="param_out_csv" format="tabular" label="${tool.name} on ${on_string}: csv"/>
```
bgruening: I would call it tabular. CSV is in Galaxy really a comma separated value file.

tools/openms/MetaProSIP.xml
```diff
@@ -182,8 +183,8 @@
...
+    <data name="param_out_peptide_centric_csv" format="tabular" label="${tool.name} on ${on_string}: peptide centric csv"/>
```
bgruening: here again

bgruening:
> @bgruening do you agree that the default output label should be ${tool.name} on ${on_string}: ${FORMAT} if a tool has multiple outputs?

This is better then before, but I would prefer a more content associated description. So what is included and not what is the name of the XML file. Not sure everyone knows the differences.

> A data collection would be fancy if one could download them as zip.

Data collection, not zip! Galaxy will fix the download of collections soon.

bernt-matthias:
> I would prefer a more content associated description

I'm not an expert of the software/method :) Is there some docs that could help me. Otherwise I could discuss with the users here.

bgruening:
Me either. Leave it like it is. Its a good improvement. At some point if we have a lot Proteomics users we make a Codefest and improve usability.

bernt-matthias:
Here is a first suggestion for the integration of the image output of metaprosip. 

If png/svg is chosen as format MetaProSIP writes a html file which is added as data set. Links to the files seem to work (but for svg they are shown only if the output is shown in a separate window or in the download that is available). 

For pdf files MetaProSIP generates no web page. So I integrated them as collection. 

Since image generation seems to need a considerable time one could also make this optional.

bernt-matthias:
Maybe also add some more details here: 

https://planemo.readthedocs.io/en/latest/writing_advanced.html#cluster-usage
https://galaxyproject.org/admin/config/galaxy_slots/

?

bernt-matthias:
So. More or less finished. Also added some tests. Found small test data at https://github.com/OpenMS/OpenMS/tree/develop/src/tests/topp and the test calls here https://github.com/OpenMS/OpenMS/blob/8fa312ff8d6d5e97aa778283e9ec0d95750d6c5e/src/tests/topp/CMakeLists.txt. Unfortunatelly the test currently fails, probably because the defaults implemented in Galaxy and the values in the command line tools differ (or the test data is for a newer version). Should I just regenerate the test outputs? 

I also stumbled over MzMLSplitter which did not work at all and fixed it. Also here the test is not working .. and I don't know why. There is some errors indicating that mzml file extension is expected but its MzML ... any suggestions?

bernt-matthias:
Just verified that the defaults specified by MetaProSIP --help and the galaxy defaults are the same...

bgruening:
@bernt-matthias I will look at it, but can you please add the test data as well?
These tools are in such a bad shape but we would need to spend so many time on fixing these .... :(

bernt-matthias:
added the test data

bernt-matthias:
@bgruening I have enabled the threads parameter which seems to be available for all 4 tools

The quality of the tools will progress over time... :)

bgruening:
@bernt-matthias with our help I'm sure! I'm just a little bit worried about the next regenerate/merge, but this is a good thing :)

bernt-matthias:
will this go automatically to the toolshed?

bgruening:
In general yes, but as OpenMS do not have tests and takes it does not work. I upload them right now, just not sure if WifiOnICE is really capable of that :)

bgruening:
Done! Its on the TS!

PR #264, opened 2018-03-02
SearchGUI- MSAmanda advanced options available

CarlosHorro:
MSAmanda advanced options tested and available.
Changed "Fragment Tolerance Units" by default from 'ppm' to 'daltons' for comparability reasons and to avoid empty results with default options.


jj-umn:
While we are updating, the comet sparse matrix param needs to be removed.   I can update after my meeting.

bgruening:
@CarlosHorro tool version should be increased and I guess a test for this option would also be nice.

PR #267, opened 2018-03-09
PeptideShaker 1.16.17 & SearchGUI's Tide advanced search options

CarlosHorro:
This PR includes:

- Reference to PeptideShaker 1.16.17 (with a little bugfix in the tool and some bugfixes in the app itself).
- Tide advanced search options. Into them:
- - Tide's file exporting options have been changed to a dropdown list (as only one can be selected at once), with Percolator and MzIdentML formats temporarily disabled because of a bug on current SearchGUI version.
- - Tides's "tide_num_ptms" and "tide_max_spectrum_mz" don't have anymore default values, following SearchGUI app model.
- - Disabled crux output folder field. It was totally useless for the user.
- Search engines window is expanded by default because of its frequent use and following SearchGUI app model.

[I've removed some details about problems with tests as I will post them apart as issues].


tools/peptideshaker/peptide_shaker.xml
```diff
@@ -36,6 +36,18 @@
             -id_params \$cwd/SEARCHGUI_IdentificationParameters.par
             -out \$cwd/peptideshaker_output.cpsx
             -zip \$cwd/peptideshaker_output.zip
+            #set $cleaned_list = str($outputs).split(',')
```
bgruening: @CarlosHorro does this not work anymore? https://github.com/CarlosHorro/tools-galaxyp/blob/25f59f61cf5e089a2c456a4a2b89779507246b93/tools/peptideshaker/peptide_shaker.xml#L129

CarlosHorro: Sorry I'm not sure if I understand the comment... I think it still works :-) do you mean that the following code (related to the 'cleaned_list') can be redundant?

bgruening: I think it is not needed as we filter the other strings already out. Do you see that this is not working without your code?

CarlosHorro: Yes, it gan give an error. As an example, if you choose only "cps" and "PSM report", it gives and error as the "if" condition is true but there is still the "cps" option which need to be cleaned before showing it with the "-reports" option.

bgruening: Could you add a test for this?

bgruening:
Thanks @CarlosHorro!

PR #271, opened 2018-03-24
fix FileMerger bug

bgruening:
@bernt-matthias does this look good to you?

### DIFF
--- a/tools/openms/FileMerger.xml
+++ b/tools/openms/FileMerger.xml
@@ -26,16 +26,8 @@
   -rt_concat:gap $param_rt_concat_gap
 #end if
 
-#if $rep_param_rt_concat_trafo_out:
--rt_concat:trafo_out
-  #for token in $rep_param_rt_concat_trafo_out:
-    #if " " in str(token):
-      "$token.param_rt_concat_trafo_out"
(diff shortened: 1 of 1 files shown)


bernt-matthias:
Sorry, I don't get the context.

bgruening:
No context, just found a bug :)

bernt-matthias:
Definitely it was a bug, but from the documentation this seems not to be the way its supposed to be used http://ftp.mi.fu-berlin.de/pub/OpenMS/release-documentation/html/TOPP_FileMerger.html

The parameter expects a space separated list of output files (same as input files). Maybe something like this: 

```
-rt_concat:trafo_out
  #for token in $param_in:
    $token.trafoXML
  #end for
```

The output needs then to be adapted to the detect the trafoXML files from work dir. 

I would also make this output optional. I **guess** the output is only needed if the gap parameter is >0.

bernt-matthias:
You can see an example call here: https://github.com/OpenMS/OpenMS/blob/3604aab47f89ab2c94de89a938c6144e6dc74faf/src/tests/topp/CMakeLists.txt search for `TOPP_FileMerger_10`

PR #278, opened 2018-05-01
fix output path quoting

jvolkening:
Output paths for `MapAlignerPoseClustering` can contain spaces and must be quoted

### DIFF
--- a/tools/openms/MapAlignerPoseClustering.xml
+++ b/tools/openms/MapAlignerPoseClustering.xml
@@ -22,12 +22,12 @@
 
 -out
 #for $counter,$infile in enumerate($param_in):
-    ./results_out/${infile.element_identifier}.${infile.extension}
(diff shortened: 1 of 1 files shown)


tools/openms/MapAlignerPoseClustering.xml
```diff
@@ -22,12 +22,12 @@
 
 -out
 #for $counter,$infile in enumerate($param_in):
-    ./results_out/${infile.element_identifier}.${infile.extension}
+    './results_out/${infile.element_identifier}.${infile.extension}'
```
bgruening: @jvolkening you might want to do this in addition: 
`#set escaped_element_identifier = re.sub('[^\w\-\s]', '_', str($bam.element_identifier))`

bgruening:
can you move this into your other PR, so we don't get two releases on the TS.

PR #282, opened 2018-05-09
Spectrast

jj-umn:
spectrast_import converts bibliospec .ms2, NIST .msp, or X!Hunter .hlf to a SpectraST splib


tools/spectrast/macros.xml
```diff
@@ -0,0 +1,15 @@
+<macros>
+    <token name="@VERSION@">5.0.0</token>
+    <xml name="requirements">
+        <requirements>
+        <requirement type="package">tpp=5.0.0</requirement>
```
bgruening: that looks wired :) you mean `<requirement type="package" version="5.0.0">tpp</requirement>`

tools/spectrast/spectrast_import.xml
```diff
@@ -0,0 +1,44 @@
...
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements" />
+    <command detect_errors="exit_code"><![CDATA[
+        ## Symlink dataset to name with recognized ext
+        #import re
+        #set $ext = '.' + str($input.extension)
+        #set $input_name = $re.sub('[.](ms2|msp|hlf)$','',$input.display_name.split('/')[-1]) + $ext
+        ln -s '$input' '$input_name' &&
+        mkdir -p $output.files_path &&
```
bgruening: single quotes

tools/spectrast/spectrast_import.xml
```diff
@@ -0,0 +1,44 @@
...
+    <description>Spectral Library</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements" />
+    <command detect_errors="exit_code"><![CDATA[
+        ## Symlink dataset to name with recognized ext
+        #import re
+        #set $ext = '.' + str($input.extension)
+        #set $input_name = $re.sub('[.](ms2|msp|hlf)$','',$input.display_name.split('/')[-1]) + $ext
```
bgruening: you could use $input.element_identifier here ...

tools/spectrast/.shed.yml
```diff
@@ -0,0 +1,11 @@
+owner: galaxyp
```
bgruening: @jj-umn I'm wondering if this should be integrated into Galaxy as part of the datatypes .... automatic conversion ....

tools/spectrast/spectrast_import.xml
jj-umn: @bgruening   I'm not familiar with the element_identifier property.  Are there some good examples I should view?

tools/spectrast/.shed.yml
jj-umn: @bgruening   Maybe.   Since these rely on naming conventions, I worry a bit about errors.
I'm also thinking of adding tools spectrast_create, spectrast_filter, and spectrast_search as well. 
This first tool is for DIA MS workflow DIA-Umpire -> SearchGui/PeptideShaker -> Bibliospec -> SpectraST and then to skyline.

tools/spectrast/spectrast_import.xml
bgruening: Here for example: https://github.com/galaxyproject/tools-iuc/blob/7036343b9ac0a0ffc2ce4f6db465b9298ef05e73/tools/picard/picard_macros.xml#L21

tools/spectrast/.shed.yml
bgruening: Ok, then we keep them here. Interesting ... Skyline in Galaxy or outside?

tools/spectrast/spectrast_import.xml
```diff
@@ -0,0 +1,44 @@
+<tool id="spectrast_import" name="SpectraST Import" version="5.0.0">
```
bgruening: Is this more a convert than an import? Otherwise, looks good!

tools/spectrast/spectrast_search.xml
```diff
@@ -0,0 +1,426 @@
...
+            && cp out/* $output
+        #end if
+    ]]></command>
+    <configfiles>
+        <configfile name="spectrast_params"><![CDATA[#slurp
+libraryFile = $libraryFile
+#if $databaseFile is not None:
+databaseFile = $databaseFile
+#end if
+#if $databaseType is not None:
```
bgruening: This needs to be $general.databaseType, similar the others in the sections.

bernt-matthias:
Can you also take care of https://github.com/galaxyproteomics/tools-galaxyp/issues/446 in this PR?

Issue #269, opened 2018-03-21
SearchGUI: Problems with tests


CarlosHorro:
Hi,

I found some problems when creating tests for SearchGui, some of them also affecting currently existing tests:

- 'fixed_modifications' and 'variable_modifications' (loaded from "macros.xml") params:
- - receive incorrect values from their test params: "carbamidomethyl c" and "oxidation of m" instead of values present in current version of 'searchgui_mods.loc.sample' file ("Carbamidomethylation of C" and "Oxidation of M" respectively.
- - values from 'searchgui_mods.loc.sample' file don't seem to be properly loaded when I've tried it for testing. I always found "None" value instead when checking the value of "$protein_modification_options.fixed_modifications". I tried to change that file's location and its name to simply 'searchgui_mods.loc' but it still does not work.
  With these problems current tests should not work...
- I've tried to check the existence of specific files into the zip returned file, but "decompress=true" option does not seem to work for a compressed file with multiple files and I've also found many "in theory" (none of them did) ways to do it (with "output_collection" + "elements", with "discovered_dataset" or with "files_extra"). Do you know how (or if) can this be done, please?

bgruening:
For tests with loc files, see this one here as an example: https://github.com/galaxyproject/tools-iuc/blob/master/tools/cutadapt/cutadapt.xml#L355

bgruening:
> I've tried to check the existence of specific files into the zip returned file, but "decompress=true" option does not seem to work for a compressed file with multiple files and I've also found many "in theory" (none of them did) ways to do it (with "output_collection" + "elements", with "discovered_dataset" or with "files_extra"). Do you know how (or if) can this be done, please?

The Q I guess is, are we happy with passing the zip file around? If so you are right there is no mechanism I'm aware of that checks for a file inside a archive. This would need to be implemented in Galaxy as a new test-case. However, I hope that just comparing the size, with sim_size, is enough. For every zip file we also get a bunch of standalone files and if they are correct and the zip-file is a few kb/mb in size I hope this is enough to be confident that the tool works.
Keep in mind we are not testing the functional correctness here, but the Galaxy tool correctness.

Does this help at all? :(

bernt-matthias:
Would the `decompress` attribute of `output` tags help?

CarlosHorro:
> Would the `decompress` attribute of `output` tags help?

I mentioned in the first comment that it didn't work...  are there new developments related to it?

[...]

Issue #31, opened 2015-04-27
SearchGUI wrapper should have one and only one decoy database option


iracooke:
There should be one option that controls whether all searches are done with decoy or not. Options to control decoy searches for individual search engines should be removed because they are confusing and would lead to weird mismatched scoring behaviour.

bernt-matthias:
Solved. See https://github.com/galaxyproteomics/tools-galaxyp/pull/283#issuecomment-397654036

PR #283, opened 2018-05-12
WIP: SearchGUI v4.0.7.0, PeptideShaker 2.0.5.1, new Identification Parameters tool, new FastaCLI tool

CarlosHorro:
SearchGUI tool v4.0.4.1
PeptideShaker tool v2.0.5.1
Identification Parameters tool v1.5.1
FastaCLI tool v1.0

----
Main improvements:

--Updated PeptideShaker version to 2.0.5.1.
Internal refactoring, new internal db and more efficient data management
Support for spectrum files in mzML format

--Updated SearchGUI version to 4.0.4.1.
Support for spectrum files in mzML format
Search engines updated to last versions. Options updated accordingly.

-Updated readme.rst

----
Main changes:

SearchGUI tool has been divided into 3 apps:
- New FastaCLI app: receives one FASTA file without decoy sequences and add them to it.
- New Identification Parameters app: generates a PAR file from the settings chosen by the user. Many of them were previously chosen from SearchGUI app, and some others from PeptideShaker app (they have been also removed from it).
- SearchGUI tool itself: all identification parameters stuff have been moved to the new Identification Parameters app. Only search-related functionality itself has been preserved.

Now, PAR files can be now generated independently from SearchGUI and reused in  different SearchGUI executions.

File macros.xml contained some parameters configurations and it has been replaced by macros_basic.xml.

New tests have been developed for the new tools.

----
For a full detail of all applications changes:

SearchGUI release notes:
http://compomics.github.io/projects/searchgui/wiki/ReleaseNotes
PeptideShaker release notes:
http://compomics.github.io/projects/peptide-shaker/wiki/ReleaseNotes


tools/peptideshaker/fasta_cli.xml
```diff
@@ -0,0 +1,63 @@
+<tool id="fasta_cli" name="FastaCLI" version="1.0">
```
bgruening: @CarlosHorro can you please check out this tool: `DecoyDatabase Create decoy protein DB from forward protein DB. (Galaxy Version 2.3.0)` it for example installed on usegalaxy.eu. Does this tool do the same?

bgruening: Or this tool: Create Decoy Database (reverse) Creates a decoy search database by adding reverse sequences to an existing database

tools/peptideshaker/fasta_cli.xml
```diff
@@ -0,0 +1,63 @@
...
+
+        mkdir output;
+        cwd=`pwd`;
+        export HOME=\$cwd;
+
+        ## echo the search engines to run
+        ##echo "$search_engines_options.engines";
+        echo "DB: ${input_database.display_name} sequences: ${input_database.metadata.sequences}";
+
+        cp "${input_database}" input_database.fasta;
```
bgruening: single quotes instead of double-quotes, this applied to all text, or data parameters.

tools/peptideshaker/fasta_cli.xml
```diff
@@ -0,0 +1,63 @@
...
+        ## #set $temp_stderr = "searchgui_stderr"
+        #set $bin_dir = "bin"
+
+        mkdir output;
+        cwd=`pwd`;
+        export HOME=\$cwd;
+
+        ## echo the search engines to run
+        ##echo "$search_engines_options.engines";
+        echo "DB: ${input_database.display_name} sequences: ${input_database.metadata.sequences}";
```
bgruening: single quotes

tools/peptideshaker/fasta_cli.xml
```diff
@@ -0,0 +1,63 @@
...
+    <requirements>
+	<requirement type="package" version="@SEARCHGUI_VERSION@">searchgui</requirement>
+	<requirement type="package" version="3.0">zip</requirement>
+    </requirements>
+    <expand macro="stdio" />
+    <command>
+<![CDATA[
+
+        ## #set $temp_stderr = "searchgui_stderr"
+        #set $bin_dir = "bin"
```
bgruening: can you use `./bin/` here ... or even better remove it here and specify `./bin/` below directly?

tools/peptideshaker/fasta_cli.xml
```diff
@@ -0,0 +1,63 @@
...
+
+        mkdir output;
+        cwd=`pwd`;
+        export HOME=\$cwd;
```
bgruening: If this is really needed, this should be fixed upstream. Relying on a HOME is not good practice.

tools/peptideshaker/fasta_cli.xml
```diff
@@ -0,0 +1,63 @@
...
+
+        cp "${input_database}" input_database.fasta;
+
+        ###########################################
+        ####       Creating decoy database     ####
+        ###########################################
+
+        echo "Creating decoy database.";
+        searchgui eu.isas.searchgui.cmd.FastaCLI --exec_dir="\$cwd/${bin_dir}" -in input_database.fasta -decoy &&
+        rm input_database.fasta;
```
bgruening: You don't need to clean things up as long as you stay in `./` :)

tools/peptideshaker/fasta_cli.xml
```diff
@@ -0,0 +1,63 @@
...
+	<requirement type="package" version="3.0">zip</requirement>
+    </requirements>
+    <expand macro="stdio" />
+    <command>
+<![CDATA[
+
+        ## #set $temp_stderr = "searchgui_stderr"
+        #set $bin_dir = "bin"
+
+        mkdir output;
```
bgruening: can you use `&&` instead of `;` this will fail the tool as soon as one command fails and not silently ignore the error as `;` is doing.

tools/peptideshaker/fasta_cli.xml
```diff
@@ -0,0 +1,63 @@
...
+        cp "${input_database}" input_database.fasta;
+
+        ###########################################
+        ####       Creating decoy database     ####
+        ###########################################
+
+        echo "Creating decoy database.";
+        searchgui eu.isas.searchgui.cmd.FastaCLI --exec_dir="\$cwd/${bin_dir}" -in input_database.fasta -decoy &&
+        rm input_database.fasta;
+        echo "Decoy database:";
```
bgruening: Why do you echo this DB again? This could be removed.

tools/peptideshaker/searchgui.xml
```diff
@@ -3,16 +3,19 @@
         Perform protein identification using various search engines and prepare results for input to Peptide Shaker
     </description>
     <macros>
-        <import>macros.xml</import>
+        <import>macros_basic.xml</import>
     </macros>
     <requirements>
 	<requirement type="package" version="@SEARCHGUI_VERSION@">searchgui</requirement>
 	<requirement type="package" version="3.0">zip</requirement>
+  <requirement type="package">JQ</requirement>
```
bgruening: indentation ... and do you mean the package `jq` here, the JSON parser? If please specify a version as well.

tools/peptideshaker/datatypes_conf.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0"?>
+<datatypes>
+  <registration display_path="display_applications">
+    <datatype extension="par" type="galaxy.datatypes.text:Json" display_in_upload="true" subclass="true" mimetype="application/json">
```
bgruening: This should go upstream into Galaxy. Local datatypes should not be used anymore.
As you subclass JSON here you don't need to specify the parser.
I would also rename this to maybe `searchgui_par`

tools/peptideshaker/searchgui.xml
```diff
@@ -328,15 +94,15 @@
...
+            -id_params $par_target_path
 
             -threads "\${GALAXY_SLOTS:-12}"
 
-            #if $advanced_options.searchgui_advanced.searchgui_advanced_selector == 'advanced'
-                -correct_titles "${advanced_options.searchgui_advanced.correct_titles}"
-                $advanced_options.searchgui_advanced.missing_titles
-                -mgf_splitting "${advanced_options.searchgui_advanced.mgf_splitting}"
-                -mgf_spectrum_count "${advanced_options.searchgui_advanced.mgf_spectrum_count}"
+            #if $searchgui_advanced.searchgui_advanced_selector == 'advanced'
```
bgruening: If you are on it, we need to use single quotes all over here instead of double quotes.

tools/peptideshaker/searchgui.xml
```diff
@@ -36,289 +38,53 @@
...
-            #if $advanced_options.comet.comet_advanced == "yes"
-
-                #if $advanced_options.comet.comet_spectrum.comet_spectrum_selector == "yes"
-                    -comet_min_peaks ${advanced_options.comet.comet_spectrum.comet_min_peaks}
-                    -comet_min_peak_int ${advanced_options.comet.comet_spectrum.comet_min_peak_int}
 
-                    -comet_remove_prec ${advanced_options.comet.comet_spectrum.comet_prec.comet_remove_prec}
 
+        ## copy the input .par file to the working folder
+        cp "${input_par}" "\$cwd/SEARCHGUI_IdentificationParameters.par";
```
bgruening: `cp '${input_par}' './SEARCHGUI_IdentificationParameters.par' &&`

tools/peptideshaker/searchgui.xml
```diff
@@ -3,16 +3,19 @@
...
 	<requirement type="package" version="@SEARCHGUI_VERSION@">searchgui</requirement>
 	<requirement type="package" version="3.0">zip</requirement>
+  <requirement type="package">JQ</requirement>
     </requirements>
     <expand macro="stdio" />
     <command>
 <![CDATA[
         #from datetime import datetime
+        #import json
+        #import os;
```
bgruening: `;` not needed

tools/peptideshaker/ident_params.xml
```diff
@@ -0,0 +1,1259 @@
...
+        cwd=`pwd`;
+        export HOME=\$cwd;
+
+        echo "DB: ${input_database.display_name} sequences: ${input_database.metadata.sequences}";
+
+        #####################################################
+        ## generate IdentificationParameters for SearchGUI ##
+        #####################################################
+        (searchgui eu.isas.searchgui.cmd.IdentificationParametersCLI
+            --exec_dir="\$cwd/${bin_dir}"
```
bgruening: simply `./bin/`

tools/peptideshaker/ident_params.xml
```diff
@@ -0,0 +1,1259 @@
...
+                -updateGeneMapping 0
+            #end if
+
+
+            ## TODO: PROTEIN INFERENCE
+
+
+            ## TODO: VALIDATION LEVELS
+
+            -protein_fdr "${advanced_options.validation_levels_options.protein_fdr}"
```
bgruening: single quotes

tools/peptideshaker/ident_params.xml
```diff
@@ -0,0 +1,1259 @@
...
+                    </param>
+                    <when value="no" />
+                    <when value="yes">
+                        <param name="ms_amanda_decoy"  type="boolean" truevalue="1" falsevalue="0" checked="false"
+                            label="MS Amanda: Generate Decoys" help="generate decoys" />
+                        <param name="ms_amanda_instrument" label="MS Amanda: instrument" type="select"
+                          help="MS Amanda instrument id option. Available ion types are listed here.">
+
+                          <option value="b, y" selected="True">b, y</option>
+
```
bgruening: This is not going to work on a cluster system with multiple object stores. You can not rely on the FASTA file is always next to the PAR file :(
In SG and PS you need to specify also the FASTA file, then link everything in `./` and use it.

CarlosHorro: Hi Bjorn, you`ve commented many things I will try to manage step by step.
About this point, I`m not totally sure about what you mean here (and, in fact, the Identification Parameters tool is a key step we need to work perfectly in the workflow). Is your comment doing reference to the tests or to the tool itself?
Talking about the tool itself, I think there should not be any problem. The PAR file is created by the Identification Parameters tool doing reference to the URL of a FASTA file present in the Galaxy history, so it should be present and available using this URL, am I wrong?
The PAR files I included in test-data are different, sure. I didn`t find a different way to test the tool than copying the FASTA to the ./ folder in order to find it.

bgruening: Its about the tool itself.
> Talking about the tool itself, I think there should not be any problem. The PAR file is created by the Identification Parameters tool doing reference to the URL of a FASTA file present in the Galaxy history, so it should be present and available using this URL, am I wrong?

So its not a relative path but on absolute path? This will work in more places, but still not in all. I can imaging cases where job staging is involved where this can break.

In general on of the biggest problem with Proteomic files is that they are referencing some `paths` which can change in various way in HPC/Cloud environments.

Another concern is that people share the param file (to other users), but delete the history or the FASTA file and then PS/SG will not work anymore. This is really counter-intuitive I guess.

tools/peptideshaker/fasta_cli.xml
CarlosHorro: I checked them but they don`t work properly with our tools as we require the headers to be set up in such a way that we can parse them properly in PeptideShaker, and the best way to ensure this is to use FastaCLI to add the decoys.

tools/peptideshaker/ident_params.xml
CarlosHorro: I`ve talked with Harald about this and we think we can solve this problem if we change the output of the Identification Parameters tool to generate a zip file that includes both the parameters and the FASTA file, in a similar way than PeptideShaker uses a zip file with the output of SearchGUI (with the FASTA, spectrum and parameters files). That way this zip file can be used as input to SearchGUI and the problem of the path is avoided in the same was as the zip file input to PS?

bgruening: Sounds good!

tools/peptideshaker/fasta_cli.xml
bgruening: That is unfortunate, but I guess nothing that we can avoid. Thanks for trying!

tools/peptideshaker/fasta_cli.xml
CarlosHorro: I think it is not necessary in this tool, but not sure about the others. I will try to check them.

jj-umn:
@bgruening @CarlosHorro I'll take a look at this too, since we are quite dependent on SG/PS

jj-umn:
@CarlosHorro One other thing that would be good to add is additional enzymes.  The EnzymesCLI is in place, but I had been waiting for the updated compomics branch before I added it to the search_gui tool.  One of our researchers was digesting samples with elastase to retrieve peptides that were being lost at the trypsin cut sites.

CarlosHorro:
@jj-umn Thank you for the info, I will talk with Harald about it.

PratikDJagtap:
Hello Carlos,

The Galaxy-P team  ( @jj-umn, @subinamehta @jraysajulga @pravs3683 @caleb-easterly @chambm @tmcgowan, Tim Griffin ) discussed this today at our meeting and we are planning to test this next week and provide questions if we have any. Our questions / suggestions will be more on usability and implementation in our current workflows.
A few questions:
1) Is this change essential for the implementation of downstream tools such as REPORTER?
2) Last year when we discussed about implementing downstream tools (REPORTER, moFF, etc.) with Lennart Martens- he mentioned that SearchGUI / PeptideShaker was going to go through a backend overhaul. Can you or your lab shed light on when is the anticipated date for the overhaul?

We will send you more questions / inputs next week. We can also discuss these in details if needed.

Thanks and Regards,
Pratik

tools/peptideshaker/macros.xml
```diff
@@ -1,179 +0,0 @@
-<macros>
-    <xml name="stdio">
-        <stdio>
-            <exit_code range="1:" level="fatal" description="Job Failed" />
-            <regex match="java.*Exception" level="fatal" description="Java Exception"/>
-            <regex match="Could not create the Java virtual machine" level="fatal" description="JVM Error"/>
```
bernt-matthias: I just stumbled over the following error in the current version of the SearchGUI tool: 

```
Error occurred during initialization of VM
Could not reserve enough space for 6291456KB object heap
```

Can you add another regex match for recognizing the out of memory condition? A question for @bgruening are the regex conditions processed in order .. and is the exit_code condition checked before the regex conditions? Then you would need to add it in front of the `Could not create the Java virtual machine` regex. 

If you would not like to include this in this PR I'm happy to open a separate PR.

bernt-matthias: FYI: just use level="fatal_oom"

CarlosHorro: Thank you for the suggestion, I will add it to macros_basic.xml as soon as possible (macros.xml will disappear after this PR).
By the way, as you suggested, the exit_code condition is checked before regex:

"" (from https://docs.galaxyproject.org/en/latest/dev/schema.html):
A note should be made on the order in which exit codes and regular expressions are applied and how the processing stops. Exit code rules are applied before regular expression rules. The rationale is that exit codes are more clearly defined and are easier to check computationally, so they are applied first. Exit code rules are applied in the order in which they appear in the tool’s configuration file, and regular expressions are also applied in the order in which they appear in the tool’s configuration file. However, once a rule is triggered that causes a fatal error, no further rules are checked.
""

CarlosHorro:
Hi @PratikDJagtap , answering to your questions:
- Current changes will affect our internal workflow in: database selection, parameters generation and SG, but it will not affect downstream tools which use PS output.
- About the backend overhaul, its refactoring is done, and the code is also compiling and running, but we are still testing all its features. I'm afraid we cannot still say when this phase will finish as there is a huge amount of functionality to be tested.

Greetings,
Carlos

tools/peptideshaker/macros.xml
bernt-matthias: In this case (exit_code is checked before regex) the regex tests seem to make no sense for this tool since searchgui seems to return exit code 1 anyway (at least for the OOM error). 

@bgruening any comment on that? I guess in the light of the recently introduced OOM checks, testing for an exit_code first seems to be not the best idea...?

CarlosHorro:
It does not give me problems in my local machine with my v0.48... I don't know what planemo version uses the Travis building system here...

CarlosHorro:
Checking https://github.com/galaxyproject/planemo/blob/master/planemo/xml/xsd/tool/galaxy.xsd, I think there is a contradiction between the documentation itself about regular expressions and the accompanying example.
Doc related to regular expressions only talks about "log", "warning" and "fatal" error levels:
--
A regular expression includes the following attributes:
* ``source``: This tells whether the regular expression should be matched against stdout, stderr, or both. If this attribute is missing or is incorrect, then both stdout and stderr will be checked. The source can be one of the following values:
  * ``stdout``: the regular expression will be applied to stdout;
  * ``stderr``: the regular expression will be applied to stderr;
  * ``both``: the regular expression will be applied to both stderr and stdout (which is the default case).
* ``match``: This is the regular expression that will be used to match against stdout and/or stderr. If the <regex> tag does not contain the match attribute, then the <regex> tag will be ignored. The regular expression can be any valid Python regular expression. All regular expressions are performed case insensitively. For example, if match contains the regular expression "actg", then the regular expression will match against "actg", "ACTG", "AcTg", and so on. Also note that, if double quotes (") are to be used in the match attribute, then the value " can be used in place of double quotes. Likewise, if single quotes (') are to be used in the match attribute, then the value ' can be used if necessary.
* ``level``: This works very similarly to the <exit_code> tag, except that, when a regular expression matches against its source, the description is added to the beginning of the source. For example, if stdout matches on a regular expression, then the regular expression's description is added to the beginning of stdout (instead of stderr). The level can be log, warning or fatal as described below.
  * ``log`` and ``warning``: If the regular expression matches against its source input (i.e., stdout and/or stderr), then a description of the error will be added to the beginning of the source, prepended with either 'Log:' or 'Warning:'. A log-level/warning-level error will not cause the tool to fail.
  * ``fatal``: If the regular expression matches against its source input, then a description of the error will be added to the beginning of the source. A fatal-level error will cause the tool to fail. If no level is specified, then the fatal error level will be assumed to have occurred.
* ``description``: Just like its ``exit_code`` counterpart, this is an optional description of the regular expression that has matched.
--

But the example also talks about fatal_oom level:

```xml
<stdio>
    <regex match="low space"
           source="both"
           level="warning"
           description="Low space on device" />
    <regex match="error"
           source="stdout"
           level="fatal"
           description="Unknown error encountered" />
    <!-- Catching fatal_oom allows the job runner to potentially resubmit to a resource with more
         memory if Galaxy is configured to do this. -->
    <regex match="out of memory"
           source="stdout"
           level="fatal_oom"
           description="Out of memory error occurred" />
    <regex match="[CG]{12}"
           description="Fatal error - CG island 12 nts long found" />
    <regex match="^Branch A"
           level="warning"
           description="Branch A was taken in execution" />
</stdio>
```

tools/peptideshaker/fasta_cli.xml
```diff
@@ -0,0 +1,60 @@
...
+    <command>
+<![CDATA[
+        #set $bin_dir = "bin"
+
+        mkdir output;
+        cwd=`pwd`;
+
+        echo 'DB: ${input_database.display_name} sequences: ${input_database.metadata.sequences}';
+
+        cp '${input_database}' input_database.fasta;
```
bernt-matthias: Wouldn't linking work as well here?

CarlosHorro: Of course, linting was not working on any tool as macros_basic.xml is used by them all.

tools/peptideshaker/fasta_cli.xml
bernt-matthias: Lets talk about FastaCLI once more. I was just using the SearchGUI version that is currently installed via the TS and realised that FastaCLI uses GBs of memory for a MBs size data base. I don't know if the problem persists in the latest version .. but it appears that the program is not the best to use. 

OpenMS includes the tool: "DecoyDatabase Create decoy protein DB from forward protein DB." here you can configure the header (append or prepend a configurable string). Would this be a good replacement?

CarlosHorro: I'm afraid it isn't, as I said I tested it and it adds the string in different positions than the one we need in PS.

bgruening: playing the devel here ... can we adopt PS to support this other format?

CarlosHorro: After talking with Harald he confirmed we need FastaCLI output at the moment, although we will try to extend that support in the future.
Also, we agreed FastaCLI cannot use GBs of memory for a MBs size database, it must be another different process. I've also done some tests with FastaCLI using uniprot-human-reviewed-trypsin-june-2017_concatenated_target_decoy.fasta and, as expected, it consumes less than 100 mb ram.

CarlosHorro:
I think this PR is ready :-)

CarlosHorro:
May this PR be merged, please?

bgruening:
@jj-umn @bernt-matthias anyone?
I can review this at the weekend maybe. But my biggest concern is about the new paradigm shift in the tools, which I can not really judge.

bernt-matthias:
I guess I can't help with the paradigm shift. I have to little experience with these tools.

jj-umn:
@CarlosHorro @bgruening @PratikDJagtap @pravs3683 @subinamehta I'll try to review agin this weekend.  I've also asked our frequent users to comment as well.   This will make every use of PeptideShaker a 4 step workflow: fasta_cli -> ident_params -> search_gui -> peptide_shaker.   This may not change our mode of operation too much, since we usually placed search_gui -> peptide_shaker in a workflow since the versions of each were often tightly coupled.   

Sorry about the delay in reviewing this, I've been dealing with some medical issues.  Seems turning age 65 and dealing with a mid-April 42 cm snow storm were not compatible.

bernt-matthias:
Does this one also affect the new version https://github.com/galaxyproteomics/tools-galaxyp/issues/293? Any ideas how to fix this? 

while searching the galaxyp issues I also stumbled over these: 

https://github.com/galaxyproteomics/tools-galaxyp/issues/269
https://github.com/galaxyproteomics/tools-galaxyp/issues/31
https://github.com/galaxyproteomics/tools-galaxyp/issues/15

CarlosHorro:
In order to move forward this PR, I ask for not mixing it with issues which are previous or not related to it, please.
About each specific topic:

#293: You're totally right, this version may be affected by it (although it didn't failed when testing it). 
-temp_folder should be specified in order to avoid any issue locating the preferences file.
I will fix it asap.
#269: This is related to a limitation of Galaxy checking results of tests with compressed files. Not sure if it will managed in the medium term.
#31: This is a very old issue related to the app itself; however, there is only one decoy option (if the fasta file has decoys or not) in the wrapper, so it can be considered as solved.
#15: It was already solved so it will have to be closed.

CarlosHorro:
Hi @bgruening I don`t know why Travis is giving now this error, which is not related at all with the changes of that commit :

RunToolException: Error creating a job for these tool inputs - Parameter variable_modifications requires a value, but has no legal values defined., Parameter fixed_modifications requires a value, but has no legal values defined.


It seems related to http://dev.list.galaxyproject.org/No-legal-value-defined-in-a-planemo-test-for-param-type-quot-select-quot-and-options-quot-from-file--td4671247.html ?

bernt-matthias:
As soon as it is in the TS / TTS I would have a user who could test it.

CarlosHorro:
Any clue about this?  It totally seems an internal planemo problem but I don`t have any idea about why it is arising now (and not before) and how to bypass it in a simple way...
I`m really tempted to comment the second test in order to temporarily avoid this problem and manage it later in a different PR loading data via DB or similar.

bgruening:
My guess is you are missing the sample files: tools/peptideshaker/searchgui_mods.loc.sample for example.

CarlosHorro:
That test was failing before I changed the name of that file from "searchgui_mods.loc.sample" to "searchgui_mods.loc"  (in a desperate attempt to make planemo finds that file :-) )
Now I recovered "searchgui_mods.loc.sample" name an the test is still failing...

bgruening:
I know what is happening, but I need more time to look why this is happening.

But the error is still that both select boxes are not populate with any values from the loc files. But the param needs at least one value and fails.

jj-umn:
@CarlosHorro @bgruening @PratikDJagtap   We finally had a meeting today at University of Minnesota to discuss the refactoring of the searchgui peptideshaker tools.   

The general consensus was that having a separate indent_param tool was indeed a good idea.   
The one change that user's preferred however was to still have the fasta db be a separate input to searchgui and to pass in the ident_params as a JSON file.   The reasoning was that this would make the ident_params settings more reusable.   We realize that this would require the searchgui tool wrapper to merge the fasta path into the input ident_params file.   While that may not seem ideal from the programmer's perspective, it seemed more useful and intuitive to the users of the tools.  

There are two more things we'll eventually want to be able to set:
1. User defined enzyme - our current interest was in using elastase for a complementary mass spec search along with the standard trypsin digestion run.
2. User defined modifications

Thanks for all the effort and we apologize for the delay in evaluating this.

CarlosHorro:
Hi all,

I've had a conversation with Harald about last topics commented by @jj-umn .
Summarising it, the new backend will cover all main features mentioned by JJ, and specially the fasta file will be apart from the input params.
So, we prefer to wait for its launch to make this PR finally accepted and published.

While this happens, I will open separate PRs to update current SG and PS galaxy tools to the last SG and PS apps versions, keeping the previous workflow.

bernt-matthias:
Does anyone has an idea about this one: https://github.com/galaxyproteomics/tools-galaxyp/issues/293 -- which happens when running as real user. Would be great to have a workaround / fix in future versions.

bernt-matthias:
Just got confused: 

The current wrappers have peptide-shaker 1.16.26 and search_gui 3.3.3. which seems to be newer than the ones in this PR.

CarlosHorro:
Hi @bernt-matthias ,

Tool versions managed in this PR (which is still "frozen") are not still linked to the new backend app versions. They will be in the future, asap.

Greetings,
Carlos

CarlosHorro:
Hi @bgruening , there are still bugs and things to be improved in the new backend that will require time to be solved, so the old backend should still be used.

bgruening:
Thanks @CarlosHorro for the update! I will keep fixing bugs in the current version then.

bgruening:
@CarlosHorro this really needs a rebase against master :)

CarlosHorro:
Just for testing all new changes, I've uploaded to
https://anaconda.org/CarlosHorro/searchgui

the SearchGUI v4.0.1_SNAPSHOT version.

CarlosHorro:
Hi @bgruening, I have a doubt maybe you can help me with.
New IdentificationSettings application has an option "-id_params" which receives a previously generated .par object taking it as base for adding or changing options.
I do not know if there is an easy way to make all options in a Galaxy tool "optional", so they are not passed by default (unless the user explicitly introduces them) when the tool receives this .par file.

bgruening:
@CarlosHorro is this an all or nothing param file? I mean does this file then contain all values or can/should individual values be overwritten?

The former is doable, the latter is problematic and I don't think a Galaxy user gains much from it, so I would skip this option.

CarlosHorro:
> @CarlosHorro is this an all or nothing param file? I mean does this file then contain all values or can/should individual values be overwritten?
> 
> The former is doable, the latter is problematic and I don't think a Galaxy user gains much from it, so I would skip this option.

The parameters file may indeed have all required parameters for a SG/PS execution. I though it could greatly facilitate replicability, just sharing the .par file and reusing it in order to execute SG/PS with the same configuration.
And, if any specific modification on that configuration is required, I though to reuse that file and just to specify the new parameters on the IdentificationParameters tool in order to overwrite only them and not other parameters previously present on the file. Does it make sense to you?

bgruening:
> The parameters file may indeed have all required parameters for a SG/PS execution. I though it could greatly facilitate replicability, just sharing the .par file and reusing it in order to execute SG/PS with the same configuration.

You could do this by having one giant condition (use param file, defaulting to No.).

> And, if any specific modification on that configuration is required, I though to reuse that file and just to specify the new parameters on the IdentificationParameters tool in order to overwrite only them and not other parameters previously present on the file. Does it make sense to you?

That makes sense, I'm just wondering how this can work. You assume that all parameter files are in the param file and then ... which params are you writing back? All that are different from the default that Galaxy offers? But maybe you want the default to be written back? I guess offering to use a param file is cool, modifying one seems complicated to me. And maybe it's just easier to create a new one from scratch.

jj-umn:
@CarlosHorro @bgruening   I think a param file could be useful.  I would probably implement this with a default param file in the tool source, an optional tool param input file, and the optional tool param fields.   These would be parsed in that order into a dictionary, and then that dictionary written to the param file that would be used by the application and also copied as a tool output.   This can be done within a <configfile> tag using cheetah.
I think had to do something similar to that for defuse and dia_umpire tools:
https://github.com/jj-umn/galaxytools/blob/defuse8/defuse/defuse.xml
https://github.com/galaxyproteomics/tools-galaxyp/blob/master/tools/dia_umpire/dia_umpire_se.xml

CarlosHorro:
Thank you all for your feedback, after considering your suggestions, the considerable modifications required to support this case and that it is not really a requirement, the tool will support just receiving spare params or the input file following @bgruening suggestion, but not the combined case.
However, @jj-umn tools are really interesting implementations to have in mind for future tools.

bgruening:
@CarlosHorro please let me know when you need a review! I'm so looking forward to deploy it.

All fine with family?

CarlosHorro:
> @CarlosHorro please let me know when you need a review! I'm so looking forward to deploy it.
> 
> All fine with family?

Hi @bgruening ,

To be honest, precisely this last one was the last big update to the tools I had in mind, apart from bugfixes and, of course, all the feedback by the galaxyp team :-)
I think all tools are mature enough to be reviewed by you (one developer of our team is already using them for his workflow), but, when I tried to upload the conda package I was said that, as they are still called "SNAPSHOT" they cannot be made Conda packages via BioConda and I had to wait until their final versions, so, for my daily work I created and own channel (carloshorro) which I'm using to make different tests, but of course it's not being used by galaxyp's travis and because of this the tools give errors here...

About my family, everything is going very well thank you :-)  My baby is already 3 months old and growing really fast and healthy ^_^  It's being amazing to be able to work 50% and being on paternal leave another 50%... :-)

bgruening:
> About my family, everything is going very well thank you :-) My baby is already 3 months old and growing really fast and healthy ^_^ It's being amazing to be able to work 50% and being on paternal leave another 50%... :-)

Ha, that sounds super cool!

@CarlosHorro do you have any ETA when there will be an official release?

@jj-umn ready for review? :)

CarlosHorro:
> @CarlosHorro do you have any ETA when there will be an official release?

I'm afraid we don't have it yet. We were close to launch a beta version some weeks ago, but new bugs raised and we postponed it.
However, whilst results in the app (from complex datasets) are still being tuned, all app bugs (just 5 at this point) are related to specific conditions rarely present on testing datasets; apart from this, I've been using the tool for some months and our other developer has been also using it, so I think the tool functionality itself can be reviewed confidently at this point.
On the other hand, I will be working at 50% until September (included), but will be on 100% paternal leave on October until February, so I think it would be great to move forward soon as much work like this as possible because this topic may be frozen for some time later...

CarlosHorro:
Um... no idea about what means the error:

```
curl: (7) Failed to connect to cli-assets.heroku.com port 443: Connection timed out
xz: (stdin): File format not recognized
tar: Child returned status 1
tar: Error is not recoverable: exiting now
Failed to update Heroku CLI
```

bgruening:
Where do you see this error?

CarlosHorro:
> Where do you see this error?

In TravisCI details:
https://travis-ci.org/CarlosHorro/tools-galaxyp/jobs/568786106

bgruening:
> AssertionError: Test file (fastacli_searchgui_tinydb1.fasta) is missing. If you use planemo try --update_test_data to generate one.

bgruening:
@CarlosHorro I restarted the tests! What is needed to get this pushed over the line. I have a few users that are eagerly looking forward to the new release :)

CarlosHorro:
> @CarlosHorro I restarted the tests! What is needed to get this pushed over the line. I have a few users that are eagerly looking forward to the new release :)

Thanks @bgruening ! Of course, tests will fail at least until we have the bioconda package ready, but whilst that happens I'm working in parallel with my own conda channel on the necessary rebuilt of current galaxy tools' test files and some little last-minute additions :-)  So I hope everything will start to pass tests soon! :-)

bgruening:
Oh, is that because of .NET and mono? We could migrate searchGUI/PS to conda-forge and see if it works over there ... if its urgent.

CarlosHorro:
> Oh, is that because of .NET and mono? We could migrate searchGUI/PS to conda-forge and see if it works over there ... if its urgent.

I think it is just because of, as you said, the glibc version required to compile dotnet dependency. Don't worry, we are not in a hurry at least for this month, we can just disable dotnet dependency for having the package ready in bioconda as usual and add it later when the building platform is updated.
I will update this night the PR in bioconda with the last bugfix versions of the apps and the removed dotnet dependency :-)

CarlosHorro:
@bgruening After this last commit, all tools and tests should work when the bioconda package is approved.

bgruening:
grafik

One test file is missing, a few size-differences in you test data. And fasta CLI has: `galaxy.tool_util.verify.interactor.JobOutputsError: Dataset metadata verification for [file_ext] failed, expected [fasta] but found [None].`

But the way you can now find this HTML as github artefact and can download it from the Actions page.

@CarlosHorro can you remove all the zip/archive outputs from the repo and use the new `has_size` attribute to check for a filesize: https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-tests-test-output-assert-contents

This will shrink your test data footprint.

Thanks.

CarlosHorro:
Thank you @bgruening , I will remove some test data, although I will have to preserve some of them as they are the input for executing some tests.

CarlosHorro:
Hi @bgruening ,

I've removed as many files as I can (ie: files which are not input of any tool). I hope it is enough...
Doing some research about the last 3 remaining searchgui test failures, I think the cause is this:

```
makeblastdb command: 
/tmp/tmppc5rq93h/job_working_directory/000/4/working/bin/resources/makeblastdb/linux/linux_64bit/makeblastdb -in input_fasta_file.fasta -dbtype prot 

Sun Nov 29 21:50:20 GMT 2020 Formatting input_fasta_file.fasta for OMSSA.

/tmp/tmppc5rq93h/job_working_directory/000/4/working/bin/resources/makeblastdb/linux/linux_64bit/makeblastdb: error while loading shared libraries: libz.so.1: cannot open shared object file: No such file or directory
```

and this:

```
xtandem command: 
/tmp/tmppc5rq93h/job_working_directory/000/4/working/bin/resources/XTandem/linux/linux_64bit/tandem /tmp/tmppc5rq93h/job_working_directory/000/4/working/bin/resources/XTandem/linux/linux_64bit/input_searchGUI.xml 

Sun Nov 29 21:50:20 GMT 2020 Processing searchgui_tinyspectra1.mgf with X!Tandem.

/tmp/tmppc5rq93h/job_working_directory/000/4/working/bin/resources/XTandem/linux/linux_64bit/tandem: error while loading shared libraries: libstdc++.so.6: cannot open shared object file: No such file or directory

```

These problems make OMSSA and XTandem results empty, which causes the difference respect to the expected output file size.
Due to this, I added to SearchGUI 2 dependencies which supposedly fixed both errors:

```
    <requirement type="package" version="1.2.11">zlib</requirement> <!-- required by OMSSA -->
        <requirement type="package">libstdcxx-ng</requirement> <!-- required by XTandem -->
```
But the problem is still there, so I don't know if I'm missing something or/and I should add them to the Conda packages themselves...
Of course, I run all tests in my local successfully.

Greetings,
Carlos

bgruening:
@CarlosHorro makeblastdb and XTandem are coming as pre-compiled binary from SG and PS I assume? Than this is the problem. Xtandem is missing the libstdc++.so.6 libraries and makeblastdb was compiled again an libz that does not exist :+1: 

This is now a problem as we made our testing stricter and are now running tests in containers, previously at least the zlib error would have been hidden in conda.

CarlosHorro:
> @CarlosHorro makeblastdb and XTandem are coming as pre-compiled binary from SG and PS I assume? 

Yes, they come as precompiled libraries.

Than this is the problem. Xtandem is missing the libstdc++.so.6 libraries and makeblastdb was compiled again an libz that does not exist 👍
> 
> This is now a problem as we made our testing stricter and are now running tests in containers, previously at least the zlib error would have been hidden in conda.


Do you think that adding those dependencies to the SG bioconda recipe would solve the problem?

bgruening:
I think so. You can test this by adding it to the tool before if you like. You can also test it locally with `planemo test --docker`

CarlosHorro:
> I think so. You can test this by adding it to the tool before if you like. You can also test it locally with `planemo test --docker`

@bgruening I have already added both dependencies to the SearchGUI tool:
```

<requirement type="package" version="1.2.11">zlib</requirement> <!-- required by OMSSA -->
        <requirement type="package">libstdcxx-ng</requirement> <!-- required by XTandem -->

```

But the failure still happens.
Now I've been struggling with different requirements for the Conda package, but I cannot get it to work. I tried things like:

```
requirements:
  run_exports:
    strong:
      - zlib
      - libstdcxx-ng
  build:
    - {{ compiler('c') }}
  run:
    - openjdk >=11
    - mono >=5.0.0
    - python
    - zlib
    - libstdcxx-ng
```

But, as I'm not an expert, I'm probably missing things...

By the way, building the Conda package I also saw these messages:
```

WARNING (searchgui,share/searchgui-4.0.5-1/resources/makeblastdb/linux/linux_64bit/makeblastdb): $RPATH/libbz2.so.1 not found in sysroot, is this binary repackaging? .. do you need to use install_name_tool/patchelf?
   INFO (searchgui,share/searchgui-4.0.5-1/resources/makeblastdb/linux/linux_64bit/makeblastdb): Needed DSO lib/libgcc_s.so.1 found in defaults::libgcc-ng-9.1.0-hdf63c60_0
   INFO (searchgui,share/searchgui-4.0.5-1/resources/makeblastdb/linux/linux_64bit/makeblastdb): Needed DSO x86_64-conda_cos6-linux-gnu/sysroot/lib/libnsl.so.1 found in CDT/compiler package defaults::gcc_impl_linux-64-7.3.0-habb00fd_1
WARNING (searchgui,share/searchgui-4.0.5-1/resources/makeblastdb/linux/linux_64bit/makeblastdb): $RPATH/libz.so.1 not found in sysroot, is this binary repackaging? .. do you need to use install_name_tool/patchelf?
```

But I also saw that patchelf is being installed...

bgruening:
try adding `{{ compiler('cxx') }}` as build dependency

can you depend on the blast package to get makeblastdb instead

CarlosHorro:
> can you depend on the blast package to get makeblastdb instead
@bgruening I don't think so :-(  Most of our users use the application directly downloading its zip file from the webpage, without Conda. So we cannot make the app internally depends on Conda.

> try adding {{ compiler('cxx') }} as build dependency

It still does not work with that, but I've realised that all libraries I need are found in "/usr/local/lib" instead of the default libraries folders.
But, there is no ldconfig utility installed and doing this in build.sh does not work too:
`
export LD_LIBRARY_PATH=/usr/local/lib
`

I guess this problem could be solved by installing ldconfig (not sure how to do that) or installing the libraries directly into /usr/lib (is it possible?)? 
As a last resort of course I could copy/link the required libraries from /usr/local/lib to /usr/lib in build.sh, but that does not seem recommendable I guess...

bgruening:
Wiat, /usr/local/lib is only true for the containers. Its more or less a conda environment installed into /usr/local/ and shipped as containers. Another solution is to actually fix those recompiled binaries. Where do you generate those? Why do they not contain zlib as well? These pre-packages binaries are killing us. We could skip the container-tests, but this would mean that the Galaxy tools can not run in containers, only conda environments :(

CarlosHorro:
> Wiat, /usr/local/lib is only true for the containers. Its more or less a conda environment installed into /usr/local/ and shipped as containers. Another solution is to actually fix those recompiled binaries. Where do you generate those? Why do they not contain zlib as well? These pre-packages binaries are killing us. We could skip the container-tests, but this would mean that the Galaxy tools can not run in containers, only conda environments :(

Oh, ok, I was reviewing the content of the created containers and I though it was comparable to the conda environments.
As you mention, I will try to include the required libraries into the pre-packages...

CarlosHorro:
Hi @bgruening , I've been struggling a lot with this (with xtandem-related problem; OMSSA is really old and we can manage to lose support for it into containers) without success. I've tried several things:


- Static linking: I removed errors related to libstdc++ or libexpat, but it showed instead

```

/lib/libc.so.6: version `GLIBC_2.25' not found
/lib/libm.so.6: version `GLIBC_2.27' not found
/lib/libm.so.6: version `GLIBC_2.29' not found
```
I compiled it into my main Linux VM, with Ubuntu 20 and a modern glibc, so I supposed I had to use some Linux with glibc2.12, like Centos6x . I tried there, but as gcc is older it lacks some features (-static flag) and I was not able to compile xtandem with the libraries.


- Dynamic linking: whenever I try this, libstdc++ or libexpat is not found.
As an example, XTandem zip comes with a zip with a dynamically linked version. Using it and including in the conda recipe:
```

requirements:
  run_exports:
    strong:
      - zlib
      - libstdcxx-ng
      - expat
  build:
    - {{ compiler('c') }}
    - {{ compiler('cxx') }}
  host:
    - zlib
    - libstdcxx-ng
    - expat
  run:
    - openjdk >=11
    - mono >=5.0.0
    - python
    - zlib
    - libstdcxx-ng
    - expat

```

I still get

`
/tmp/tmptvy3v7si/job_working_directory/000/20/working/bin/resources/XTandem/linux/linux_64bit/tandem: error while loading shared libraries: libexpat.so.1: cannot open shared object file: No such file or directory
`
When compiling xtandem on my own with the default flags I also gets this error.


So, I don't know if I'm still missing something...
If cannot make SG to work in containers, we will have to, at least, have it working as usual, so containerised tests could be removed or, if it is easier (I guess so?) I can remove SG tests related to XTandem and OMSSA at least until we find a solution...

Greetings,
Carlos

bgruening:
> I compiled it into my main Linux VM, with Ubuntu 20 and a modern glibc, so I supposed I had to use some Linux with glibc2.12, like Centos6x . I tried there, but as gcc is older it lacks some features (-static flag) and I was not able to compile xtandem with the libraries.


Use the centos container but conda compilers, does this work? Essentially the Bioconda build-boxes.

bgruening:
> If I cannot make SG to work in containers, we will have to, at least, have it working as usual, so containerised tests could be removed or, if it is easier (I guess so?) I can remove SG tests related to XTandem and OMSSA at least until we find a solution...

We can skip container tests if this is really needed. We have a skip list for that. You call in the end - we are happy for all the work that you put in here and don't want to be a pain in the neck.

CarlosHorro:
> > I compiled it into my main Linux VM, with Ubuntu 20 and a modern glibc, so I supposed I had to use some Linux with glibc2.12, like Centos6x . I tried there, but as gcc is older it lacks some features (-static flag) and I was not able to compile xtandem with the libraries.
> 
> Use the centos container but conda compilers, does this work? Essentially the Bioconda build-boxes.

Um... I haven't found information about Bioconda build-boxes, just found the Holy Build Box (https://github.com/phusion/holy-build-box). Is it the same?

bgruening:
See here: https://bioconda.github.io/contributor/build-system.html#stages-of-a-bioconda-build

> bioconda/bioconda-utils-build-env

CarlosHorro:
> See here: https://bioconda.github.io/contributor/build-system.html#stages-of-a-bioconda-build
> 
> > bioconda/bioconda-utils-build-env

I don't fully understand that doc, although I think it is the same than
https://bioconda.github.io/contributor/building-locally.html
isn't it?
If so, I was able to run

`bioconda-utils build --docker --mulled-test --git-range master --packages searchgui
`
both with current tests and adding one test for testing searchgui using xtandem... or at least I don't see any error in the output...

CarlosHorro:
I will leave on holidays for some weeks and later will be 10 days on quarantine (coming back from my country) working part-time, so I wanted to move forward this as much as possible before leaving...

bgruening:
@CarlosHorro can you popup on gitter tomorrow?

CarlosHorro:
> @CarlosHorro can you popup on gitter tomorrow?

@bgruening yep sure :-)

bgruening:
@CarlosHorro @bernt-matthias I fixed it the dirty way. The perfect is the enemy of the good. Of course, if we can fix it the proper way this would be appreciated, but we can proceed here now.

tools/peptideshaker/peptide_shaker.xml
```diff
@@ -316,99 +483,189 @@
...
+        </data>
+        <data format="fasta" name="output_proteins_sequences" from_work_dir="proteins_sequences.fasta" label="${tool.name} on ${on_string}: Proteins Sequences">
+            <filter>(exporting_options['zip_conditional']['zip_output_boolean'] is False or (exporting_options['zip_conditional']['zip_output_boolean'] is True and exporting_options['zip_conditional']['export_reports_when_zip'] is True) ) and exporting_options['followup_conditional']['followup_conditional_selector'] == 'yes' and exporting_options['followup_conditional']['followup_export_sequences'] != 'off' </filter>
+        </data>
+        <data format="txt" name="output_proteoforms" from_work_dir="proteins_proteoforms.txt" label="${tool.name} on ${on_string}: Proteins Proteoforms">
+            <filter>(exporting_options['zip_conditional']['zip_output_boolean'] is False or (exporting_options['zip_conditional']['zip_output_boolean'] is True and exporting_options['zip_conditional']['export_reports_when_zip'] is True) ) and exporting_options['followup_conditional']['followup_conditional_selector'] == 'yes' and exporting_options['followup_conditional']['followup_export_proteoforms'] != 'off' </filter>
```
bernt-matthias, line 516: I think 

```
(exporting_options['zip_conditional']['zip_output_boolean'] is False or (exporting_options['zip_conditional']['zip_output_boolean'] is True and exporting_options['zip_conditional']['export_reports_when_zip'] is True) )
```

can be simplified to 

```
(exporting_options['zip_conditional']['zip_output_boolean'] is False or exporting_options['zip_conditional']['export_reports_when_zip'] is True)
```

and maybe a bit nicer

```
(not exporting_options['zip_conditional']['zip_output_boolean'] or exporting_options['zip_conditional']['export_reports_when_zip'])
```

But its all equivalent and maybe you prefer the longer version for clarity?

CarlosHorro:
> @CarlosHorro @bernt-matthias I fixed it the dirty way. The perfect is the enemy of the good. Of course, if we can fix it the proper way this would be appreciated, but we can proceed here now.

Hi @bgruening ! I don't understand what you did for making it to work :-O  Did you finally remove the tests into containers?

bgruening:
I created a container with a more-recent base image: https://github.com/BioContainers/multi-package-containers/pull/1416/files

Normally, those containers are created after tool PRs are merged and normally they do have an older glibc installed. We circumvented this now. The downside is that those tools are not working on older (CentOS 6) machines, not with conda at least. But we have containers and with newer system it works.

I appreciate all your work you put in here, but I could not see your suffering anymore. We try to be perfect but at some point we need to stop and appreciate all your effort and I do think this solution covers 99% of all users probably. We can at any time fix it for the missing 1% if this is needed.

bgruening:
@CarlosHorro can you maybe use those version schemas? https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#tool-versions

We might have a bot, similar to the conda one, that bumps versions in Galaxy tools at some point.

CarlosHorro:
> I created a container with a more-recent base image: https://github.com/BioContainers/multi-package-containers/pull/1416/files
> 
> Normally, those containers are created after tool PRs are merged and normally they do have an older glibc installed. We circumvented this now. The downside is that those tools are not working on older (CentOS 6) machines, not with conda at least. But we have containers and with newer system it works.

Really interesting and glad about the tools are working in containers too. I think this is a safe movement as systems as old as CentOS 6 (almost 10 years old) are even officially obsolete (https://endoflife.software/operating-systems/linux/centos) and everyone using it is officially encouraged to leave it and change to a more modern one. They even have had their official repositories removed...

> 
> I appreciate all your work you put in here, but I could not see your suffering anymore. We try to be perfect but at some point we need to stop and appreciate all your effort and I do think this solution covers 99% of all users probably. We can at any time fix it for the missing 1% if this is needed.

Thank you very much for alleviating my suffering XD  I think it is very good in the long them to look for perfection but sometimes a balance is also necessary :-)

By the way, what do I do with https://github.com/bioconda/bioconda-recipes/pull/25876 ?  I guess this Conda dependency attempt is not longer necessary and it may be removed (the updated version of SG may be uploaded later, I rather prefer to focus on having at least one new version of the tools available).

bernt-matthias:
> By the way, what do I do with bioconda/bioconda-recipes#25876 ? I guess this Conda dependency attempt is not longer necessary and it may be removed (the updated version of SG may be uploaded later, I rather prefer to focus on having at least one new version of the tools available).

Lets keep suffering over there :) I think its worth on the long run.

tools/peptideshaker/peptide_shaker.xml
CarlosHorro: Happy new year to you all :-)

@bernt-matthias , of course you are right about 
`
(exporting_options['zip_conditional']['zip_output_boolean'] is False or (exporting_options['zip_conditional']['zip_output_boolean'] is True and exporting_options['zip_conditional']['export_reports_when_zip'] is True) )
`
may be simplified, although regarding to filters I prefer to keep them as descriptive as possible due to it is really easy to make a mistake and not notice it for a long time.
However, I think the second option you suggest is clear enough and it will reduce the length of some filters (currently excessive). This will also be good for clarity purposes, so I would go for it.

tools/peptideshaker/searchgui.xml
```diff
@@ -171,7 +171,7 @@
         <!-- Search Engine Selection -->
         <section name="search_engines_options" expanded="true" title="Search Engine Options">
             <param name="engines" type="select" display="checkboxes" multiple="True" label="DB-Search Engines">
-                <help>Comet and Tide shouldn't both be selected since they use a similar algoritm.</help>
+                <help>Comet and Tide shouldn't both be selected since they use a similar algoritm. OMSSA may not work into isolated environments like containers.</help>
                 <option value="X!Tandem" selected="True">X!Tandem</option>
                 <option value="MSGF" selected="True">MS-GF+</option>
                 <option value="OMSSA" selected="True">OMSSA</option>
```
bernt-matthias, line 177: Should we disable OMSSA by default?

tools/peptideshaker/searchgui.xml
```diff
@@ -317,7 +317,7 @@ Runs multiple search engines on any number of MGF peak lists using SearchGUI.
 
 Default:     X! Tandem, OMSSA and MS-GF+ are executed.
 
-Optional:     MyriMatch, MS-Amanda, Comet, Tide, DirecTag and Novor can be executed.
+Optional:     MyriMatch, MS-Amanda, OMSSA (it may not work into isolated environments like containers), Comet, Tide, DirecTag and Novor can be executed.
```
bernt-matthias, line 320: Listing OMSSA as default and optional is confusing.

tools/peptideshaker/searchgui.xml
CarlosHorro: Yes, you are probably right. I will disable it by default.

PR #286, opened 2018-05-24
Update customProDB dependency to 50a44bb8

chambm:
This update allows writing the reference FASTA without a BAM file, a new annotation source, PrepareAnnotationGCF, which should work for any GCF assembly with mRNA and the `_cds_from_genomic.fna.gz`, `_translated_from_cds.faa.gz`, and `_genomic.gff.gz` files
Fix easyRun() to indefinitely look for "#CHROM" instead of having a capped number of rows to skip; necessary for assemblies that have not been assembled into chromosomes


bgruening:
Tool version bump?

bgruening:
The tests are failing? :(

chambm:
There's a bunch of dependency resolution conflicts at the top that weren't there before (they couldn't be affected by just me updating the GitHub hash). Any advice for how to resolve those? I assume they're responsible for the later error:
```
--------------------------------------
| 7 - test1_sort_rpkm.fasta (HID - NAME) 
| Dataset State:
|  error
| Dataset Blurb:
|  error
| Dataset Info:
|  Error in loadNamespace(name) : there is no package called 'devtools'
|  Calls: withCallingHandlers ... tryCatch -> tryCatchList -> tryCatchOne -> <Anonymous>
|  Execution halted
|  Fatal error: Exit code 1 (Job Failed)
| Peek:
|  None
| Dataset Job Standard Output:
|  Error in loadNamespace(name) : there is no package called 'devtools'
|  Calls: withCallingHandlers ... tryCatch -> tryCatchList -> tryCatchOne -> <Anonymous>
|  Execution halted
| Dataset Job Standard Error:
|  Fatal error: Exit code 1 (Job Failed)
|
--------------------------------------
```

chambm:
Why aren't those packages needed?

bgruening:
In theory you just need to have the top level package. By including "all" packages you make it unnecessarily complicated for the solver to find a package list that is compatible.

chambm:
What top level package? I can't use the official bioconductor-customProDB conda package because we need the unofficial changes in my version. And it seems odd to make a conda package for an unofficial fork, right?

bgruening:
> What top level package? I can't use the official bioconductor-customProDB conda package because we need the unofficial changes in my version. And it seems odd to make a conda package for an unofficial fork, right?

Could we use https://github.com/galaxyproject/conda-iuc or give the package a different name?

chambm:
What is the least-effort way to get this in? It's definitely customProDB, it wouldn't make sense to change the name.

bgruening:
I do not know @chambm. From my point its not customProDB as the results from Galaxy can not be reproduced in R with the original packacge, or the one in bioconda. So either we fix the dependencies here (which seems to be hard) or we create an additional package.

bgruening:
ping @bedroesb

bgruening:
```
 Output variation table and variant protein sequence caused by SNVs... Error in result_create(conn@ptr, statement) : no such table: mtable
|  Calls: withCallingHandlers -> do.call -> <Anonymous> -> easyRun -> aaVariation
```

@chambm this does not seem to be a dependency error anymore.

chambm:
Hallelujah! :tada: We can cross this off the bucket list!

bgruening:
Mächtig gewaltig, Egon! Thanks.

PR #288, opened 2018-06-04
metaprosip fix and improvements

bernt-matthias:
- fixed number of columns in peptide centric tsv files differed (because the program just does not output empty columns at the end)
- better processing of the file names in the table (no allows for filtering)
- added option to output images to collection (before for png and svg output a html table was created and only for pdf a collection. but collection are better for postprocessing, eg filtering)


bernt-matthias:
travis problems are due to wrong urls in some openms tools.

bgruening:
Thanks Matthias!

PR #289, opened 2018-06-08
PathwayMatcher: new Galaxy tool.

CarlosHorro:
App written in Java to search for pathways related to a list of proteins in Reactome.

https://github.com/LuisFranciscoHS/PathwayMatcher


tools/pathwaymatcher/LICENSE.txt
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
bgruening: you do not need to include this.

tools/pathwaymatcher/pathwaymatcher.xml
```diff
@@ -0,0 +1,332 @@
+<tool id="pathwaymatcher" name="Pathway Matcher" version="1.7">
+    <description>
+        PathwayMatcher is a software tool writen in Java to search for pathways related to a list of proteins in Reactome.
```
bgruening: I would remove the information about Java ... and just tell the user what the tool is doing.

tools/pathwaymatcher/pathwaymatcher.xml
```diff
@@ -0,0 +1,332 @@
...
+
+          <param format="txt" name="input_vcf" type="data" label="Variant Call Format Specification" optional="true"
+            help="The input follows the Variant Call Format Specification[2] v4.3.
+            It also allows the possibility to specify only the first 4 columns in the data section of the file:
+            CHROM, POS, ID, REF.  "/>
+
+        </section>
+
+
+        <section name="genes" expanded="true" title="Genes">
```
bgruening: Does it make sense to have a section with just one option?

CarlosHorro: I think it separates visual and clearly different functional areas of PathwayMatcher app... and it looks far better :-)

bgruening: It makes the wrapper more complicated imho. If the style is sub-optimal we should fix this in Galaxy for all tools and keep the look and feel the same for all wrappers. No?

tools/pathwaymatcher/pathwaymatcher.xml
```diff
@@ -0,0 +1,332 @@
...
+
+        #####################
+        ## Pathway Matcher ##
+        #####################
+        (pathwaymatcher src.main.java.no.uib.pap.pathwaymatcher.PathwayMatcher
+
+            ## GENETIC VARIANTS
+
+            #if $genetic_variants.input_rsid:
+                -t rsid -i "${genetic_variants.input_rsid}"
```
bgruening: single quotes, for security reasons :)

bgruening: Please everywhere.

tools/pathwaymatcher/pathwaymatcher.xml
```diff
@@ -0,0 +1,332 @@
...
+                -gu
+            #end if
+
+            #if 'gp' in $output_graphs_list:
+                -gp
+            #end if
+
+        2>> $temp_stderr);
+
+        ## We create a zip file to contain graphs files.
```
bgruening: Why not creating a collection with all tsv files?

tools/pathwaymatcher/pathwaymatcher.xml
```diff
@@ -0,0 +1,332 @@
...
+        exit_code_for_galaxy=\$?;
+        cat $temp_stderr 2>&1;
+        (exit \$exit_code_for_galaxy)
+]]>
+    </command>
+    <inputs>
+
+        <section name="genetic_variants" expanded="true" title="Genetic variants">
+
+          <param format="txt" name="input_rsid" type="data" label="SNP rsId list" optional="true"
```
bgruening: As far as I can see all inputs are optional? This mean a tool can be submitted without any input?

CarlosHorro: Yes, in theory the tool can be submitted without any input, but of course it will give an error.
At least one input is necessary to get results, but I'm not sure about how to do that...

tools/pathwaymatcher/pathwaymatcher.xml
CarlosHorro: Well, I think the wrapper is still quite simple and (internally) very repetitive, so the additional complexity is almost zero and the visual hierarchy and functional separation (in the same way than the app itself) is an improvement that compensates it... 
But, of course, this may be quite subjective... about the functional separation, I think it's actually an improvement, but any designer may have different opinions about this visual style and aesthetics...

tools/pathwaymatcher/pathwaymatcher.xml
CarlosHorro: ok sorry :-)

tools/pathwaymatcher/pathwaymatcher.xml
CarlosHorro: I thought about it but, when you generate at least one graphic, you obtain 3 files; with the 3 graphs chosen, you get 9 different files (all of them related in groups of 3). You may be only interested in some specific ones (ie:internalEdges of genes, proteins and proteoforms) but still you would fill your history with them all (9), so in order to avoid this I think it's useful to have only one compressed file (which will also be much lighter to download with big datasets) with everything into it, also independently of the number of graphs you chose.

bgruening: I don't get this argument. If you just want to have one triple, then you need to download the entire zip, extract it and search for the files. In collections you just open them, browse and click on the file your are interested in. A collection is just one dataset in our history, so it does not pollute your history. Downloading should be possible as entire collection. I do agree its more complicated than a zip hack, but I think from a UX point of view its the correct things to do.

It your tool, I do not want to push you away, sorry if this is the impression :( - just want to provide hopefully useful tips.

tools/pathwaymatcher/pathwaymatcher.xml
bgruening: To make this a little bit less subjective we tried to develop standards for UX and tools styles here: http://galaxy-iuc-standards.readthedocs.io/en/latest/

tools/pathwaymatcher/pathwaymatcher.xml
bgruening: what about a repeat?

CarlosHorro: repeat? I don't understand sorry

bgruening: https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-inputs-repeat

You add a repeat and people select the type + file. And can add up to X repeat tuples.

tools/pathwaymatcher/pathwaymatcher.xml
CarlosHorro: Don't worry, I really appreciate your advice. It seems I was thinking on collections wrongly, as if they occupied many rows in the history. I will check and try to use them :-)

tools/pathwaymatcher/pathwaymatcher.xml
CarlosHorro: Thank you, I will check it.

CarlosHorro: Um... I think I'm not understanding the repeat clause. Adding the repeat clause you can still press the execute button without selecting any file. Even after pressing over "insert input files" you can still press the Execute button without adding files at all. And pressing again "insert input files" duplicates the entire interface without necessity, since with the original structure you can add extra files directly without any extra step...

bgruening: What about something like this:

```
<repeat name="samples" title="Samples" min="1">
  <param argument="-i" type="data" format="data" label="UniProt Accession list"
       help="File with a one Uniprot Accession [4] in each line."/>
  <param argument="-t" type="select" label="type">
       <option value="rsid">rsid</option>
       .....
  </param>
</repeat>
```

CarlosHorro: Yes, I did it. Although I specify min="1" you can still press on "execute" without adding any file... and the duplicated visualisation of input params when pressing on "insert input files" I think is not a good solution...

bgruening: You need to remove the optional=True from your data param.

CarlosHorro: Um... it still doesn`t do what I need. If I remove "optional=True" of all parameters, what happens is all of them have one value chosen by default without the possibility of removing it.
What it should do is to have them all empty by default (as only one or some of them will be used) and not allow the user to execute the tool without choosing at least one value for one param...

bgruening: Carlos can you show me the code? I think I misunderstood what you want?

CarlosHorro: Sure, I simplified it to this to check how "repeat" works:

'''
<inputs>

        <repeat name="input_files" title="Input files" min="1">

              <param format="txt" argument="-i" name="input_rsid" type="data" label="SNP rsId list"
                help="The file contains one rsid identifier as defined in dbSNP[1] on each row.
                The list must be ordered by chromosome and base pair (bp). The list must not have duplicates.
                All rsids must appear in the human assembly GRCh37.p13. "/>

              <param format="txt" argument="-t" name="input_gene" type="data" label="Genes"
                help="File with a one gene name in each line. Genes follow the HUGO gene nomenclature[3]."/>

        </repeat>

    </inputs>
'''

bgruening: So this is how I understand your use case you have X tuples of (-i and -t) parameters. You can add 1-X different tuples, right? But at least one tuple. So you start with one repeat. Yes you need to choose one file and one -i parameter (which I would do as select, why is this a file in your case?) The file is not allowed to be empty, because it is the first repeat. The -i is also not to be allowed to be empty because the tuple needs to be complete.
If you now choose to enter a second file you add a new repeat block and start all over again. Again if you have a second repeat blog, you need to enter -i & -t otherwise it is not complete, or remove the second repeat block. No?

CarlosHorro: Um... I think I understood now... I have to think about it

CarlosHorro: What do you think about this example?

'''
        <repeat name="input_types" title="Input Types" min="1">
            <conditional name="input_type">
                  <param name="input_type_selector" type="select" label="Input type"
                      help="">
                      <option value="rsid">Genetic variants - SNP rsId list</option>
                      <option value="chrbp">Genetic variants - Chromosomes and base pairs</option>
                      <option value="vcf">Genetic variants - Variant Call Format Specification</option>
                      <option value="peptide">Peptides - Simple list</option>
                     
                  </param>
                  <when value="rsid">
                      <param format="txt" name="input_rsid" type="data" label="SNP rsId list"
                        help="The file contains one rsid identifier as defined in dbSNP[1] on each row.
                        The list must be ordered by chromosome and base pair (bp). The list must not have duplicates.
                        All rsids must appear in the human assembly GRCh37.p13. "/>
                  </when>

                  <when value="chrbp">
                      <param format="txt" name="input_chrbp" type="data" label="Chromosomes and base pairs"
                        help="Genetic variants can also be represented using the chromosome and the base pair numbers.
                        The input should be sorted by chromosome number and then by base pair.  "/>
                  </when>

                  <when value="vcf">
                        <param format="txt" name="input_vcf" type="data" label="Variant Call Format Specification"
                          help="The input follows the Variant Call Format Specification[2] v4.3.
                          It also allows the possibility to specify only the first 4 columns in the data section of the file:
                          CHROM, POS, ID, REF.  "/>
                  </when>

                  <when value="peptide">
                        <param format="txt" name="input_peptide" type="data" label="Simple list"
                          help="File with a one peptide sequence in each line. You also need to choose a FASTA file to search them against it."/>

                        <param name="ptm_range" type="integer" value="0" label="PTM position range" optional="true"
                            help="Plus minus positions for the same PTM site."/>
                  </when>


            </conditional>
          
        </repeat>

'''

Main changes:
1) Select clause replaced by a conditional one. It allows to add necessary extra (and optional) options to some inputs (ie: peptides, for choosing PTM range)

2) It allows a more customised second field for the input files (with help and others)

bgruening:
@CarlosHorro can you replace all tabs by 4spaces.

tools/pathwaymatcher/pathwaymatcher.xml
```diff
@@ -0,0 +1,381 @@
...
+                  </when>
+
+                  <when value="chrbp">
+                      <param format="txt" name="input_chrbp" type="data" label="Chromosomes and base pairs"
+                          help="Genetic variants can also be represented using the chromosome and the base pair numbers.
+                          The input should be sorted by chromosome number and then by base pair.  "/>
+                  </when>
+
+                  <when value="vcf">
+                      <param format="txt" name="input_vcf" type="data" label="Variant Call Format Specification"
```
bgruening: format `vcf`

tools/pathwaymatcher/pathwaymatcher.xml
```diff
@@ -0,0 +1,381 @@
...
+                      <param format="fasta" name="input_peptide_database" type="data" label="Protein Database"
+                          help="Select FASTA database from history"/>
+
+                      <param name="ptm_range" type="integer" value="0" label="PTM position range" optional="true"
+                          help="Plus minus positions for the same PTM site."/>
+                  </when>
+
+                  <when value="modifiedpeptide">
+                      <param format="txt" name="input_modifiedpeptide" type="data" label="Peptide List with PTM types and sites"
+                         help="Each line of the file corresponds to a single peptide with post-translational modifications. You also need to choose a FASTA file to search them against it."/>
```
bgruening: the sentence with the fasta file is redundant, as the user can not submit without it

tools/pathwaymatcher/pathwaymatcher.xml
```diff
@@ -0,0 +1,381 @@
...
+                  </when>
+
+                  <when value="ensembl">
+                      <param format="txt" name="input_ensembl" type="data" label="Ensembl identifier list"
+                          help="File with a one Ensembl identifier [5] in each line."/>
+                  </when>
+
+                  <!-- Proteoforms  -->
+                  <when value="proteoforms">
+                      <param format="txt" name="input_proteoforms" type="data" label="Proteoforms"
```
bgruening: this is then a tabular?

tools/pathwaymatcher/pathwaymatcher.xml
```diff
@@ -0,0 +1,381 @@
...
+              <conditional name="input_type">
+                  <param name="input_type_selector" value="gene"/>
+                  <param name="input_gene" value="genes.txt" ftype="txt" />
+              </conditional>
+          </repeat>
+          <output name="search" file="genes_search.tsv" ftype="tsv" compare="sim_size" delta="3000" />
+      </test>
+
+      <!-- Test graphs from proteoforms -->
+      <test>
```
bgruening: You could add `<test expect_num_outputs="3">` to indicate the number of outputs and test for these.

tools/pathwaymatcher/pathwaymatcher.xml
```diff
@@ -0,0 +1,381 @@
...
+                      <param format="txt" name="input_gene" type="data" label="Genes"
+                          help="File with a one gene name in each line. Genes follow the HUGO gene nomenclature[3]."/>
+                  </when>
+
+                  <!-- Peptides  -->
+                  <when value="peptide">
+                      <param format="txt" name="input_peptide" type="data" label="Simple list"
+                          help="File with a one peptide sequence in each line. You also need to choose a FASTA file to search them against it."/>
+
+                      <param format="fasta" name="input_peptide_database" type="data" label="Protein Database"
```
bgruening: This could be a macro.

tools/pathwaymatcher/pathwaymatcher.xml
CarlosHorro: Um...it`s weird, I`ve done it (expect_num_outputs="3" in test 2) but it fails:

Exception: Incorrect number of outputs - expected 3, found 2.

expect_num_outputs="2" in test 1 works.
Maybe doesn`t collections count for this condition (expect_num_outputs)?

tools/pathwaymatcher/pathwaymatcher.xml
CarlosHorro: Well, expected format is a bit of a mix. This is the official documentation about it:
‘’’

Each line must follow any of these formats:

A single protein with no modifications
P00519
A protein with one PTM. The two fields are separated by a ','
P16220;00046:133
A protein and a set of PTMs separated by ';'. The PTMs can be ordered randomly
P62753;00000:235,00000:236,00000:240
In case the PTM type is not known, the modification id used is "00000". For example: "00000:245".


Example:

P00519;00046:245,00048:412
P31749;00047:473,00047:308
P11274;00187:177
P22681;00098:774
P22681
P16220;00046:133
P46109;01192:207
P27361;00047:202,00048:204
Q9UQC2;00000:452
Q15759;00048:182,00047:180
O15530;00048:241
P62753;00048:235,00049:236,00126:240
P12931;00048:419
P40763;00046:705,00046:727
P42229;00048:694

‘’’

CarlosHorro:
I think they are nice suggestions, I will try to include them :-)

tools/pathwaymatcher/pathwaymatcher.xml
bgruening: ok, lets leave it like this for the moment :)

tools/pathwaymatcher/pathwaymatcher.xml
bgruening: mh good question, then leave it out for the colleciton stuff.

bgruening:
Thanks @CarlosHorro! Great work!

bgruening:
Its now installed on usegalaxy.eu :)

PR #301, opened 2018-07-09
fix output of SequenceCoverageCalculator

bernt-matthias:
see https://github.com/OpenMS/OpenMS/issues/3605

+ add CDATA and tests


bernt-matthias:
linting seems to fail because of wrong URLs in other tools. Can't take care of this at the moment.

bgruening:
You rock! I have so many mails from you in my inbox! :)

PR #302, opened 2018-07-20
MSI tools updated for pixel annotation with tabular files

foellmelanie:


bgruening:
@foellmelanie can you double check if all files in this PR are also consumed in tests?

PR #287, opened 2018-05-30
updated cravatool and craviz with additional inputs and annotation

jraysajulga:
**Note:** _The Galaxy-P team is submitting a paper soon that requires this tool and plugin on the proteogenomics gateway. We were hoping to get this online ASAP._

PR #304, opened 2018-08-11
Cravat-P Tool and Visualization Plugin

jraysajulga:
Hello! I am opening a new pull request based off of previous CRAVAT pull requests (i.e., #287 and #284). I hope this is okay since it has been a few months. If so, we can close those PRs. 

We are looking to make this tool and visualization available on the Proteogenomics gateway for our paper's revision by next weekend (I was originally using a personal JetStream server to host these).

I have implemented Björn's suggestions from PR [#284](https://github.com/galaxyproteomics/tools-galaxyp/pull/284) for syntax, python3 compatibility, and functional tests (thank you). It would be great to have some eyes on this again. The visualization plugin is functional but still needs come clean-up. Thanks!


tools/cravatp_score_and_annotate/cravatp_submit.xml
```diff
@@ -0,0 +1,306 @@
+<tool id="cravatp_submit" name="CRAVAT-P Submit, Intersect, Check, and Retrieve" version="1.0.0">
+    <description>| Submits, checks for, and retrieves data for cancer annotation. Can support proteogenomic proBED inputs for intersection with genomic VCF input.</description>
+  <command detect_errors="aggressive" interpreter="python">
```
bgruening: interpreter python is deprecated, please use `$__tool_directory__/cravatp_submit.py` instead.

tools/cravatp_score_and_annotate/cravatp_submit.xml
```diff
@@ -0,0 +1,306 @@
+<tool id="cravatp_submit" name="CRAVAT-P Submit, Intersect, Check, and Retrieve" version="1.0.0">
+    <description>| Submits, checks for, and retrieves data for cancer annotation. Can support proteogenomic proBED inputs for intersection with genomic VCF input.</description>
```
bgruening: Imho the description is too long. The `Can support awesome filetype` is too much description imho. This should go into the help.

tools/cravatp_score_and_annotate/cravatp_submit.xml
```diff
@@ -0,0 +1,306 @@
...
+  <command detect_errors="aggressive" interpreter="python">
+  <![CDATA[
+#if $proteo.proteoInput == 'yes':
+    #if $analysis.type == 'CHASM' or $analysis.type == 'CHASM+VEST':
+        cravatp_submit.py $input $GRCh $Variant $Gene $Noncoding $Error $analysis.type --classifier $analysis.classifier --proBED $proBED --intersectOnly $proteo.intersectedVariants --vcfOutput $vcf_output
```
bgruening: please single-quote all data-parameter and text-parameter

tools/cravatp_score_and_annotate/cravatp_submit.xml
```diff
@@ -0,0 +1,306 @@
...
+    #if $analysis.type == 'CHASM' or $analysis.type == 'CHASM+VEST':
+        cravatp_submit.py $input $GRCh $Variant $Gene $Noncoding $Error $analysis.type --classifier $analysis.classifier 
+    #else: 
+        cravatp_submit.py $input $GRCh $Variant $Gene $Noncoding $Error $analysis.type
+    #end if
+#end if
+]]></command>
+  
+  <inputs>
+    <param format="tabular" name="input" type="data" label="Source file" help="Accepts transcriptomic or genomic inputs (e.g., tabular, VCF). Additional details can be found below."></param>
```
bgruening: VCF is missing in `format`?

tools/cravatp_score_and_annotate/cravatp_submit.xml
```diff
@@ -0,0 +1,306 @@
...
+    <conditional name="proteo">
+      <param format="tabular" name="proteoInput" type="select" display="radio" label="Intersect with proteogenomic input?" help="Source file (first input) must be in genomic input to enable intersection with this proteogenomic file.">
+        <option value="yes">Yes</option>
+        <option value="no" selected="true">No</option>
+      </param>
+      <when value="yes">
+        <param format="tabular" name="proBED" type="data" label="Peptides with Genomic Coordinates (ProBED Format)"></param>
```
bgruening: I think JJ added the proBED format already to Galaxy. So format should be probed https://github.com/galaxyproject/galaxy/blob/dev/config/datatypes_conf.xml.sample#L52

jraysajulga:
Thanks for the quick review. I will add these changes in separate PRs.

tools/cravatp_score_and_annotate/cravatp_submit.xml
jraysajulga: The proteogenomics gateway does not have proBED yet. I'm thinking I will add this specification later.

bgruening: Can we update or patch the proteomic gateway?

jraysajulga: Yes. That would be best. I'm wrote an e-mail to @chambm about this with the link that you provided.

chambm: I thought a main point of the proBAM and proBED formats was that they would be backward-compatible with existing BAM and BED processing tools. I know this is the case with proBAM at least. Tools that know nothing about proBAM are still able to process proBAM files properly. But I see there is a probam format in the new datatypes as well. I see proBED is a subclass of interval. I don't know much about proBED. Can BED-6 or BED-12 processing tools use proBED and get some benefit from the PSM information? If so it seems like having a separate type could be problematic.

PR #305, opened 2018-08-13
Cravat-P Tool

jraysajulga:
Branching off from PR #304, this PR contains only the tool, not the visualization plugin. Changes were made based on Björn's initial comments in PR #304 (e.g., added VCF and proBED file specifications and formatted "command" tag appropriately).


tools/cravatp_score_and_annotate/cravatp_submit.xml
```diff
@@ -0,0 +1,302 @@
...
+    #else: 
+        python '$__tool_directory__/cravatp_submit.py' '$input' '$GRCh' '$Variant' '$Gene' '$Noncoding' '$Error' '$analysis.type'
+    #end if
+#end if
+]]></command>
+  
+  <inputs>
+    <param format="vcf" name="input" type="data" label="Source file" help="Accepts transcriptomic or genomic inputs (e.g., tabular, VCF). Additional details can be found below."></param>
+    <conditional name="proteo">
+      <param format="tabular" name="proteoInput" type="select" display="radio" label="Intersect with proteogenomic input?" help="Source file (first input) must be in genomic input to enable intersection with this proteogenomic file.">
```
bgruening: Format can be removed here.
I would also remove the display="radio"

tools/cravatp_score_and_annotate/cravatp_submit.xml
```diff
@@ -0,0 +1,302 @@
...
+          <option value="Skin">Skin</option>
+          <option value="Stomach">Stomach</option>
+          <option value="Thyroid">Thyroid</option>
+          <option value="Uterus">Uterus</option>
+        </param>
+      </when>
+    </conditional>
+    <param format="tabular" name="GRCh" type="select" label="Genome Reference Consortium Human Build (GRCh)"
+        help="The default human reference genome used for annotation is GRCh38, released on December 24th, 2013 from the Genome Reference Consortium.">
+      <option value="GRCh38">GRCh38/hg38</option>
```
bgruening: just for you information, we could in theory retrieve this information from the dbkey of a dataset, but we will leave it as it is for now.

tools/cravatp_score_and_annotate/cravatp_submit.xml
```diff
@@ -0,0 +1,302 @@
...
+      <option value="GRCh37">GRCh37/hg19</option>
+    </param>
+  </inputs>
+  <outputs>
+    <collection name="results" type="list" label="CRAVAT Results on ${on_string}">
+      <data format="tabular" label="CRAVAT: Gene Level Annotation Report on ${on_string}" name="Gene" />
+      <data format="tabular" label="CRAVAT: Variant Report on ${on_string}" name="Variant" />
+      <data format="tabular" label="CRAVAT: Non-coding Variant Report on ${on_string}" name="Noncoding" />
+      <data format="tabular" label="CRAVAT: Errors on ${on_string}" name="Error" />
```
bgruening: I would write all `name`s with small letters. They become variables and usually we stick to small letters.

jraysajulga:
Great. Thanks for the review. One question for toolshed upload: on the galaxyp toolshed account, should I update the cravatool repository or create a new one with the new cravatp_score_and_annotate name? Also, I'm thinking of keeping the proBED input as a BED input for now in the xml file.

tools/cravatp_score_and_annotate/cravatp_submit.xml
jraysajulga: I used uppercase here since the filenames within the collection on Galaxy match with these variable names. Lowercase looks fine too.
image

tools/cravatp_score_and_annotate/cravatp_submit.xml
jraysajulga: That's a good point. I'll comment this in as a to-do for the future.

Issue #293, opened 2018-06-15
searchgui: FastaCLI fails with getting user preferences from home dir


bernt-matthias:
SearchGUI fails with: 

```
An error occurred while saving /home/petrusch/.compomics/userpreferences.cup (see below).
java.io.FileNotFoundException: /home/petrusch/.compomics/userpreferences.cup (No such file or directory)
	at java.io.FileOutputStream.open0(Native Method)
	at java.io.FileOutputStream.open(FileOutputStream.java:270)
	at java.io.FileOutputStream.<init>(FileOutputStream.java:213)
	at java.io.FileOutputStream.<init>(FileOutputStream.java:162)
	at com.compomics.util.io.SerializationUtils.writeObject(SerializationUtils.java:23)
	at com.compomics.util.preferences.UtilitiesUserPreferences.saveUserPreferences(UtilitiesUserPreferences.java:626)
	at com.compomics.util.preferences.UtilitiesUserPreferences.loadUserPreferences(UtilitiesUserPreferences.java:645)
	at eu.isas.searchgui.cmd.FastaCLI.generateTargetDecoyDatabase(FastaCLI.java:169)
	at eu.isas.searchgui.cmd.FastaCLI.call(FastaCLI.java:119)
	at eu.isas.searchgui.cmd.FastaCLI.<init>(FastaCLI.java:59)
	at eu.isas.searchgui.cmd.FastaCLI.main(FastaCLI.java:218)
```

The file actually does not exist. However, in Galaxy user preferences from the user HOME should not be used anyway. This is particularly bad for Galaxies running jobs as real user. 

To solve this we need to make fastaCLI work without the preferences or to autogenerate  and use preferences in the job working directory.

bgruening:
@bernt-matthias wouldn't it be enough to set $HOME inside the Galaxy wrapper to /tmp or similar?

bernt-matthias:
Hmm. Maybe I should rather consider this as a configuration problem and set `shared_home_dir` in galaxy's config.  

But I'm not sure if the will have any effect, since @jmchilton's PR only creates `JOBDIR/home` and some galaxy variables, but leaves `$HOME` seemingly untouched. But I will try it these days.

bernt-matthias:
Just have seen that $HOME is already set. Trying to fix now based on comments in https://github.com/compomics/searchgui/issues/107 but ran into a now problem: https://github.com/compomics/searchgui/issues/180

[...]

PR #312, opened 2018-09-10
searchgui home fix and enzyme list update

bernt-matthias:
- fixes https://github.com/galaxyproteomics/tools-galaxyp/issues/293
- removes enzymes that are not supported anymore. 
- adds OOM regex checks to stdio
- removed unnecessary default enzyme (since there was a typo in the command block the option never worked anyway, so we can savely remove it)

Ping @CarlosHorro, @hbarsnes. Thanks for your help .. could you have a look?

The second part of the PR solves the problem described below: 

I noticed that many of the parameters to -enzyme parameter of IdentificationParametersCLI do not work at the moment, e.g. "Trypsin, no P rule". I guess the options are wrong in some cases, "Trypsin (no P rule)" seems to work. 

The command line help mentions:
- EnzymesCLI, but I could not find it. 
- the possibility give a comma separated list of enzymes which does not work if there are commas in the enzyme names
- also the example "Trypsin, Glu-C" does not work .. actually gives an exception

A fixed the one enzyme that our users interested in at the moment.


bernt-matthias:
Thanks a lot for your comments. I updated the list of enzymes.

bernt-matthias:
again the fatal_oom problem

bernt-matthias:
wondering why the wrong output name did work... I tested and noticed that any name con be used. Is this because it's a single output?

bgruening:
@bernt-matthias sorry I did not get your last comment? Can you explain this more?

bgruening:
Thanks all!

bernt-matthias:
> sorry I did not get your last comment? Can you explain this more?

Sure I was wondering why the version before 
https://github.com/galaxyproteomics/tools-galaxyp/pull/312/commits/e238bae9ddea533533a2ccdf7703bdfc7124cf0a

could be tested successfully with planemo. Apparently the name of the output did not matter at all.

bgruening:
Yes, please. Planemo can and should be a lot smarter about variable-names and their existence.

bernt-matthias:
done

PR #314, opened 2018-09-18
openms: added out of memory tests

bernt-matthias:
I'm not sure if all tools involve java. Then this should be fine. Alternatively the oom tests could be applied only for java based tools. 

I'm doing this in particular for MSGFPlusAdapter

### DIFF
--- a/tools/openms/macros.xml
+++ b/tools/openms/macros.xml
@@ -16,7 +16,7 @@
       <regex match="java.lang.OutOfMemoryError" level="fatal_oom" description="Java memory Exception"/>
       <regex match="Could not allocate metaspace" level="fatal_oom" description="Java memory Exception"/>
       <regex match="Cannot create VM thread" level="fatal_oom" description="Java memory Exception"/>
-      <regex match="Exception:" level="fatal" description="Java Exception"/>
+      <regex match="qUncompress: could not allocate enough memory to uncompress data" level="fatal_oom" description="Java memory Exception"/>
     </stdio>
   </xml>
   <xml name="references">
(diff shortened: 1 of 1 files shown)


tools/openms/macros.xml
```diff
@@ -13,7 +13,10 @@
     <stdio>
       <exit_code range="1:"/>
       <exit_code range=":-1"/>
-      <regex match="Exception:"/>
+      <regex match="java.lang.OutOfMemoryError" level="fatal_oom" description="Java memory Exception"/>
+      <regex match="Could not allocate metaspace" level="fatal_oom" description="Java memory Exception"/>
+      <regex match="Cannot create VM thread" level="fatal_oom" description="Java memory Exception"/>
+      <regex match="Exception:" level="fatal" description="Java Exception"/>
```
bgruening: The last one seems not very java specific to me. All others I'm fine to merge. Can you just make the description of the last one less Java centric :)

bernt-matthias: the exact output of the adapter (which is as for as I know directly from java) is.

```
Error occurred during initialization of VM
Cannot create VM thread. Out of system resources.
```

So maybe we add `Out of system resources.`?

bgruening: I think its all ok, besides the `Exception:` regex. Exception can be written from everywhere ... and does not need to be a Java issue.

bernt-matthias: OK. I removed it. Even though it was in the original I guess it wont be a problem, since the exit codes should kick in before.

bernt-matthias: And added a new one I just observed in PeakPickerHighRes.

bgruening:
Thanks @bernt-matthias!

PR #317, opened 2018-09-28
try to fix featurefindermetabo

bgruening:


eschen42:
When I tried this again (on UseGalaxy.eu after Björn made this change) it worked fine for me. 
And, the change it self makes sense, I think this change is all good.

bgruening:
Thanks for reporting back!

PR #318, opened 2018-09-29
PeptideShaker: updated reference to version 1.16.29 + 3 reports

CarlosHorro:
Added 3 default reports: Default PSM Report with non-validated matches, Default Peptide Report with non-validated matches and Default Protein Report with non-validated matches.

ReportCLI: reports` ids have changed:
0: Certificate of Analysis,
1: Default Hierarchical Report,
2: Default PSM Phosphorylation Report,
3: Default PSM Report,
4: Default PSM Report with non-validated matches,
5: Default Peptide Phosphorylation Report,
6: Default Peptide Report,
7: Default Peptide Report with non-validated matches,
8: Default Protein Phosphorylation Report,
9: Default Protein Report,
10: Default Protein Report with non-validated matches,
11: Extended PSM Report,
12-n: Your own custom reports.

PR #319, opened 2018-09-30
peptideshaker update to 1.16.31

bernt-matthias:
Update to 1.16.30 (will be released soon https://github.com/compomics/peptide-shaker/issues/331).

- fixes peptide shaker writing to home dir 
- simplifies wrapper command (ReportCLI and MzidCLI calls are removed)
- integrates new reports that are generated by the new version


bgruening:
@bernt-matthias have you seen https://github.com/galaxyproteomics/tools-galaxyp/pull/318

bgruening:
ping @CarlosHorro as well

CarlosHorro:
Hi,

Indeed this version seems better than my PR, so I will cancel it and I will test this one.
About the tests, their content is supposed to be deterministic although I'm not totally sure if they will have 100% exactly the same content in every execution.

bernt-matthias:
@CarlosHorro cool. I will try to implement improved test. 

Note that testing is only useful when the 1.16.30 version is in bioconda. Wondering why the planemo tests are successful anyway... ping @bgruening

CarlosHorro:
Sure, as soon as we do some more testing of 1.16.30 we will deploy a new PeptideShaker version and then I will create the new PR for a new conda package.

bgruening:
If the version is not found planemo will fallback to the requirement without version number I think,

CarlosHorro:
Hi,

I'm having a problem testing this galaxy wrapper. When running, I don't find any generated report into \$cwd/output_reports... does it happen to you?

bernt-matthias:
I did not test the updated wrapper with 1.16.30 .. 

Problem is probably that with 1.16.29 (and before) the command line option for specifying the output directory for the reports does not exist (as far as I know). 

So I guess the reports are just written to $cwd? 

Would also be OK (actually this was the previous behaviour: PeptideShakerCLI wrote to $cdw and ReportCLI wrote to $cwd...).

CarlosHorro:
I think I found the problem. The output directory directive was there in previous versions :
(peptide-shaker eu.isas.peptideshaker.cmd.ReportCLI
...
-out_reports \$cwd/output_reports
...

but in 1.16.30 there is an extra change in the behaviour: when choosing -zip option, reports are not longer available out of the zip file. And I've found out that, independently of if the zip option is checked or not, we are always using that parameter into the command line:

(peptide-shaker eu.isas.peptideshaker.cmd.PeptideShakerCLI
...
            -zip \$cwd/peptideshaker_output.zip
...

So it seems that param should be used only when the zip box is also checked. I will confirm this point with Harald, but removing the -zip parameter the reports are properly generated.

CarlosHorro:
After talking with Harald, he confirmed that the desired behaviour should be:

- When user checks "zip" option, all files, reports included, should be contained into the returned zip file. This implies there will only be one return to the user: the zip file itself.
- When users does not check "zip" option, all files should be returned separately as usual, and -zip param must not be internally used.

By the way, I found out another error: MzidentML generation is failing because -output_file param from peptide-shaker eu.isas.peptideshaker.cmd.MzidCLI call has been removed.

bernt-matthias:
Note, my last commit is untested. 

In the `<outputs>` you find a commented part that should show the idea how to simplify 
the outputs section and could be rid of most of the `find`+`mv` things in the beginning.

In would be nice if the report names would be more easy parsable by a regular expression: eg `SAMPLENAME_SEPARATOR_REPORTTYPE.txt` then `pattern=".*\.(?P&lt;designation&gt;.+)\.txt"` would work and we get the data set names easily. Instead of separate data sets also a collection could be nice?

> When user checks "zip" option, all files, reports included, should be contained into the returned zip file. This implies there will only be one return to the user: the zip file itself. When users does not check "zip" option, all files should be returned separately as usual, and -zip param must not be internally used.

I suggested a solution that is based on an additional conditional. 

> By the way, I found out another error: MzidentML generation is failing because -output_file param from peptide-shaker eu.isas.peptideshaker.cmd.MzidCLI call has been removed.

has been added again.

bernt-matthias:
@CarlosHorro thanks for https://github.com/bioconda/bioconda-recipes/pull/11215 

I just bumped the version. Will try to fix potential travis errors these days.

tools/peptideshaker/peptide_shaker.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="peptide_shaker" name="Peptide Shaker" version="1.16.30">
+<tool id="peptide_shaker" name="Peptide Shaker" version="@PEPTIDESHAKER_VERSION@">
```
bgruening: should the peptide-shaker version also be used in the requirements?

bernt-matthias: oh. sure. :)

bernt-matthias:
There you go :) Please check if everybody is happy

bgruening:
`.. ERROR: Found output tag with unknown name [output], valid names [['searchgui_results']]`

bernt-matthias:
@CarlosHorro Any idea whats going on with the peptide shaker tests? Seems that it would like to have a X11 display...

bernt-matthias:
Added `-Djava.awt.headless=true` to both peptide shaker calls. Lets see if this helps. If so, I guess this should be don in SearchGUI as well?

bernt-matthias:
One thing that I see in the peptideshaker output is: 

`ZIP file contained more than one file, only the first file was added to Galaxy.`

Should we worry about this?

bgruening:
> Should we worry about this?

Where do you see this? This should only be seen if you upload the zip file and Galaxy tries to unpack it. Adding an explicit `zip` datatype as `ftype=` should help here imho. Then Galaxy do not unpack it.

bernt-matthias:
@bgruening : During the test. The ftype is already `searchgui_archive` which derives from `compressed_archive`.

bernt-matthias:
Hi @CarlosHorro can you run `planemo t` on your machine? At which point you are actually struggling. 

Warning ahaed: tests are running locally for me. Only the travis checks are failing since they seem to run head less. Definitely need to solve this since also many clusters run headless.

bernt-matthias:
For the record: Using `-gui 0` and  `-Djava.awt.headless=true` for the `PeptideShakerCLI` call gives: 

```
|  java.awt.HeadlessException:
|  No X11 DISPLAY variable was set, but this program performed an operation which requires it.
|  at java.awt.GraphicsEnvironment.checkHeadless(GraphicsEnvironment.java:204)
|  at java.awt.Window.<init>(Window.java:536)
|  at java.awt.Frame.<init>(Frame.java:420)
|  at java.awt.Frame.<init>(Frame.java:385)
|  at javax.swing.JFrame.<init>(JFrame.java:189)
|  at eu.isas.peptideshaker.gui.PeptideShakerGUI.<init>(PeptideShakerGUI.java:498)
|  at eu.isas.peptideshaker.cmd.PathSettingsCLI.setPathSettings(PathSettingsCLI.java:70)
|  at eu.isas.peptideshaker.cmd.PathSettingsCLI.extractAndUpdatePathOptions(PathSettingsCLI.java:274)
|  at eu.isas.peptideshaker.cmd.PeptideShakerCLI.main(PeptideShakerCLI.java:892)
```

CarlosHorro:
Hi again,

We are using on many servers Galaxy 17.09 (and it is not possible to update them now), but I think it does not support things like use_shared_home or the fatal_oom error level so it gives us errors. 
Could you remove them, please?

bernt-matthias:
Hi @CarlosHorro I disabled them temporarily. But the tools should run anyway (I just tested `planemo t --galaxy_branch release_17.09`), do you really get errors related to use_shared_home or the fatal_oom? I was hoping that they are just silently ignored or raise warnings on older versions.

CarlosHorro:
Thank you @bernt-matthias , as those keywords were not found in the schema, the entire test suite failed before starting...
I will try to test it again.

CarlosHorro:
Hi again,

After testing the wrapper, there are some little things to retouch, and one doubt: when choosing zip file, the check-boxes to choose which files we want to create disappears. This selector should be kept as, although we may choose to create a zip file, we should still be able to choose which files we want to create into the zip file. And, by the way, when choosing the zip file creation, I dont find the reports into the zip file...
In order to "retouch" these things, could you give me permissions to push to your repo? I think it would be a lot faster...

Apart from this, I think there is a bugfix to be made in the application itself in relation to the graphics request when executing the tool (although it works without the Djava.awt.headless=true). I will talk with Harald about this.

bernt-matthias:
Hi @CarlosHorro I added you as a collaborator. Edit as much as you want :) 

I also guess we need a bugfix for the application.

CarlosHorro:
I am refactoring some stuff we wanted to do for a long time, and after that it will be ready to point to the new app version.

CarlosHorro:
Hi,

Everything is ready for the new app version (I will change the version number in the macros file when ready). Just one question: was there a reason to remove the check of the content in all tests? I was planning to add it again...

Greetings,
Carlos

bernt-matthias:
The current tests check for the complete file content which is a bit more stringent than the old tests.

bernt-matthias:
After updating the test data tests should work now. Remaining: 

- I would really like to get the test for the out of memory condition back .. @bgruening is there a way to have this such that the new version also runs on older galaxy versions? Or do I need a local patch...?

bgruening:
I guess you need to patch your local Galaxy instance then.

bgruening:
18.09 will be out soon and you probably want to update, so many cool features :)

bernt-matthias:
Sure. My instance is up to date and will be upgraded as soon as 18.09 is out.  @CarlosHorro seems to maintain some older instances... 

Anyway my suggestion would be to include the oom checks. Since the changes are mostly of (important) technical nature there is no need to install this peptideshaker update .. if the bug did not affect you so far. This gives everybody some time to upgrade Galaxy.

bernt-matthias:
Furthermore, is the local reference data (https://github.com/galaxyproteomics/tools-galaxyp/issues/269) still an issue? Then this would be a good opportunity to fix it .. (even if my users would be more than happy to have a working searchgui/peptideshaker in our Galaxy).

CarlosHorro:
> Furthermore, is the local reference data (#269) still an issue? Then this would be a good opportunity to fix it .. (even if my users would be more than happy to have a working searchgui/peptideshaker in our Galaxy).

I do not know about that, as far as I know there were no related changes.

About the oom checks, as I cancelled my own PR with the support for the new reports in order to use this one, here we are not only talking about technical improvements but also about new features highly required by us. We would really need to keep anything not supported by galaxy 17.09 out of this release please... we are finishing a publication (it is because of it that we cannot update now our main galaxy server) and we need to install the tool in our server without extra problems, apart from our paper requires anyone to be able to install the tool easily, and it would be far better if we support not just the latest galaxy versions...

Greetings,
Carlos

bgruening:
@bernt-matthias is that fine with you? That you patch your local version and the next version after this we require a new Galaxy release?

bernt-matthias:
Yep, fine with me, I reverted my last commit. 

I will start a new PR immediately after this is merged :) Theoretically we could also push both versions to the toolshed. But I'm also fine with patching my local installation of the tool.

bgruening:
Thanks all!

PR #324, opened 2018-11-12
Contig2fastg Update

jraysajulga:
Hello. I have updated the descriptions for contig2fastg. Hoping to get this on Galaxy.eu.


tools/megahit_contig2fastg/megahit_contig2fastg.xml
```diff
@@ -14,11 +14,11 @@
         '${contigs}' > '${fastg}'
     ]]></command>
     <inputs>
-        <param name="contigs" type="data" format="fastq,fastqsanger,fasta,fastq.gz,fastqsanger.gz" multiple="true" label="Single-end file(s)" help="FASTQ/FASTA/FASTQ.GZ files accepted" />
-        <param name="kmer" type="integer" value="99" label="K-mer length" />
+        <param name="contigs" type="data" format="fasta" label="Contig file" help="Ex: k99.contigs.fa" />
```
bgruening: This example does not make sense in the Galaxy context as the user never sees a filename.

tools/megahit_contig2fastg/megahit_contig2fastg.xml
```diff
@@ -14,11 +14,11 @@
...
+        <param name="kmer" type="integer" value="99" label="K-mer length" help="Usually indicated within the filename (when derived from MEGAHIT)"/>
```
bgruening: I'm not sure this help is useful either as the filename is not seen by the user.

tools/megahit_contig2fastg/megahit_contig2fastg.xml
```diff
@@ -1,8 +1,8 @@
-<?xml version='1.0' encoding='utf-8'?>
-<tool id="megahit_contig2fastg" name="MEGAHIT Contig2fastg" version="@VERSION@">
-    <description>for converting MEGAHIT's contigs to fastg</description>
+<?xml version='0.9' encoding='utf-8'?>
+<tool id="megahit_contig2fastg" name="megahit contig2fastg" version="@VERSION@">
+    <description>for converting MEGAHIT's contigs (.fa) to assembly graphs (.fastg)</description>
```
bgruening: what about `for converting MEGAHIT's contigs to assembly graphs`?

tools/megahit_contig2fastg/megahit_contig2fastg.xml
```diff
@@ -1,8 +1,8 @@
-<?xml version='1.0' encoding='utf-8'?>
-<tool id="megahit_contig2fastg" name="MEGAHIT Contig2fastg" version="@VERSION@">
-    <description>for converting MEGAHIT's contigs to fastg</description>
+<?xml version='0.9' encoding='utf-8'?>
```
bgruening: this XML should not be changed, or can be completly removed.

tools/megahit_contig2fastg/megahit_contig2fastg.xml
jraysajulga: Ah. That makes sense. Thanks. I removed those.

tools/megahit_contig2fastg/megahit_contig2fastg.xml
jraysajulga: Okay. Removed this as well.

tools/megahit_contig2fastg/megahit_contig2fastg.xml
jraysajulga: I like that.

tools/megahit_contig2fastg/megahit_contig2fastg.xml
jraysajulga: I reverted it back to the original version.

jraysajulga:
Once this looks satisfactory, can you add this tool to your https://usegalaxy.eu server?

bgruening:
@jraysajulga this now goes to tools-iuc? So can we close this one?

jraysajulga:
@bgruening Now that it's been approved, this should go on tools-iuc. Shall we remove this tool from the tools-galaxyp repo?

PR #325, opened 2018-11-16
remove megahit

bgruening:


bgruening:
ping @jraysajulga

jraysajulga:
Bye bye MEGAHIT. Thanks!

bgruening:
I have also deprecated the galaxyp version on the TS. @jraysajulga we need to be more careful where we put those tools. As background info, a Galaxy tool has an ID, consisting out of a triplet - toolshed + owner + tool-id.

As soon as we change one of those three, Galaxy will not recognize it as one and the same tool and hence you will see two tools in the toolbar and in workflows the tool can not be updated from one version to the other.

jraysajulga:
Thanks and agreed... I've seen a lot of "duplicate" tools as a result of those differences.

PR #326, opened 2018-11-20
Peptideshaker updated to v1.16.36 & SearchGUI updated to v3.3.10

CarlosHorro:


CarlosHorro:
Very weird error, I will try to push an empty commit to see if the process works now.

`Fetching package metadata .....
CondaHTTPError: HTTP 502 BAD GATEWAY for url <https://conda.anaconda.org/conda-forge/linux-64/repodata.json>
Elapsed: 00:19.508897
CF-RAY: 47c9f5b8bf889f05-ORD
An remote server error occurred when trying to retrieve this URL.
A 500-type error (e.g. 500, 501, 502, 503, etc.) indicates the server failed to
fulfill a valid request.  The problem may be spurious, and will resolve itself if you
try your request again.  If the problem persists, consider notifying the maintainer
of the remote server.
conda_init failed - failed to install to '/home/travis/conda/bin/conda'
The command "planemo conda_init" failed and exited with 1 during .`

CarlosHorro:
I think I've missed to add the PTMs to the searchgui_mods.loc.sample (and I think there are others missed ones), so it will be better to wait until I add them to merge this PR.

CarlosHorro:
I think this PR is ready to be merged :-)

bernt-matthias:
Regarding the `searchgui_mods.loc.sample` file: wouldn't it be better to have this implemented as a macro?

CarlosHorro:
Hi,

As far as I know, loading options to a dropdown list from a file may be done using a .loc file or a data table (in xml). I do not know if it is possible to do that from a macro file, and I do not know if that would be an improvement over current .loc file...

bernt-matthias:
For an example see the macro here: 

https://github.com/galaxyproject/tools-iuc/blob/4364e7d61b72e12afb7ef294955b1c1e57f757b0/tools/stacks/macros.xml#L48

and its use: 
https://github.com/galaxyproject/tools-iuc/blob/4364e7d61b72e12afb7ef294955b1c1e57f757b0/tools/stacks/stacks_procrad.xml#L115

The problem with the .loc file is that it is used for all versions, i.e., the selection possibilities are used in all installed versions of the tool. I think this is undesired. 

If admins want to add custom additions (is this actually a use case?) they can still do this in the macro file.

CarlosHorro:
Hi Matthias,

Sorry, I think I don't understand your point here:
"The problem with the .loc file is that it is used for all versions, i.e., the selection possibilities are used in all installed versions of the tool. I think this is undesired."

Why using the loc file in all versions is a bad thing? every version of the galaxy tool has its own PR with its own changes pointing (it it is the case) to a specific app version, like this one which may include into the loc file any desired modification we need for the specific release...
Apart from that, I also think that the loc file is managed in a very easier way than the macros file (it just have the content in plain text format, without any extra tag). 

About adding custom additions, in current backend of the tool it's not possible to do that (using the command line). It will be possible when new backend is available.

bernt-matthias:
The loc file is installed globally in `GALAXY_ROOT/tool-data/searchgui_mods.loc` and all installed searchgui versions use this file. 

Hence also mods added in newer version are available in older versions which probably don't support them. 

I see your point with the additional tags, but these could be added automatically using some bash magic, eg.
somthing like `sed -e "s/(.*)/<option>\1<option>/g"` searchgui_mods.loc will all the xml tags necessary.

CarlosHorro:
Hi Matthias,

Great point, I did not know about the location of that file. After some research, it stills seem to me very weird that the same searchgui_mods.loc file can be found in different places: I find it into the location you mention, but it may also be found into every specific tool folder (ie: /galaxy/tool-data/toolshed.g2.bx.psu.edu/repos/galaxyp/peptideshaker/5814883aa217 , ie: /galaxy/tool-data/toolshed.g2.bx.psu.edu/repos/galaxyp/peptideshaker/baeced706dbc , etc).
What I do not understand is why these last versioned files are not being used, and the shared file is in fact being used (I have checked it) ? any clue @bgruening ?

About the format, I know those tags could be added but still think plain format is clearly more manageable. 
But, of course, your point about sharing the same file by different tool versions is far more important. I would like to wait for some feedback by Bjorn (if possible), and, if it is the best option (it seems to), to ask for approving this PR and open a new PR with the specific issue related to this file. I think we must manage it separately without causing extra delay to this update of the tool.

CarlosHorro:
At this point I think this PR can be merged (we need the updates working after so many days) and I can open later another PR just with the searchgui_mods.loc.sample topic.

bgruening:
@CarlosHorro sorry did not see that discussion.

> If admins want to add custom additions (is this actually a use case?) they can still do this in the macro file.

I see that a use-case and this was probably the intention to do is this way. Hacking a macro is not recommended afaik.

However, if the `mods` are version dependent that might be a problem and we might want to add a third column. There is something to enhance here in general, as `from_data` is not `from_data_table`. And afaik `from_data` should not be not shared at all. See here: https://docs.galaxyproject.org/en/latest/dev/schema.html#from-dataset

@blankenberg can you have a look at this and correct me please.

bernt-matthias:
I thnk `from_file` (@bgruening I guess you did not mean `from_data`) takes the loc file in the `tool-data` dir (here: https://github.com/galaxyproject/galaxy/blob/e98f63caf2d2adfd7857ff9fcba6df0b5923b668/lib/galaxy/tools/parameters/basic.py#L1349) .. contrary `from_data_table` takes it from the merged data tables of all installed versions. 

Both are wrong in my opinion. Since in both cases old versions of the tool will have entries available that have been added in later versions. 

I would suggest to code this in xml (eg macro) .. you may still maintain the plain text file and use a python script to create the xml.

bgruening:
@bernt-matthias I need to test this locally. And we need to improve documentation about this.
Do you agree that we can merge this PR as it improves the tool and does not make things worse?

bernt-matthias:
agreed.

bgruening:
Merci! And happy new year!

PR #332, opened 2018-12-18
QuanTP updated to support interactive graphs via visualization plugin

jraysajulga:
Updated @pravs3683's QuanTP tool to support interactive visualization through a visualization plugin (Interactive Viewer). I am planning to use this plugin to make [metaquantome](https://github.com/galaxyproteomics/metaquantome) interactive as well.


tools/quantp/quantp.xml
```diff
@@ -1,11 +1,13 @@
-<tool id="quantp" name="QuanTP" version="1.0.0">
+<tool id="quantp" name="QuanTP" version="1.1.0">
```
bgruening: The only thing that I find strange is that with so many changes in the R script, the test does not need to be changed. But I guess this is due to the limited scope of the current test. Do you think the current test is enough to cover the functionality of this tool?

visualizations/interactiveviewer/config/interactiveviewer.xml
```diff
@@ -0,0 +1,20 @@
...
+<!DOCTYPE visualization SYSTEM "../../visualization.dtd">
+<visualization name="Interactive Viewer">
+    <data_sources>
+        <data_source>
+            <model_class>HistoryDatasetAssociation</model_class>
+            <test test_attr="creating_job.tool_id">
+                quantp
+            </test>
+            <test test_attr="creating_job.tool_id">
+                toolshed.g2.bx.psu.edu/repos/galaxyp/quantp/quantp/1.1.0
```
bgruening: Wow, cool didn't know this is possible :)

jraysajulga: It's super useful! I learned this trick [here](url
https://github.com/galaxyproject/galaxy-hub/blob/master/src/visualizations-registry/cookbook/index.md#i-want-my-visualization-to-work-only-on-datasets-produced-by-a-specific-tool).

tools/quantp/quantp.xml
jraysajulga: No. I added another criterion to cover the extra functionality [here](https://github.com/galaxyproteomics/tools-galaxyp/pull/332/commits/060bf1c2422453f07a80346ce6e89341f031c715).

bgruening:
Cool thanks!

jraysajulga:
Great. Thanks for the review.

PR #333, opened 2019-01-09
added support for logfold cases for QuanTP

jraysajulga:
Hi! Just adding interactive support for logfold input types for a version 1.1.1 of QuanTP


visualizations/interactiveviewer/config/interactiveviewer.xml
```diff
@@ -10,6 +10,9 @@
             <test test_attr="creating_job.tool_id">
                 toolshed.g2.bx.psu.edu/repos/galaxyp/quantp/quantp/1.1.0
             </test>
+            <test test_attr="creating_job.tool_id">
+                toolshed.g2.bx.psu.edu/repos/galaxyp/quantp/quantp/1.1.1
```
bgruening: can you maybe remove the version? And with this enable it for all versions?

jraysajulga: Hmm... The issue is that the plugin does not serve any function with version 1.0, but it also doesn't throw any issues, so this might work. I can remove the version if that makes things cleaner.

bgruening: My concern here is that we forget at some point to update the version :)

bgruening:
Thanks!

jraysajulga:
Great. Thank you!

PR #335, opened 2019-01-15
SearchGUI - PTMs are loaded from a macros file.

CarlosHorro:
PTMs are moved from searchgui_mods.loc.sample to macros_modifications.xml file
Changed ptm_score_selector default value from A-score (0) to PhosphoR


bgruening:
We need to bump the tool version I think. @CarlosHorro should we merge this or wait until a new version of PS/SG is released?

Thanks for catching up with this!

CarlosHorro:
> We need to bump the tool version I think. @CarlosHorro should we merge this or wait until a new version of PS/SG is released?
> 
> Thanks for catching up with this!

As this is a version-independent improvement, I think we can just merge this update without needing new SG/PS apps versions...

bgruening:
Merci!

PR #343, opened 2019-02-07
Add unified msconvert.xml wrapper using Docker container

chambm:
Add unified msconvert.xml wrapper using the Docker container `chambm/pwiz-skyline-i-agree-to-the-vendor-licenses`; it uses Wine to run the Windows version of msconvert on Linux, and it works for almost all vendor formats
Add functional tests and test data for Thermo RAW, Bruker BAF, Bruker TDF, Waters MassLynx, Agilent MassHunter, and Sciex WIFF (Agilent/Bruker YEP does not work, and WIFF2 might not work)
Add General Options section, with combineIonMobilitySpectra, scanSumming, simAsSpectra, srmAsSpectra, acceptZeroLengthSpectra, ignoreUnknownInstrumentError, and runIndexSet
Change output format of refinement file from csv to tsv
Convert from external msconvert_wrapper.py to only Bash/Cheetah


tools/msconvert/msconvert_macros.xml
```diff
@@ -379,6 +422,37 @@
...
+      </conditional>
+
+      <param argument="--simAsSpectra" type="boolean" label="SIM as Spectra" help="Write selected ion monitoring as spectra, not chromatograms" />
+      <param argument="--srmAsSpectra" type="boolean" label="SRM as Spectra" help="Write selected reaction monitoring as spectra, not chromatograms" />
+      <param argument="--acceptZeroLengthSpectra" type="boolean" label="Accept zero-length spectra" help="Some vendor readers have an efficient way of filtering out empty spectra, but it takes more time to open the file" />
+      <param argument="--ignoreUnknownInstrumentError" type="boolean" label="Ignore unknown instrument error" help="If true, if an instrument cannot be determined from a vendor file, it will not be an error" />
+
+      <conditional name="multi_run_output">
+        <param name="do_multi_run_output" type="boolean" truevalue="true" falsevalue="false" label="Output multiple runs per file" help="Some input types can store multiple runs (samples) in a single file (e.g. WIFF). Each run must be written to a separate output file, so check this option if you want to output all runs for a file (each file will create a dataset collection)" />
```
bgruening: IUC guidelines state to use a select-param in conditionals.

tools/msconvert/msconvert_macros.xml
```diff
@@ -21,40 +22,80 @@
...
+    #end if
+    
+    #if $data_processing.precursor_refinement.use_mzrefinement
+      #set input_ident_name = ".".join((os.path.splitext($basename)[0], $data_processing.precursor_refinement.input_ident.ext))
+      #set output_refinement_name = os.path.splitext($basename)[0] + '.mzRefinement.tsv'
+      ln -s '$data_processing.precursor_refinement.input_ident' '$input_ident_name' &&
+    #end if
+
+    wine64 msconvert '${basename}'
+    --outdir `dirname ${output}`
```
bgruening: This is specifying the galaxy database dir. This is dangerous as msconvert could write into this folder without Galaxy knowing. So these files will live there forever.

I would recommend using `./out/` here and move the file to the final destination. For --outfile I would use `./out/foo.dat` or something similar.

chambm: You're worried that msconvert will write some file there other than the named one (--outfile)? It won't but I admit it requires knowledge of its internals to not worry about that. A bigger issue is I didn't yet add support for the `multi_run_output` because I'm not sure what that would look like. It would certainly change the outdir/outfile (outfile would cause all the files to be written to the same name). In that case all the files for the collection would need to be dynamically collected from outdir, so it wouldn't make sense to write directly to `dirname ${output}`. Can you help me with what the `<outputs>` element should look like in the `multi_run_output==true` case?

bgruening: Yes I'm worried. Anything against using the current-working-dir?

What about https://github.com/galaxyproject/galaxy/blob/dev/test/functional/tools/collection_split_on_column.xml#L9 ?

tools/msconvert/msconvert_macros.xml
chambm: Yes, and in general I still disagree with that guideline. :) Seeing "no" and "yes" (or a repeat of the label) in a select box is jarring to me. Much more so than options appearing/disappearing/changing based on the yes/no slider.

But reviewing all the current msconvert options there's at least one place where it would make sense to switch from boolean to something else: `Filter MS levels`. I think that would look better as a repeat, like the adjacent `Filter Scan Indices` and "Filter Scan Numbers`.

bgruening: ;) you can start discussing this with the IUC. But we could also not use Yes/No in the select, but maybe some short sentence.

chambm:
> > #set basename = $re.sub(r'[^\w.-+]','_',$input.display_name)
> 
> We should use element_identifier here I guess, to also support collections. `display_name` should not be used imho. Is the basename really needed in all cases?
Yes basename is needed, or whatever its equivalent is for collections. Many sources don't contain the sample name, and propagating the sample/run name is something that's just expected in proteomic workflows.


> What are the license conditions of the container. How do we make sure that the user-agrees to them?
The new shed.yml links to the license page. No real way to make sure the user agrees to them. Maybe a boolean param like "Do you agree to the licenses?" which defaults to No? :) Anyway I won't lose sleep over it.

> Any idea how we can enable the tests on this repo?
Hmm. I thought adding the `.shed.yml` would do it but there's 2 pieces missing:
1. New directory-based datatypes in the 19.01 PR. No idea how to deal with this here.
2. Requires running in Docker. Do you know if Docker in Docker is possible in Travis?

> @chambm I don't need to mentione how awesome this is, do I? Thanks a bunch!
Thanks!

bgruening:
> Yes basename is needed, or whatever its equivalent is for collections. Many sources don't contain the sample name, and propagating the sample/run name is something that's just expected in proteomic workflows.

ok, please use `.element_identifier`

> The new shed.yml links to the license page. No real way to make sure the user agrees to them. Maybe a boolean param like "Do you agree to the licenses?" which defaults to No? :) Anyway I won't lose sleep over it.

Yes. I think this is a good idea. As the shed.yml and the TS is also not exposed to the user. I like your idea a lot. Let's include this boolean and also put this license thing into the readme.

> 1. New directory-based datatypes in the 19.01 PR. No idea how to deal with this here.

As soon as it gets merged, we can use 19.01 here.

> 2. Requires running in Docker. Do you know if Docker in Docker is possible in Travis?

Travis is running in a VM, isn't it? I don't have a good idea at the moment. Singularity works as well? Need to think about it ... I will run it locally and can merge if we don't find a good solution.

chambm:
OK I think that resolves all the issues except testing with CI. I was happy that the license agreement could be validated client-side.

bgruening:
I'm not a lawyer, but that's best effort I think.

Thanks @chambm! I will test this locally asap and merge. I also need to push from local as if the tests are not passing nothing is pushed.

Thanks again!
Bjoern

chambm:
Maybe we should disable tests somehow on CI then so it can be automatically updated?

bgruening:
Galaxy starts containers with --user 1000:1000 it would be nice if the container can support this, so this tool runs more or less out of the box.

chambm:
You mean make the /wineprefix64 owned by 1000? That seems too Galaxy-specific to do for the official pwiz Docker image. I ran into the same problem but it was running with 1450 by default, not 1000. I haven't solved it yet, but thinking about it some more now I wonder if we can just add `chown $UID -R $WINEPREFIX &&` before running the wine command.

bgruening:
We could, but only if /wineprefix is now owned by root. The problem is not Galaxy imho, but that the container can not be run as `user`. Which is a wine problem more or less. Is there a way to set `WINEPREFIX` to something else, but point wine to use C:\ in /wineprefix64?

bgruening:
`|  chown: changing ownership of '/tmp/tmpZVLWzM/files/000/dataset_12.dat': Operation not permitted` is what travis is showing.

bgruening:
@chambm awesome! I think this is working now.

If you have time, maybe you can look into travis and hack something up that only in the case of msconvert the --docker argument is used?

bgruening:
Only 3 tests fail, but this is due to test-data updates, like:

```
-      <software type="conversion" name="ProteoWizard software" version="3.0.19044"/>
+      <software type="conversion" name="ProteoWizard software" version="3.0.19052"/>
```

Maybe adding lines_diff is the way to go here. So you can update the container without updating the test-data everytime.

bgruening:
@chambm these are your errors now :)

sudo is missing I guess. And do we need a test for those multiple outputs?

bgruening:
Yes, a first one is green! :)

bgruening:
Thanks a lot @chambm!

PR #347, opened 2019-02-14
Filter by fasta

pravs3683:
Added a feature where id list can be accession number in the fasta header (necessarily not the complete fasta header). 
Eg. filter fasta id based on Uniprot accession number from a Uniprot like fasta file. 

This modification uses pattern recognition to match the accessions in the header.


tools/filter_by_fasta_ids/filter_by_fasta_ids.xml
```diff
@@ -51,6 +54,24 @@ $dedup
...
+                
+                
+                <conditional name="id_regex">
+                    <param name="find" type="select" label="Match IDs by">
+                        <option value="beginning">ID is expected at the beginning: &gt;ID</option>
+                        <option value="pattern">regex pattern</option>
+                    </param>
+                    <when value="beginning"/>
+                    <when value="pattern">
+                        <param name="pattern" type="text" value="" label="regex search pattern for ID">
```
bgruening: The default value in the python script is `>([^| ]+)` isn't it? Maybe we should add this here or mention it in the help of this param.

tools/filter_by_fasta_ids/filter_by_fasta_ids.py
```diff
@@ -64,14 +63,21 @@ def main():
     parser.add_argument('-d', help='Path to discarded entries file')
     header_criteria = parser.add_mutually_exclusive_group()
     header_criteria.add_argument('--id_list', help='Path to the ID list file')
+    parser.add_argument('--pattern', default=None, help='regex earch attern for ID in Fasta entry')
```
bgruening: "search pattern" ... mention the implicit default value as well?

tools/filter_by_fasta_ids/filter_by_fasta_ids.py
```diff
@@ -64,14 +63,21 @@ def main():
...
     sequence_criteria.add_argument('--sequence_regexp', help='Regular expression pattern the header should match')
     parser.add_argument('--max_length', type=int, help='Maximum sequence length')
     parser.add_argument('--dedup', action='store_true', default=False, help='Whether to remove duplicate sequences')
     options = parser.parse_args()
-
+    
+    
+    pattern =  options.pattern if options.pattern else '>([^| ]+)'
+    if not re.match('^.*[(](?![?]:).*[)].*$',pattern):
+        print('pattern: "%s" did not include "%%s"' % pattern)
```
bgruening: is %% really want you want?

pravs3683: Yes.

pravs3683: We want user to specify ( ) in the regex for capturing group which can be used to pass in the string to match the fasta header.

bgruening: I mean in the string substitution :)

tools/filter_by_fasta_ids/filter_by_fasta_ids.xml
pravs3683: Modified according to your suggestion

tools/filter_by_fasta_ids/filter_by_fasta_ids.py
pravs3683: Modified

tools/filter_by_fasta_ids/filter_by_fasta_ids.py
pravs3683: How about if I change it to 

print('pattern: "%s" did not include capture group "()" in regex ' % pattern)

bgruening: sure

pravs3683: Done!! Thanks a lot!!

bgruening:
Thanks!

PR #351, opened 2019-02-25
new maxquant tool

dglaetzer:
wrapper for a more up-to-date maxquant version. Doesn't extend the functionality of the old one (more the opposite), but hopefully is ready to be used soon.


bgruening:
@dglaetzer tests are now running but they are failing. You can see this here: https://travis-ci.org/galaxyproteomics/tools-galaxyp/jobs/498344074#L9989

dglaetzer:
The tests are passing on my machine but they seem to take too long for travis. We already tried to keep the raw files and searches as short and fast as possible. Any idea what we could do?

bgruening:
How long does it take on your machine locally?

dglaetzer:
around 20 minutes, I think.

bgruening:
@dglaetzer green! :)

tools/maxquant/macros.xml
```diff
@@ -0,0 +1,106 @@
...
+                </data>
+        </xml>
+        <xml name="output_option" token_label="default label" token_name="default">
+                <option value="@NAME@">@LABEL@</option>
+        </xml>
+        <xml name="mod_option" token_value="default">
+                <option value="@VALUE@">@VALUE@</option>
+        </xml>
+        <xml name="modification">
+                <expand macro="mod_option" value="Acetyl (K)"/>
```
bgruening: How did you build those lists? Are the scripts from above, which you removed useful for this? How can we update those long list of modifications easily in the future?

dglaetzer: I used the init.py script in the tool dir. It reads the modifications.xml and enzymes.xml from maxquant and writes them into macros.xml. After that, the macros.xml doesn't comply with your conventions anymore... i'll try to fix this.

bgruening: Ok, could you also add a README.md file to the repo that describes this procedure please?

bgruening:
@dglaetzer top! Any other feedback from your testers or is this ready to go :)

dglaetzer:
no more feedback at the moment... so i'd say it's ready.

tools/maxquant/test-data/exp_design_test.txt~
```diff
@@ -0,0 +1,3 @@
+Name	Fraction	Experiment	PTM
```
chambm: Was this file committed on accident?

chambm:
Have you tried a trimmed mzML or mzXML file to produce a shorter test? It is a real pain in the ass to get them to work with MaxQuant (if I had a dollar for every "MaxQuantGui has stopped responding..."!!!), but using open formats is likely to be something users will want to do. Restricting it to Thermo RAW files would be sad (although IIUC it's the only vendor format that works with MQ on Linux at the moment AFAIK).

chambm:
Here's the 3 test RAW files converted to mzXML, peak picked (profile MS1s are why the RAWs are so big with just 1min acquisition), and filtered to only the first 100 scans:
[Thermo.zip](https://github.com/galaxyproteomics/tools-galaxyp/files/2993794/Thermo.zip)

I tested that they do run through MaxQuant without error. 400K each instead of 2MB each. And at least it proves that mzXML should be a valid input to the tool (although it might not work for OTHER vendor formats converted to mzXML).

tools/maxquant/maxquant.xml
```diff
@@ -1,315 +1,273 @@
...
-      </sanitizer>
-    </macro>
-    <macro name="protease_opts">
-      <options from_data_table="proteases" />
-    </macro>
-    <macro name="group_params">
-      <param name="protease" label="Enzyme" type="select">
-        <expand macro="protease_opts" />
-      </param>
-      <param name="first_search_tol" label="First Search Tolerance (ppm)" type="float" value="20" />
```
chambm: Am I reading it correctly that the new wrapper no longer provides a way to interactively set most of the mqpar.xml settings (such as this precursor tolerance)?

bgruening: Yes, currently yes. It would be great to keep the same functionally as before, but on the other hand, this is the first version that really works out of the box. So we could merge that and add more and more options progressively. What do you think?

chambm: I agree, just wanted to clarify it was intended. Maybe add it to the PR description?

tools/maxquant/maxquant.xml
```diff
@@ -72,7 +72,8 @@
                help="Specify one or more Thermo RAW files" />
         <param format="fasta" multiple="true" name="fasta_files"
                type="data" label="FASTA files"
-               help="Specify one or more FASTA databases." />
+               help="Specify one or more FASTA databases. Parse rules are hard coded:
+                     '>([^\s]*)' for the identifier, '>(.*)' for the description." />
```
bgruening: You should use CDATA here I guess.

tools/maxquant/test-data/exp_design_test.txt~
dglaetzer: The file is needed for the doctests in mqparam.py, so not really necessary, but maybe helpful.

dglaetzer:
@chambm Sorry for not responding to your review, I was on holiday... Thanks for converting the raw files, I wasn't aware of that option. I'll change the tests to use these files instead of the original ones. The results of maxquant seem to differ between a .raw file and the corresponding mzxml file (without any other modifications) though, but for the tests it should be ok. Feel free to reuse the mqparam.py script in your project, if it is useful to you.

tools/maxquant/test-data/exp_design_test.txt~
chambm: Sorry, I meant this looks like a temporary editor file. The file without the `~` at the end is also committed.

tools/maxquant/maxquant.xml
```diff
@@ -73,7 +73,7 @@
         <param format="fasta" multiple="true" name="fasta_files"
                type="data" label="FASTA files"
                help="Specify one or more FASTA databases. Parse rules are hard coded:
-                     '>([^\s]*)' for the identifier, '>(.*)' for the description." />
+                     '&gt;([^\s]*)' for the identifier, '&gt;(.*)' for the description." />
```
bgruening: You can also use CDATA here ...

bgruening:
@dglaetzer @chambm please test this tool here: https://usegalaxy.eu/root?tool_id=maxquant

bernt-matthias:
Just pointed one of our users who is interested in maxquant to the usegalaxy installation. Lets see if he has feedback.

tools/maxquant/maxquant.xml
```diff
@@ -92,24 +93,24 @@
...
             <param name="choose_mode" type="select" label="choose mode">
                 <option value="interactive">specify search parameters</option>
                 <option value="mqpar">specify mqpar.xml</option>
             </param>
             <when value="interactive">
+                <conditional name="fractions">
+                    <param name="set_fractions" type="boolean"
+                           label="specify fractions?" checked="false"/>
+                    <when value="true">
+                        <param format="tabular" name="template" type="data"
```
bgruening: @dglaetzer you can remove the entire conditional here and make the data-param `optional="True"` that serves the same purpose.

tools/maxquant/maxquant.xml
```diff
@@ -23,9 +23,7 @@
     --version='@VERSION@'
     --fasta_files='$fasta_files'
     #if $mode.choose_mode == "interactive"
-        #if $mode.fractions.set_fractions=="true"
-            --exp_design='$mode.fractions.template'
-        #end if
+        --exp_design='$mode.template'
```
bgruening: Please note that $mode.template can be None ... in the case the user does not provide the file. So the if is probably needed.

tools/maxquant/mqwrapper.py
```diff
@@ -74,7 +74,8 @@
     mqpar_in = mqpar_temp
 mqpar_out = args['mqpar_out'] if args['mqpar_out'] != 'None' else mqpar_temp
 
-m = mqparam.MQParam(mqpar_out, mqpar_in, args['exp_design'],
+exp_design = args['exp_design'] if args['exp_design'] != 'None' else None
+m = mqparam.MQParam(mqpar_out, mqpar_in, exp_design,
```
dglaetzer: That's why I made this change. Should work, shouldn't it?

bgruening: If it's really 'None' that is passed through yes, it's a little bit unusual to pass None through the command line but yes, this should work.

bernt-matthias:
Few points from one of our users (who was in general very happy with the tool). He misses some options:

- heavy modifications: option deuterium 2H is missing. 
- device  and mass settings (mass tolerance in MS1 and MS2), Peptide length (Max and min)
- FDR settings 

In particular the first is missed a lot. For the other ones he seemed to be fine with using the defaults for now. 


@bgruening The other user couldn't upload 1.5GB input files to usegalaxy.eu. Any tipps? Should he use ftp?

bgruening:
> @bgruening The other user couldn't upload 1.5GB input files to usegalaxy.eu. Any tipps? Should he use ftp?

please use FTP at the moment. We limited it to 1GB for the moment. For you as information, it seems that uwsgi is producing heavy IO during an upload and this makes problem on our head-node when many users upload files in parallel.

dglaetzer:
@bernt-matthias 

> (who was in general very happy with the tool)
>
Good to hear :)
 
>     * heavy modifications: option deuterium 2H is missing.
> 
Can you clarify what is meant by this? What is the exact composition? The problem might be that the required label is not in the standard modifications.xml of maxquant 1.6.3.4. Adding entries to this file is a problem I have no solution for yet.

>     * device  and mass settings (mass tolerance in MS1 and MS2), Peptide length (Max and min)
> 
min peptide length and max peptide mass are coming soon. The other options could take a while...

bernt-matthias:
> > ```
> > * heavy modifications: option deuterium 2H is missing.
> > ```
> 
> Can you clarify what is meant by this? What is the exact composition? The problem might be that the required label is not in the standard modifications.xml of maxquant 1.6.3.4. Adding entries to this file is a problem I have no solution for yet.

Seems that this is what the user would like to have in maxquant (they label their proteins with deuterium). Maybe this is not possible with maxquant at the moment. Where is this `modifications.xml` file and maybe some docs. If I find the time I will have a look.

> > ```
> > * device  and mass settings (mass tolerance in MS1 and MS2), Peptide length (Max and min)
> > ```
> 
> min peptide length and max peptide mass are coming soon. The other options could take a while...

Wonderful.

bgruening:
The "perform LFQ" boolean question should be a select.
https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#booleans

bgruening:
After GCC and so many people talking about maxquant in Galaxy, we should probably merge this soon and add features in a second iteration?

@dglaetzer do you have time to focus on UX and tests for this version and then we will merge it?

Any other suggestion?

dglaetzer:
I agree. One test should imo be sufficient for the mqpar version of the tool. I can add a second one for LFQ to the other version. SILAC would still remain untested, but already has proven to deliver good results (in a test scenario), so this can wait a bit longer, i think. If i get the LFQ data soon enough, i would suggest to merge it end of next week.

bgruening:
Ok, I will come back next week to this PR :) Thanks @dglaetzer

dglaetzer:
@bgruening I didn't manage to create lfq tests. Maxquant keeps failing with my small files - either while 'calculating peak properties' or during the normalization step. The results from searches with 'real' LFQ data look good though.
So I'd be ok with merging it as it is. If I get some suitable data I will add more tests. Or maybe just test the creation of a correct mqpar file.

bgruening:
Let's get this one in and call for broader testing. I will review in the evening the last time and probably will merge it if no one else complains before :)

Thanks!
Bjoern

dglaetzer:
yes, there have been a few unnecessary files. Hope I found them all now...

bgruening:
Merci! This is awesome!!!! Thanks so much!

bgruening:
fyi: https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/galaxyp/maxquant/maxquant/1.6.3.4

PR #352, opened 2019-02-27
Samples File Creation Tool for metaQuantome

jraysajulga:
Looking to add another module to the metaquantome branch.

@caleb-easterly With this tool, we might need to remove the MAKE_SAMPS_TSV token in the macros.xml, unless we are thinking of incorporating this in each of the other modules.


jraysajulga:
@caleb-easterly Let me know if you want me to update the version in the macros.xml file. Currently, it is 0.99.2.g1.

caleb-easterly:
Thanks @jraysajulga! I'm doing some tweaking but I'll merge it for now

bgruening:
@caleb-easterly please note that every merge will trigger a TS upload.
We should minimize the TS version and only merge if we really want to have a upload.

I will stop the upload ok? And then wait for your changes?

caleb-easterly:
@bgruening I assumed that was only true for merges into master. Is that true for merges into any branch? 

But yes, I wasn't intending to upload to toolshed, so if that is happening stopping it would be great. Thanks!

bgruening:
Uh, my fault. Sorry you are completely correct.

PR #354, opened 2019-03-01
Update dia_umpire_se

jj-umn:
Allow for use without dia_umpire_quant.
Add test to dia_umpire_se.xml  using smallest mzXML that would produce
any MGF output.


jj-umn:
@bgruening   I only updated dia_umpire_se.xml   I don't think anyone is interested in the dia_umpire_quant application anymore.  Perhaps we can delete that tool, and remove the dia_umpire.ser output from dia_umpire_se.xml  along with the datatype.xml

bgruening:
> @bgruening I only updated dia_umpire_se.xml I don't think anyone is interested in the dia_umpire_quant application anymore. Perhaps we can delete that tool, and remove the dia_umpire.ser output from dia_umpire_se.xml along with the datatype.xm

Sounds good. But maybe we move it into the deprecated folder and not remove it completely?

Thanks for working on it!

bgruening:
@jj-umn thanks a lot!

PR #358, opened 2019-03-18
update wrapper to use metaquantome version 0.99.4

caleb-easterly:
also fix HTML display


bgruening:
@caleb-easterly the test was still failing ....

caleb-easterly:
yes, i thought i'd keep working on the error on the metaquantome branch. Do you know what it might be due to? I'm seeing `RunToolException: Error creating a job for these tool inputs - The selected case is unavailable/invalid.` That sounds to me like a parameter is incorrect in some way, but that is confusing because its running on my local computer when I use `planemo test`

PR #364, opened 2019-03-28
Fix metaquantome_db

jraysajulga:
A fix to the issue described [here](https://github.com/galaxyproteomics/metaquantome/issues/28).


caleb-easterly:
thanks - looks good!

PR #368, opened 2019-04-14
openms: aggressive error checks instead of stdio

bernt-matthias:
I had repeated problems with OOMs that are indicated by `std:bad_alloc` which is currently not caught by the tools. Instead of adding it to stdio I suggest to switch to aggressive error checks. 

Galaxy instances without https://github.com/galaxyproject/galaxy/pull/7095 (currently only in dev) will loose the checks in stdio, but I think only the qUncompress message is a problem. Reason is that I'm now convinced that the other two can be avoided by setting Xmx (etc) properly .. maybe also resubmission doesn't help here.

If we find out which tool(s) are affected by the qUncompress message, then we could revert to stdio for those tools.

- also added cheetah for all command blocks


bgruening:
@bernt-matthias is the plan to only merge this after 19.05? Or do we want to add a profile?

bernt-matthias:
My suggestion would be to merge it now (after fixing the linter issues) without adding a profile, because the errors that are not caught due to the change are (probably) irrelevant.

bgruening:
Merci!

bernt-matthias:
Hi @bgruening seems that the push to the toolshed also failed for these :(

bgruening:
Yeah, tests are failing with openms, because of linting... pushing, but this one takes a while.

bgruening:
I think I killed the TS :(

bgruening:
Done!

PR #369, opened 2019-04-15
Update for gffcompare v0.11.2

jj-umn:
Adds class_code ~ for intron chain match or single exon when gffcompare
is run with —strict-match


jj-umn:
@bgruening Yes.   That's what prompted me to update this tool.   gffCompare now has a  --strict-match option which can output a new classification code: '~'

PR #370, opened 2019-04-16
update metaquantome help documentation

jraysajulga:
Just some minor fix suggestions in the help documentation for the expand and database modules.

For the database module, I wasn't entirely sure what this describes:

> Also, note that the databases will be stored and not modified, so the date of download can be referenced later. The important exception to this is the NCBI database, which is updated every time it is loaded by the ete3 package, used within metaQuantome.

Does this mean that the NCBI database is built in to metaQuantome? In other words, it is not downloaded into the output Galaxy dataset?

Also, I saw that the visualize module does not get the correct column for the **Taxonomic rank to restrict to in the plot** in the **Taxonomic analysis** mode. As it is on MPG, I see that it gets the namespace column (index 3 I believe). Seems like you fixed this to be the 5th index, which looks correct.

### DIFF
--- a/tools/metaquantome/metaquantome_db.xml
+++ b/tools/metaquantome/metaquantome_db.xml
@@ -39,9 +39,10 @@
     <help>
 <![CDATA[
 metaQuantome database
-===================
+=====================
 
 metaQuantome uses freely available bioinformatic databases to expand your set of direct annotations.
+For most cases, all 3 databases can be downloaded (the default).
 
 The databases are:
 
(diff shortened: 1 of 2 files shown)


caleb-easterly:
Thanks @jraysajulga!

The documentation could maybe be a little clearer on the taxonomy database point. It is included in the output Galaxy dataset, but we just can't guarantee that it will stay exactly the same. Is there a way I could say that better?

jraysajulga:
No problem @caleb-easterly.

I'm just confused about how it differs with the other databases? Doesn't this module take the latest databases for all three?

caleb-easterly:
Yes - but the code for the NCBI database will update it automatically on disk. So if I download all 3 databases today, then the GO and EC databases will stay "stuck" unless I download new ones. But the `ete3` package will automatically detect the presence of a new NCBI database and update it, with no option to do otherwise. Does that make sense? it's hard to explain

jraysajulga:
That makes more sense, I think. So, the `ete3` package will update the NCBI database in the Galaxy  history every time metaquantome is run?

caleb-easterly:
Strictly speaking, every time `metaquantome expand` is run. But that is correct.

jraysajulga:
> Also, note that the databases will be stored and not modified, so the date of download can be referenced later. The important exception to this is the NCBI database, which is updated every time it is loaded by the ete3 package, used within metaQuantome.

Okay. I see what you're saying now. What threw me off is the "The important exception to this..." bit because I thought that referred to the storage part as well. Maybe we can separate the "be stored" and the "not modified" bits. For example:

> Also, note that the databases will be stored in the history so that the date of download can be referenced later. This means that the databases will not be modified, except for the NCBI database, which is automatically detected in the history and updated every time it is loaded by the ete3 package, used within metaQuantome expand.

caleb-easterly:
Sounds good! I adapted that a bit and pushed.

PR #377, opened 2019-05-08
Option for exporting reports out of the zip file

CarlosHorro:
Added option for exporting reports even when "zip" option is activated.
ptm_score_selector default value changed to PhosphoRS.


CarlosHorro:
> @CarlosHorro looks good. Can you please add expet_num_outputs (https://docs.galaxyproject.org/en/latest/dev/schema.html#id77) to the tests, so we know the complex filter do work?

I have included it and it was great because I found out a bug in the mzidentML file filter :-)

bgruening:
> I have included it and it was great because I found out a bug in the mzidentML file filter :-)

you are very welcome :)

bgruening:
We had 1000 PS runs in the last months... this is a very important tool, thanks for maintaining it!!!

PR #378, opened 2019-05-13
filter_by_fasta_id: speedup by using set

bernt-matthias:
For longer id lists the tool was horribly slow. I suggest to use `set` instead of `list` to store the ID list which speeds up the calculations dramatically and should not hurt in other respects.


bernt-matthias:
Failing test seems unrelated.

bgruening:
Was ist denn das für ein Fehler, narf ...

pravs3683:
I would also suggest using dictionaries (by making accession ids as key and
its corresponding sequence, including header, as values). This makes it
much faster.

bernt-matthias:
@pravs3683 thanks for the suggestion, but I don't see how this could help.

bernt-matthias:
hooray

bgruening:
@pravs3683 the set() should not be slower then a dict() right?
I'm going to megre this.

Thanks @bernt-matthias!

bernt-matthias:
Thanks @bgruening . Has galaxyp an automatic upload to the toolshed?

bgruening:
Yes. But the packages failed to upload on master. So I pushed manually this time. Check again.

Issue #390, opened 2019-08-07
Maxquant linking error


bernt-matthias:
```
Traceback (most recent call last):
  File "/gpfs1/data/galaxy_server/galaxy/shed_tools/toolshed.g2.bx.psu.edu/repos/galaxyp/maxquant/8bac3cc5c5de/maxquant/mqwrapper.py", line 94, in <module>
    os.link(f, l)
OSError: [Errno 18] Invalid cross-device link: '/gpfs1/data/galaxy_server/galaxy/database/files/081/dataset_81846.dat' -> '21-12-2018_S81_GF_F1_HP.thermo.raw'
```

I guess os.symlink will do the trick.

Ping @dglaetzer

PR #391, opened 2019-08-07
maxquant: use symlink instead of hard link

bernt-matthias:
fixes https://github.com/galaxyproteomics/tools-galaxyp/issues/390


bernt-matthias:
damn .. forget to increase the tool version again. Should I still do this?

bgruening:
All good. I think this is a safe replacement.

PR #397, opened 2019-08-24
Thermo raw file parser updated to 1.1.10 & most options supported

CarlosHorro:
Supported:
- new input directory option (ie: supported conversion of multiple input raw files at the same time)
- indexed mzML format.
- json metadata
- verbose logging option
- gzip option
- "ignore missing instrument properties" option
- zlib compression option
- peack peacking option

And added real tests.


tools/ThermoRawFileParser/thermo_converter.xml
```diff
@@ -1,60 +1,227 @@
...
+          optional="False" />
+
+        <param name="output_format" type="select" label="Output format">
             <option value="0">mgf</option>
             <option value="1" selected="True">mzml</option>
+            <option value="2">Indexed mzml</option>
+        <!--    <option value="3">Parquet</option> -->
+        </param>
+
+        <param name="gzip_boolean" type="boolean" truevalue="1" falsevalue="0" checked="false"
```
bgruening: I would not offer this option. No tool can deal with this at the moment.

tools/ThermoRawFileParser/thermo_converter.xml
```diff
@@ -1,60 +1,227 @@
...
     -f=$output_format
-    #if $metadata:
-        --metadata=1
+    #if $gzip_boolean == 1:
+    -g
+    #end if
+    #if $output_metadata_selector != "off":
+        --metadata="${output_metadata_selector}"
+    #end if
+    #if $zlib_boolean == 0:
```
bgruening: These are booleans you can use `truevalue="-z" falsevalue=""` and here you can save the if condition and just wite `$zlib_boolean`, which will evaluate to an empty string or `-z`

tools/ThermoRawFileParser/thermo_converter.xml
```diff
@@ -1,60 +1,227 @@
...
+            #else if $output_format == "1":
+                mv ./output_folder/input.mzML ./output_file.out &&
+            #else if $output_format == "2":
+                mv ./output_folder/input.mzML ./output_file.out &&
+            ## #else if $output_format == "3":
+            ##    mv ./output_folder/input.parquet ./output_file.out &&
+            #end if
+
+        #else  ## if we have gzipped output
+            #if $output_format == "0":
```
bgruening: without gzip this would be much easier and I don't think we can use this feature effectively ... we could if we dare by enabling all datatypes with compression="true" and getting a `mzml.gz` datatype ... but not sure this is worth it now

tools/ThermoRawFileParser/thermo_converter.xml
```diff
@@ -1,60 +1,227 @@
...
+        <param name="ignore_instrument_errors_boolean" type="boolean" truevalue="1" falsevalue="0" checked="true"
+            label="Ignore missing instrument properties" help="If false, it stops the conversion if instrument properties are missing" />
+
+        <param name="output_metadata_selector" type="select" label="Output metadata" >
+            <option value="off" selected="True">No</option>
+            <option value="0">json</option>
+            <option value="1">txt</option>
         </param>
+
+        <param name="verbose_boolean" type="boolean" truevalue="1" falsevalue="0" checked="false"
```
bgruening: I would not expose this option to the user.

tools/ThermoRawFileParser/thermo_converter.xml
```diff
@@ -1,60 +1,227 @@
...
+        <test>
+            <param name="input" value="really_small.raw"/>
+            <param name="output_format" value="1"/>
+            <param name="gzip_boolean" value="1"/>
+            <output name="output_gzipped" file="really_small.mzml.gzip" ftype="zip" compare="sim_size" delta="3000"  >
+            </output>
+        </test>
+
+        <!-- Testing contents of converted mgf file-->
         <test>
```
bgruening: you can use `expect_num_outputs` to enhance your tests for all your output filers.

CarlosHorro: Um... using that option, do you know if there is a way to know which outputs are found when running the tests? I am surprisingly having errors on all tests (on both non zipped and zipped output tests) when using "expect_num_outputs="1":
`
Exception: Incorrect number of outputs - expected 1, found 2.
`

I do not know how it may be possible...

tools/ThermoRawFileParser/thermo_converter.xml
```diff
@@ -116,12 +108,12 @@ mkdir ./output_folder &&
                 <when input="output_format" value="2" format="txt" />
             </change_format>
             <filter>len(input) == 1</filter>
-            <filter>gzip_boolean == 0</filter>
+            <filter>gzip_boolean is not True</filter>
```
bgruening: not gzip_boolean should also work imho

tools/ThermoRawFileParser/thermo_converter.xml
```diff
@@ -26,28 +26,20 @@ mkdir ./output_folder &&
...
+    $peakpicking_boolean
+    $ignore_instrument_errors_boolean
 
     2>> $temp_stderr)
 
     &&
 
     #if len($input) == 1:
-        #if $gzip_boolean == "0":
+        #if $gzip_boolean == "":
```
bgruening: if $gzip_boolean

tools/ThermoRawFileParser/thermo_converter.xml
bgruening: I recommend grepping in https://github.com/galaxyproject/galaxy/tree/dev/test/functional/tools and getting inspiration. This should work. If you have collection the elements count differently. I think it was every element + 1 for the collection.

tools/ThermoRawFileParser/thermo_converter.xml
bgruening: @CarlosHorro maybe we drop this option and make the wrapper way simpler?

CarlosHorro: Yep I will do it :-)

tools/ThermoRawFileParser/thermo_converter.xml
CarlosHorro: I am quite surprised with the way it is working.

This works:
```

        <!-- Basic mzml collection test -->
        <test expect_num_outputs="1">
            <param name="input" value="really_small.raw,really_small_2.raw"/>
            <param name="output_format" value="1"/>
            <output_collection name="output_mzml_collection" type="list" count="2"/>
        </test>
        
        <!-- mgf collection test with metadata -->
        <test expect_num_outputs="2">
            <param name="input" value="really_small.raw,really_small_2.raw,really_small_3.raw"/>
            <param name="output_format" value="0"/>
            <param name="output_metadata_selector" value="0"/>
            <output_collection name="output_mgf_collection" type="list" count="3"/>
            <output_collection name="output_metadata_collection" type="list" count="3"/>
        </test>
```

So, at least with collections, it seems that "expect_num_outputs" just count the number of <output_collection> elements with content in the output.

But both single input tests, as
```

<!-- Basic test -->
        <test expect_num_outputs="1">
            <param name="input" value="really_small.raw"/>
            <param name="output_format" value="1"/>
            <output name="output" file="really_small.mzml" ftype="mzml" compare="sim_size" delta="3000" />
        </test>
```
fails with

Incorrect number of outputs - expected 1, found 2.

I don’t see any way of that test producing 2 outputs… apart from being also counting the “output_metadata” output (although it is empty)?

CarlosHorro: > I recommend grepping in https://github.com/galaxyproject/galaxy/tree/dev/test/functional/tools and getting inspiration. This should work. If you have collection the elements count differently. I think it was every element + 1 for the collection.

I am doing it but it do not find many uses of "expect_num_outputs" specially mixed with single outputs, collections and things like "change_format"...

bgruening: Try this one: https://github.com/galaxyproject/tools-iuc/search?l=XML&q=expect_num_outputs

bgruening: Mh, does planemo server generate the metadata output? Maybe <filter>output_metadata_selector != "off"</filter> is wrong and it needs to be `<filter>str(output_metadata_selector) != "off"</filter>`

CarlosHorro: Tank you for the examples... which metadata do you mean?
I am doing many checks and I have think that filter "len(input)>1" is not working/filtering when doing tests (but it seems to work perfectly when executing the tool) as output_metadata_collection seems to be always present even when I introduced just one input...
If I put some impossible condition in output_metadata_collection like <filter>"1" == "2"</filter> then output_metadata_collection is not generated and the number of outputs when generating single "output" and "output_metadata"  is 2.. If I do not specify that kind of impossible condition, expect_num_outputs is 3.

CarlosHorro: Well, after some try & error, it seems the filter works wih
`
(str(input)).count(',') == 0
`
:-)

tools/ThermoRawFileParser/thermo_converter.xml
```diff
@@ -1,60 +1,193 @@
...
+        <param name="input" type="data" format="thermo.raw" label="Thermo RAW file" help="" multiple="true"
+          optional="False" />
+
+        <param name="output_format" type="select" label="Output format">
             <option value="0">mgf</option>
             <option value="1" selected="True">mzml</option>
+            <option value="2">Indexed mzml</option>
         </param>
+
+        <param name="zlib_boolean" type="boolean" truevalue="" falsevalue="-z" checked="true"
```
bgruening: Is this parameter still needed if we do not offer compressed outputs?

tools/ThermoRawFileParser/thermo_converter.xml
```diff
@@ -1,60 +1,193 @@
...
+        </param>
+
     </inputs>
     <outputs>
-        <data format="mzml" name="output" from_work_dir="input.mzML" label="${tool.name} on ${on_string}">
+
+        <!-- We use simple data outputs if we just have one file, for backwards compatibility -->
+
+        <data name="output" format="mzml" from_work_dir="output_file.out" label="${tool.name} on ${on_string}">
+            <filter>(str(input)).count(',') == 0</filter> <!-- funny way of counting the number of input files! -->
```
bgruening: wired ... we should create a simple Galaxy tool that confirms this and add it to Galaxy core as a test tool ...

tools/ThermoRawFileParser/thermo_converter.xml
CarlosHorro: Well, zlib compression is used by default (-z flag disables it, not the opposite) and its compression is quite notorious (4mb size vs 2mb, applied to the small small.raw file - 1.4 mb ) although. I am not an expert but I think the ability of disabling this compression may be useful in some contexts without proper zlib management or for readability purposes...

tools/ThermoRawFileParser/thermo_converter.xml
CarlosHorro: Well, I am just counting the number of commas present in the input hehe. The really weird thing (and dangerous, because is misleading) is what happens with the len(input) function in the filter, which seems to work nicely when running the tool itself but it does not work with the tests (seems to crash internally but the condition is approved)... :-S

CarlosHorro:
Any more action on my side?

bgruening:
Thanks for all you help!

PR #399, opened 2019-09-28
manual changes based on user request

bgruening:
ping @bernt-matthias ... fyi

### DIFF
--- a/tools/openms/OpenSwathWorkflow.xml
+++ b/tools/openms/OpenSwathWorkflow.xml
@@ -10,14 +10,25 @@
   <expand macro="references"/>
   <expand macro="stdio"/>
   <expand macro="requirements"/>
-  <command detect_errors="aggressive"><![CDATA[OpenSwathWorkflow
+  <command detect_errors="exit_code"><![CDATA[
+
+  #for $i, $token in enumerate($param_in):
+    ln -s '${token}' ${i}.${token.ext} &&
+  #end for
+
+ln -s '${param_tr}' ./input_tr.${param_tr_type} &&
+
+
+OpenSwathWorkflow
 
 -in
-  #for token in $param_in:
-    $token
+  #for $i, $token in enumerate($param_in):
+    ${i}.${token.ext}
   #end for
+
+
 #if $param_tr:
-  -tr $param_tr
+  -tr ./input_tr.${param_tr_type}
 #end if
 #if $param_tr_type:
   -tr_type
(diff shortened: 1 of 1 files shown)


bernt-matthias:
@bgruening can you add a bit on the motivation of the changes (would be good to know for the automatic converter). I add specific questions to the code.

tools/openms/OpenSwathWorkflow.xml
```diff
@@ -10,14 +10,25 @@
   <expand macro="references"/>
   <expand macro="stdio"/>
   <expand macro="requirements"/>
-  <command detect_errors="aggressive"><![CDATA[OpenSwathWorkflow
+  <command detect_errors="exit_code"><![CDATA[
```
bernt-matthias: Why change from aggressive to exit_code?

tools/openms/OpenSwathWorkflow.xml
```diff
@@ -10,14 +10,25 @@
...
+
+  #for $i, $token in enumerate($param_in):
+    ln -s '${token}' ${i}.${token.ext} &&
+  #end for
+
+ln -s '${param_tr}' ./input_tr.${param_tr_type} &&
```
bernt-matthias: Could the extension be used to eliminate param_tr_type

tools/openms/OpenSwathWorkflow.xml
```diff
@@ -10,14 +10,25 @@
   <expand macro="references"/>
   <expand macro="stdio"/>
   <expand macro="requirements"/>
-  <command detect_errors="aggressive"><![CDATA[OpenSwathWorkflow
+  <command detect_errors="exit_code"><![CDATA[
+
+  #for $i, $token in enumerate($param_in):
+    ln -s '${token}' ${i}.${token.ext} &&
```
bernt-matthias: Use element_identifier instead of i

tools/openms/OpenSwathWorkflow.xml
```diff
@@ -30,15 +41,19 @@
...
-#if $param_out_features:
-  -out_features $param_out_features
-#end if
-#if $param_out_tsv:
-  -out_tsv $param_out_tsv
-#end if
-#if $param_out_osw:
-  -out_osw $param_out_osw
+
+## not recommended
```
bernt-matthias: Why not recommend

tools/openms/OpenSwathWorkflow.xml
```diff
@@ -30,15 +41,19 @@
...
+##if $param_out_features:
+##  -out_features $param_out_features
+##end if
+
+#if $out_format == 'tsv':
```
bernt-matthias: Can the tool only output one of both?

tools/openms/OpenSwathWorkflow.xml
bgruening: Because OpenMS is writing warnings or things like that with "Error:" ... its wired, but users ensured me that despite the fact the dataset is red the result is expected.

bernt-matthias: Good to know. Then I will adapt accordingly in the CTDconverter. 

Would be good to have an example to report upstream.

bernt-matthias:
Still needed?

PR #406, opened 2019-11-07
Add eggnog_mapper data_managers

jj-umn:
Add eggnog_mapper data_managers, for absolute path install, and relative to galaxy_data_manager_data_path.
The abspath DM allows one to incrementally add hmm_levels to an exisitng eggnog_data, and also alleviates hours of data copying.
Adding data_maangers required a change in the data_tables.  The files are now: eggnog_mapper_db.loc and eggnog_mapper_hmm_dbs.loc


bgruening:
Thanks @jj-umn! Just for your information, data managers can now have tests too :)

PR #407, opened 2019-11-19
Remove list-style terminating ,s from format attributes

wm75:
While this style is currently tolerated by Galaxy, it prevents correct param format detection in the API and, thus, interferes with scripts trying to extract meaningful info from your tool wrappers.

Autodetection of the pattern also fixed a param label :-)


bgruening:
Merci

wm75:
Thanks!
https://github.com/galaxyproject/galaxy/pull/9022 addresses the parsing problem upstream in Galaxy.

Issue #11, opened 2015-01-25
move COPYING file into readme.md file


bgruening:
Can we remove the COPYING into `~` or put them into every readme.md file to get rid of it?
Now that we have one single galaxyp repository this would make much sense, imho.

@jmchilton ok with you? Your are probably the copyright holder?

trevor:
As I understand it the existing tools have been developed under public funding and are therefore in the public domain. The copyright holders are listed in each separate tool in its respective README.

The licenses are assigned on a per-tool basis, instead of under a formally established project-wide charter.

This allows for the GalaxyP project to incorporate other tool contributions that may have more restrictive terms. The hope is that this provides the most flexibility and allows for the greatest developer freedom.

Feel free to bring this up on the mailing list if you'd like to discuss it further so more group members may participate. https://groups.google.com/a/umn.edu/forum/#!categories/galaxyp

bgruening:
Ok, this is fine. Just thought about getting rid of the COPYING file :)

Issue #10, opened 2015-01-24
Integrate the OpenSWATH into Galaxy


bgruening:
More informations here:
 * http://www.openswath.org/example_OpenSWATH.zip
 * http://www.openswath.org/openswath_instructions.html

Programs needed:
------------------------
 * [ ] xinteract (TPP)
 * [ ] spectrast (PP)
 * [ ] spectrast2spectrast_irt.py
 * [ ] spectrast
 * [ ] spectrast2tsv.py
 * [ ] ConvertTSVToTraML
 * [ ] OpenSwathDecoyGenerator

jmchilton:
The TPP tools are so hard to wrap since they create files with so many references embedded right in the file - which is not compatible with the way Galaxy stores files or handles outputs. It might be worth wrapping large portions of the workflow into a single tool for this reason.

PratikDJagtap:
John and Bruening - do you think using PeptideShaker outputs instead of TPP tools would be an alternative?

bgruening:
I hope that @iracooke awesome protk library can help here. Otherwise we can fallback to one single wrapper.
Many of these tools are also integrated in OpenMS, or this is wip. So we do not need to take care of the links at some point.

trevor:
In the list, should that be "spectrast ~~(PP)~~ (TPP)"?

It looks like it's listed twice, is the second one from another source?

bgruening:
@trevor I hope you can also edit the list above. It was listed twice because it is used twice in the workflow. I corrected this mistake. Thanks.

sparker43:
A few additional tools that are pretty integral to the workflow: 
pyprophet https://github.com/uweschmitt/pyprophet and a feature_alignment.py script (written by Hannes et al at ETH) to combine and align results files within a single experiment (https://code.google.com/p/msproteomicstools/source/browse/trunk/analysis/alignment/feature_alignment.py).

bgruening:
Thanks @sparker43 I added it to the list.

bgruening:
added ConvertTSVToTraML and OpenSWATHWorkflow in https://github.com/bgruening/galaxytools/commit/68b57f3370af48f8f0b31c5a4cdad2e6ca38e529

bgruening:
I will start to work on the missing tools here: https://github.com/galaxyproteomics/tools-galaxyp/tree/msproteomicstools
Contributions welcome!

bgruening:
[...]

PR #415, opened 2020-01-12
OpenMS 2.5 .. autogenerated

bernt-matthias:
Important:

- [ ] github actions modifications need to be removed again
- [ ] Commits must be squashed before merge

Related PRs / Issus: 

- CTDConverter & CTDopts https://github.com/WorkflowConversion/CTDopts/pull/21  (and https://github.com/WorkflowConversion/CTDopts/pull/19 - but this one is not necessary) https://github.com/WorkflowConversion/CTDConverter/pull/54 
- Galaxy data types https://github.com/galaxyproject/galaxy/pull/8109, linter errors for `auto_format` only outputs https://github.com/galaxyproject/galaxy/pull/9081
- correct inputs config file https://github.com/galaxyproject/galaxy/pull/9279
- Dependencies metaissue https://github.com/bioconda/bioconda-recipes/issues/18953
- OpenMS improvements to ease future conversions: https://github.com/OpenMS/OpenMS/issues?utf8=%E2%9C%93&q=+author%3Abernt-matthias+

TODOs:

* [x] set tool profile
* [x] remove xml as allowed input format, otherwise to many formats are allowed
  - added mapping to bioml (XTandem output) to filetypes
* [x] add a test (for one tool) where the advanced options are not used (otherwise, if there is a bug with the basic choice it won't be detected)
  * https://github.com/galaxyproject/galaxy/pull/9656
* [x] add ftype to all test outputs?
* [ ] unrestricted text params 
* [x] sections wo title 
  * Looks OK to me. Galaxy just shows the section name.
* [x] implement that output of tools with missing tests
  * planemo l [A-Z]*.xml -s output --report_level warn
* [ ] missing tests : 
  - [ ] ERPairFinder, 
  - [ ] MRMPairFinder
  - [x] OpenSwathRewriteToFeatureXML
  - [x] PhosphoScoring, 
  - [ ] QCEmbedder, 
  - [ ] QCExporter, 
  - [ ] QCExtractor, 
  - [ ] QCImporter, 
  - [ ] SpecLibCreator, 
  - [ ] SvmTheoreticalSpectrumGeneratorTrainer
* [ ] test wo expectations: DTAExtractor, FileInfo, IDRipper, MzMLSplitter, TICCalculator
- [x] tools generating output with delta > 100b
  - FidoAdapter (idXML 5633b), IDFileConverter #15,17,18 (pepXML 2kb), MaRaClusterAdapter #2 (idXML), QCCalculator #1-3(qcML size diff, 300b), SiriusAdapter #7 (idXML, 2k)
  - checked all cases: parameters seem to be correct, diff due to path differences
  - increased delta to 5700b 

Tool specific problems

- [x] IDMapper: there are 2 input parameters `in` and `spectra.in` .. it seems that `format_source="in"` can't cope with this: https://github.com/galaxyproject/galaxy/pull/9493
  - rename one to _in and fix it in fill_ctd https://github.com/galaxyproteomics/tools-galaxyp/pull/415/commits/92682ac3a8e5f908fc5d6263eb30900091ba094e
- MaRaClusterAdapter #1: out missing
- MassCalculator #1,#2: tool insists on txt output
- [x] PepNovoAdapter: model directory
  - per patch: https://gist.github.com/bernt-matthias/703e18cb4c33c276b2f229f1007e958d
- ProteomicsLFQ: in the design file (a table) file names are mentioned which don't match where the mzML files are actually created, e.g. BSA1_F1.mzML (in design) and in/BSA1_F1_mzML.mzML  -> blacklist parameter

Problems with tools that might be solved with composite data types:

Best would be something like this: https://github.com/galaxyproject/galaxy/issues/9503 otherwise one would need special cases for treating composite data types all over the converter. Or follow https://planemo.readthedocs.io/en/latest/writing_how_do_i.html#test-composite-dataset-contents

```
from galaxy.datatypes.registry import Registry

registry = Registry()
registry.load_datatypes(root_dir=GALAXY_ROOT, config=DATATYPES_CONFIG)
```

- cachedMzML consists of a mzML file and a mzML.cached file, I guess this would require a Composite Datatype
  - OpenSwathAnalyzer (Test #10, #11) the .cached file would be needed to be added to the test data
  - OpenSwathMzMLFileCacher #1-7,10 composite data type cachedMzML
  - [ ] blacklist data type/ tool
- [ ] svm models might require Composite Datatype https://github.com/OpenMS/OpenMS/issues/4401
  - RTPredic, PTPredict 
  - InclusionExclusionListCreator (test #3, #4)
- fa.psq:
  - OMSSAAdapter: composite data type needed for fasta.psq, or extra makeblastdb
- spidx
  - SpectraSTSearchAdapter #1-2 composite data type needed .spidx

Some tools output to some prefix that would need to be discovered

- [ ] IdRipper outputs to directory (-out / -out_path)
  see also https://github.com/OpenMS/OpenMS/issues/4404#issuecomment-563133501
- [ ] DTAExtractor `-out <file>*       Base name of DTA output files (RT, m/z and extension are appended)`
- [ ] SeedListGenerator #1-3 needs output by prefix https://github.com/OpenMS/OpenMS/pull/4527   maybe blacklist tool or data type for now

Tools expected to fail here due to missing dependencies: 

- MSFraggerAdapter 
- NovorAdapter
- MaRaCluster

Implement workflows from here: 
- https://www.openms.de/tutorials/


tools/openms/BaselineFilter.xml
```diff
@@ -1,76 +1,83 @@
...
+  <expand macro="stdio"/>
+  <command detect_errors="exit_code"><![CDATA[@QUOTE_FOO@
+@EXT_FOO@
+#import re
+mkdir param_in &&
+ln -s '$param_in' 'param_in/${re.sub("[^\w\-_]", "_", $param_in.element_identifier)}.$gxy2omsext($param_in.ext)' &&
+mkdir param_out_out &&
+@EXECUTABLE@
+-in
+'param_in/${re.sub("[^\w\-_]", "_", $param_in.element_identifier)}.$gxy2omsext($param_in.ext)'
```
bgruening: Oh, you are still constructing the cmd? Did the configfile approach not worked out? For example, those replacements could be avoided in the config file mode I believe. 

I will try to look at this in more detail on Wednesday!

Thanks @bernt-matthias!

bernt-matthias: > Did the configfile approach not worked out? For example, those replacements could be avoided in the config file mode I believe.

Unfortunately the replacements seem inevitable:

- It seems that input files can not be specified in the (ini) config files at all. I just tried it. Other indicators that this is actually the case are 1) the `type="input-file"` tags don't have a value field 2) The tests described in  https://github.com/OpenMS/OpenMS/blob/develop/src/tests/topp/CMakeLists.txt that use ini files also don't set the input files in the ini files but on the command line. To be sure we could still ask the developers.
- The extensions is important for determining the data type of input files in many cases. OpenMS implements sniffers for many but not all data types. For those without sniffers the file extension is used (for those with sniffers and all output files `.tmp` would work as extension). Unfortunately these are the "under specified" data types which will require a thorough review also in https://github.com/galaxyproject/galaxy/pull/8109.
- Also for the proper detection of the Galaxy data types of the outputs the extension is used in many cases.

bernt-matthias:
LOL tests exceeded max runtime ... I try to increase the number of chunks by 1 .. lets see if this can cope with the approx 400 tests

tools/openms/BaselineFilter.xml
bernt-matthias: With https://github.com/WorkflowConversion/CTDConverter/pull/54/commits/bdf37bf4ab3827c592e44171a8e4b46482f3075c we are on a good way to CTD based input :) 

Still some bits to fix (in particular the test generation).

bgruening: Just awesome!!!

bgruening:
It seems that a few test files are missing.

* database/proteins_fasta.fa.psq
* MetaboliteSpectralDB.mzML
* home/berntm/projects/tools-galaxyp/tools/openms/OpenMS2.5.0-git/THIRDPARTY/Linux/64bit/MaRaCluster/maracluster

bernt-matthias:
> database/proteins_fasta.fa.psq

Jep. If a fasta file is given to OMSSAAdapter it just expects a psq file in addition (which would need to be created with makebalsdb). This is one of the tools where files are expected that can't be specified by an argument. Could be solved via a composite data type or in this case by an extra program call.

> MetaboliteSpectralDB.mzML

This was just to big (>100 MB) so I excluded it for now. Somehow I think this could be a good idea here https://github.com/galaxyproject/tools-iuc/issues/2906

> /home/berntm/projects/tools-galaxyp/tools/openms/OpenMS2.5.0-git/THIRDPARTY/Linux/64bit/MaRaCluster/maracluster

Yep. This does not have a conda package yet (could not complie it https://github.com/bioconda/bioconda-recipes/pull/19219). Maybe we would start with a conda package of the binary?

bgruening:
OMSSA is dead afaik. So maybe we should skip it.

@bernt-matthias how do you want to proceed here? Is utterly amazing that so many tests are passing!

bernt-matthias:
Hi @bgruening. 

I will post a list of tools tomorrow that still fail together with the reason, then we can decide what to do with each of the tools. 

As soon as everything is green:

-  we should somehow get this merged https://github.com/WorkflowConversion/CTDopts/pull/21 and into conda (and maybe also the CTDconverter PR). Do you know somebody who could review this?
- then I would suggest to implement something like this https://github.com/galaxyproject/tools-iuc/pull/2908 also for this repo in order to get a clean and small PR .. 
- also we need to merge the datatypes and maybe backport (and set the profile of the tools accordingly)

bgruening:
>     * we should somehow get this merged [WorkflowConversion/CTDopts#21](https://github.com/WorkflowConversion/CTDopts/pull/21) and into conda (and maybe also the CTDconverter PR). Do you know somebody who could review this?

I ask Timo, let's see. For the 2.5 release, I would have no problem to ship CTDopts with the wrapper.

>     * then I would suggest to implement something like this [galaxyproject/tools-iuc#2908](https://github.com/galaxyproject/tools-iuc/pull/2908) also for this repo in order to get a clean and small PR ..

Maybe we should do this here at first and test it ...

>     * also we need to merge the datatypes and maybe backport (and set the profile of the tools accordingly)

Ok, I will look at it again. Backport might be tricky, but I guess we can backport it in our instances and add a hint to the readme.

bernt-matthias:
Another update. Locally I'm now at 25 failing tests. I updated the list of TODOs on the top of the PR.

Will continue with porting https://github.com/galaxyproject/tools-iuc/pull/2908

bernt-matthias:
Hi @bgruening test-data is now excluded from the PR and auto generated before the planemo test. 

Still most of the test chunks time out. Do you see a way to fix this?

bgruening:
> Still most of the test chunks time out. Do you see a way to fix this?

Only by migrating to github actions. But travis is a dead end here, I assume :(

bernt-matthias:
Is there any downside to github actions? Other wise this could be a starting point: https://github.com/galaxyproject/tools-iuc/blob/master/.github/workflows/ci.yaml

bgruening:
No downside for github actions, it just needs to be done :(

bgruening:
@bernt-matthias do you think you can build a real action: https://help.github.com/en/actions/building-actions

So we can import this into any other repo?

bernt-matthias:
Maybe. Would be docker based .. I guess. Lets see if I find time. Need to read a bit.

I think porting the ci workflow to here would also be nice.

tools/openms/OpenSwathWorkflow.xml
```diff
@@ -1,598 +1,480 @@
...
+        <param name="use_mi_score" argument="-Scoring:Scores:use_mi_score" type="boolean" optional="true" truevalue="true" falsevalue="false" checked="false" label="Use the MI (mutual information) score" help=""/>
+        <param name="use_dia_scores" argument="-Scoring:Scores:use_dia_scores" type="boolean" optional="true" truevalue="true" falsevalue="false" checked="true" label="Use the DIA (SWATH) scores" help="If turned off, will not use fragment ion spectra for scoring"/>
+        <param name="use_ms1_correlation" argument="-Scoring:Scores:use_ms1_correlation" type="boolean" optional="true" truevalue="true" falsevalue="false" checked="false" label="Use the correlation scores with the MS1 elution profiles" help=""/>
+        <param name="use_sonar_scores" argument="-Scoring:Scores:use_sonar_scores" type="boolean" optional="true" truevalue="true" falsevalue="false" checked="false" label="Use the scores for SONAR scans (scanning swath)" help=""/>
+        <param name="use_ion_mobility_scores" argument="-Scoring:Scores:use_ion_mobility_scores" type="boolean" optional="true" truevalue="true" falsevalue="false" checked="false" label="Use the scores for Ion Mobility scans" help=""/>
```
bgruening, line 348: @bernt-matthias boolean should not be optional I think.

tools/openms/OpenSwathWorkflow.xml
```diff
@@ -1,598 +1,480 @@
...
+        <option value="ppm">ppm</option>
+        <expand macro="list_string_san"/>
+      </param>
+      <param name="mz_extraction_window_ms1_unit" argument="-mz_extraction_window_ms1_unit" display="radio" type="select" optional="false" label="Unit of the MS1 m/z extraction window" help="">
+        <option value="ppm">ppm</option>
+        <option value="Th" selected="true">Th</option>
+        <expand macro="list_string_san"/>
+      </param>
+      <param name="use_ms1_ion_mobility" argument="-use_ms1_ion_mobility" type="boolean" optional="true" truevalue="true" falsevalue="false" checked="true" label="Also perform precursor extraction using the same ion mobility window as for fragment ion extraction" help=""/>
+      <param name="matching_window_only" argument="-matching_window_only" type="boolean" optional="true" truevalue="true" falsevalue="false" checked="false" label="Assume the input data is targeted / PRM-like data with potentially overlapping DIA windows" help="Will only attempt to extract each assay from the *best* matching DIA window (instead of all matching windows)"/>
```
bgruening, line 379: `matching_window_only` and `use_ms1_ion_mobility` do not seems to be present in the command section.

bernt-matthias: Indeed you will find none of the parameters (except those for input/output files) in the command, but they find their way directly to the ini file via the json-config file. The magic in here and the following lines:

https://github.com/galaxyproteomics/tools-galaxyp/blob/2a2f239afd08a49d08d09e0d47bd57768777f24c/tools/openms/OpenSwathWorkflow.xml#L66

the agruments as json are due to this:

https://github.com/galaxyproteomics/tools-galaxyp/blob/2a2f239afd08a49d08d09e0d47bd57768777f24c/tools/openms/OpenSwathWorkflow.xml#L148

1. the binary (in this case OpenSwathWorkflow) is used to write a ctd file (with defaults).
2. then `fill_ctd.py` takes the parameters of the tool as json and fills them into the ctd file.
3, the binary is executed with the filled ctd file.

tools/openms/OpenSwathWorkflow.xml
bernt-matthias: True.

bgruening:
@bernt-matthias what do you think about adding the CDT file to all tools as a) optional output or b) to the stdout.

This could also make a pretty good test.

bernt-matthias:
> what do you think about adding the CDT file to all tools as a) optional output or b) to the stdout.

+1 seems easy and useful. would go for a)

> This could also make a pretty good test.

you suggest to add assertions on the content of this output in the tests which check if all parameters are set as desired?

bgruening:
> you suggest to add assertions on the content of this output in the tests which check if all parameters are set as desired?

Whatever is easiest. There is also a valid_xml assert or something like that.

bgruening:
Wow, grün! Und nun? :)

bernt-matthias:
Struggling to get tests running. For the last tests some chunks got stuck at some point...

bernt-matthias:
> Wow, grün! Und nun? :)

I think the next step might be the data types https://github.com/galaxyproject/galaxy/pull/8109?

bgruening:
@bernt-matthias I also found this here:

```
    <param name="threads" argument="-threads" type="hidden" value="${GALAXY_SLOTS:-1}">
      <sanitizer>
        <valid initial="string.printable"/>
      </sanitizer>
    </param>
    <section name="Debugging" title="Debugging" help="" expanded="false">
```

The slots should not be exposed to the user and imho also not the debug or verbose options.

bernt-matthias:
> The slots should not be exposed to the user and imho also not the debug or verbose options.

True. Probably something wrong with the hardcoding.

bernt-matthias:
> The slots should not be exposed to the user and imho also not the debug or verbose options.

Actually they are not really exposed, but `type="hidden"`. Alternatively there would be the possibility to add those via `fill_ctd.py` ... ?

bgruening:
Ok, I see .. have you checked that `${GALAXY_SLOTS:-1}` really works this way? It needs to be on bash, as this is bash syntax that needs to be evaluated.

bernt-matthias:
This is actually covered by https://github.com/galaxyproteomics/tools-galaxyp/blob/179567305f7fbb7f8c29146d40d7c05d06e1700c/tools/openms/fill_ctd.py#L45

So, if there is a text parameter value starting with `$` and consisting of upper case letters its looked up in the environment with a default of `""` (also `${VAR:-DEFAULT}` is covered. I don't really like this, because it clearly fails if users enter something starting with `$`. Maybe injecting hardcoded parameters into the json would be better...

bgruening:
@bernt-matthias I found a wired error.

It seems in the json is the `test` parameter, which is hidden, set to false: `"test": "false"`. That is good.
Unfortunately, the CTD sets it to `true`. 
```
        <ITEM name="test" value="true" type="bool" description="Enables the test mode (needed for internal use only)" required="false" advanced="true"/>
```

bernt-matthias:
> It seems in the json is the `test` parameter, which is hidden, set to false: `"test": "false"`. That is good.

I guess this is for all tools? Or a specific one.

bgruening:
I assume that is for all. I have seen it in the OpenSwathWorkflow tool.

bernt-matthias:
I remember that In the tests the value of test is always set to true, but I guess you are not referring to a test...

bgruening:
No, what I'm saying is that test seems to be set always to "true" in a real world run. And this should not happen.

bernt-matthias:
Main reason seemed to be that `hardcoded_params.json` used `"false"` instead of `false` and therefore the parameter was not updated in the CTD file (I will see if I can add a warning or error in such a case .. this should not be silent).

More important: I implemented that hardcoded parameters (like test) are specified via another config file. This adds the additional security that variable lookup in the environment (and execution of commands with `$(command)`) only happens for those parameters (and not for stuff users enter in the tool form). 

Problem now is that I can't change test to true in the tests anymore .. which is important for the tests.  Lets see if I can change this.

bgruening:
> Main reason seemed to be that hardcoded_params.json used "false" instead of false and therefore the parameter was not updated in the CTD file (I will see if I can add a warning or error in such a case .. this should not be silent).

I think this is expected, as the hidden-parameter is __not__ a boolean. I guess you need to special-case hidden parameter and in this case cast it to boolean.

bernt-matthias:
Nearly there :) .. just discovered another "problem". All tests use the advanced options since I (need to) enable the test option. So bugs caused by not using the advanced section are not covered yet. 

I guess adding one test for one tool that does not use the advanced section should be sufficient.

bernt-matthias:
Hey @bgruening I'm a bit lost with the PeptideIndexer error. For some reason the JSON parameter config file contains: `{"decoy_string": "{'src': 'hda', 'id': '3b3b4832b2563c85'}", ...` but it should be empty text .. 

Even more weird is that the error does not occur if I test only this tool / in local tests.

Any ideas? Will try to increase the number of chunks. Maybe some weird limitation (ulimit?) causing a bug in galaxy / planemo .. just wild guessing .. I really have no idea.

bgruening:
Which Galaxy version are you using locally? I have pinned here 20.05, so we can test against the next stable release. What I don't understand is how a text parameter can become a `hda` - which looks like an input file.

bernt-matthias:
I've seen this hda problem before when parameter naming colissions .. but I could not find one here.

bernt-matthias:
Good idea to switch to 20.05 .. even if there should not be many differences to dev at the moment.

bgruening:
@bernt-matthias what is your plan? We have a few tools that we needed added to EU and I have a few manual changes aggregated already.

bernt-matthias:
Plan is to get this "finished" ;) have 1 week left before vacation. Let's use this as a deadline. 

I will check the test results these days and compile a todo list. Maybe we can have a talk these days? 

Do you have your manual changes somewhere? Still aiming for automatisation.. If possible.

bernt-matthias:
First thing needed would be to merge and publish https://github.com/WorkflowConversion/CTDopts/pull/24

bgruening:
> Maybe we can have a talk these days?

Yes, please :)

Here are my changes: https://github.com/usegalaxy-eu/temporary-tools/tree/master/test/OpenMS25

bernt-matthias:
> Here are my changes: https://github.com/usegalaxy-eu/temporary-tools/tree/master/test/OpenMS25

I only found significant changes in OpenSwathWorkflow:

- `out_*_FLAG` removed
- `in` is multiple=false in your version. Is this intended? Could be covered easily by the converter (hardcoded_params.json).

bgruening:
Yes, both changes are intended based on user-feedback.

bernt-matthias:
> Yes, both changes are intended based on user-feedback.

For the input parameter this is easy: https://github.com/galaxyproteomics/tools-galaxyp/blob/77768ccd3f2b8535b12780e7abdddddc4abf8c1a/tools/openms/hardcoded_params.json#L203

For the output selectors, is it preferable to have a single select instead of multiple bools in general?

bernt-matthias:
I expect the the current current test fails only for tools without binaries (MSFraggerAdapter MaRaClusterAdapter NovorAdapter SpectraSTSearchAdapter). 

Plan for tomorrow: 

- [ ] disable tests for those tools
- [ ] enable containerized testing 
- [ ] remove the PYTHONPATH and the local copy of CTDOpts. 
- [ ] if this fails I will add them to `.tt_biocontainer_skip` 
- [ ] CI workflow needs to be adapted for tests data generation

Then we only need the CTDopts release (on the repo) and the bioconda package bump.

Lets see if this plan works out (by definition plans don't work out usually) ... fingers crossed

bernt-matthias:
Hi @bgruening 

Seem that https://github.com/galaxyproteomics/tools-galaxyp/commit/9dcd8ec444e5073cce7ac10f67b2248e28f00e4b is working .. see here https://github.com/galaxyproteomics/tools-galaxyp/actions/runs/197816225. 

So currently no containerized testing. The linter errors are due to wrong urls in the generated help .. I have reported this, but I guess it can be ignored. 

I will revert to this commit and remove the local copy of CTD opts. All that is the left is to wait for the new release of CTDopts (v1.4) and bump the conda recipe. 

I hope that all this can be done without me... 

Everything else needs to wait :)

bgruening:
Awesome! I will then take over, now that nothing is to do anymore! :)

Thanks a bunch!

bernt-matthias:
Is there anything that needs to be done for the newly blacklisted and revived tools? Check the diff in tools_blacklist.txt

bgruening:
No, the list is really just for us to not generate CDT and then Galaxy. Nothing special.

bernt-matthias:
@bgruening `All 583 test(s) successfully executed.` but only in un-containerized testing. Linting fails because of wrong URLs -- which won't fix for 2.5.

I will spend a bit of time to debug the stalling containerized tests. If I'm not successful we can merge https://github.com/galaxyproteomics/tools-galaxyp/pull/415/commits/5844a485d5589d2682d7974f1156fcd1ced18452

bgruening:
Awesome! Thanks! Do we wait for a CTDopt package?

bernt-matthias:
> Do we wait for a CTDopt package?

Already done: https://github.com/conda-forge/ctdopts-feedstock/pull/3

I will just try to debug the containerized testing problems. If I can't solve it we merge as is. I will let you know.

bgruening:
Do you see the problem locally when you run `planemo test --docker foo_bar.xml`?

bernt-matthias:
@bgruening I give up .. can't find the problem. Feel free to merge or request further changes.

bgruening:
Boah, wow. Sorry for all your pain. What a herculean work!

bernt-matthias:
Deployment on the TS seems to fail: 

```
"Metadata may have been defined for some items in revision '90f3670ec66b'.  Correct the following problems if necessary and reset metadata.<br/><b>FeatureFinderMRM.xml<\/b> - The tool FeatureFinderMRM targets version 20.05 of Galaxy, you should upgrade Galaxy to ensure proper functioning of this tool.<br/>"
```

PR #419, opened 2020-02-20
Fix typo in peptideshaker

bernt-matthias:
Leads to the following otherwise

```
Invalid Syntax
Line 129, column 20

Line|Cheetah Code
----|-------------------------------------------------------------
126 |            #end if
127 |            #if '2' in $output_reports_list:
128 |                && find \$cwd/output_reports -name '*Default_PSM_Phosphorylation_Report.txt' -exec bash -c 'mv "$0" "psm_phospho.txt"' {} \;
129 |            #end if`
                        ^
130 |            #if '3' in $output_reports_list:
131 |                && find \$cwd/output_reports -name '*Default_PSM_Report.txt' -exec bash -c 'mv "$0" "psm.txt"' {} \;
132 |            #end if
```

### DIFF
--- a/tools/peptideshaker/peptide_shaker.xml
+++ b/tools/peptideshaker/peptide_shaker.xml
@@ -30,7 +30,7 @@
 
 
         peptide-shaker eu.isas.peptideshaker.cmd.PathSettingsCLI
-	         --exec_dir="\$cwd/${bin_dir}"
+            --exec_dir="\$cwd/${bin_dir}"
             -temp_folder \$cwd/PathSettingsCLI
             -log \$cwd/peptideshaker.log &&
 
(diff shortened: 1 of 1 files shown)


tools/peptideshaker/peptide_shaker.xml
```diff
@@ -135,7 +135,7 @@
             && unzip \$cwd/peptideshaker_output.zip 'reports/*' -d \$cwd/output_reports
         #end if
 
-	      #if len(output_reports_list)>0:
+        #if len(output_reports_list)>0:
```
bgruening: mh, this indentation is off and do we miss a $ here?

bernt-matthias: As far as I can seen in the github editor indentation is now correct. 

good point with the `$` .. wonder why this works also without.

bgruening:
Merci!

PR #423, opened 2020-03-25
Updating some cardinal tools

foellmelanie:
Updated preprocessind and filtering to 2.4, added options to read processed files in combine


bgruening:
Wonderful work, thanks @foellmelanie!

PR #424, opened 2020-03-26
Updated MSI mz images to new Cardinal version

foellmelanie:
lattice option not needed anymore and removed
new colours
added normalization
options for legend position in overlay image


foellmelanie:
* preprocessing: removed peaks function, added peakBin instead; added variable to define CPUs
* data exporter, filtering, mz images, spectra plot, qualitycontrol: everything updated to 2.4
* segmentation, classification, combine: only read function updated to 2.4., internal code still uses old R objects

PR #427, opened 2020-03-29
hardklor: add missing python requirement

bernt-matthias:
- fix cheetah quoting
- fix python code style


tools/hardklor/kronik.xml
```diff
@@ -6,7 +6,7 @@
     <stdio>
         <exit_code range="1:" level="fatal" description="Error occurred while running Kronik"/>
```
bgruening: bump version number?

bgruening:
Mh, this now failed https://github.com/galaxyproteomics/tools-galaxyp/runs/565935316

PR #429, opened 2020-03-30
regex: add missing py requirement

bernt-matthias:
and remove unused and deprecated import


bernt-matthias:
@bgruening if you rerun the failing test it should be fine. 

Note each tests appears 2x since I closed & reopened.

bgruening:
What I find strange is that we can not restart single tasks. If deployment get stuck we can not just restart that one, we need to rerun all tests as well.

PR #430, opened 2020-03-31
flashlfq: detect mzML reading error

bernt-matthias:
In the containerized test the logfile contains

```
Problem opening .mzML file ./spectrum_dir/sliced-mzml.mzML; One or more errors occurred.
```

which is now at least detected. Problem still persists. Odd thing is that some output files are still produced... maybe non-fatal error?


bgruening:
bump ...

PR #437, opened 2020-04-07
hardklor: more flake

bernt-matthias:
follow up to https://github.com/galaxyproteomics/tools-galaxyp/pull/427

### DIFF
--- a/tools/hardklor/hardklor.py
+++ b/tools/hardklor/hardklor.py
@@ -13,7 +13,7 @@ extension_lookup = {
     'mzml': 'mzML',
     'mzxml': 'mzXML',
     'ms1': 'MS1',
-    }
+}
 
 # input:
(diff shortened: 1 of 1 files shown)


tools/hardklor/hardklor.py
```diff
@@ -13,7 +13,7 @@
     'mzml': 'mzML',
     'mzxml': 'mzXML',
     'ms1': 'MS1',
-    }
+}
```
bgruening, line 16: I don't understand this diff ...
I fixed that already in master: https://github.com/galaxyproteomics/tools-galaxyp/blob/master/tools/hardklor/hardklor.py

Why is github showing this diff?

bernt-matthias: Me neither.  Lets just close it,

PR #439, opened 2020-04-10
new versions of maxquant and maquant_mqparam

dglaetzer:
* updated the maxquant version to 1.6.10.43
* reorganised python scripts for better maintainability
* added features:
  * ptxqc report (probably unstable)
  * parameter groups
  * reporter ion ms2
  * a few other new parameters


bgruening:
Two tests are failing, at least one with:

```
|  tar: invalid option -- 'z'
|  BusyBox v1.22.1 (2014-05-23 01:24:27 UTC) multi-call binary.
|  Usage: tar -[cxthvO] [-X FILE] [-T FILE] [-f TARFILE] [-C DIR] [FILE]...
|  Create, extract, or list files from a tar file
|  Operation:
|  c	Create
|  x	Extract
|  t	List
|  f	Name of TARFILE ('-' for stdin/out)
|  C	Change to DIR before operation
|  v	Verbose
|  O	Extract to stdout
|  h	Follow symlinks
|  exclude	File to exclude
|  X	File with names to exclude
|  T	File with names to include

bgruening:
@dglaetzer just for your information in the case you ever come back :)  For XML files you can test with `is_valid_xml`, `has_n_elements_with_path`, `attribute_is` ...and many more from here: https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-tests-test-output-assert-contents

bgruening:
And green! Thanks a lot @dglaetzer! Happy Easter.
I will give @foellmelanie a chance to review and then merge if she agrees ...

foellmelanie:
Thank you @dglaetzer it looks great :)

dglaetzer:
Thank you both for reviewing and fixing the issues. A potentially unresolved one is a problem with `maxquant_mqpar` on my own galaxy instance. It (still) appears directly after hitting the "Execute"-button (see attached files). When using the tool via `planemo serve` it works fine however. 
galaxy_error

bgruening:
Hopefully, just a local Galaxy problem. If it works on planemo that is good for me. We will see how this works on EU :)

PR #445, opened 2020-04-27
Using full file locations as arguments to avoid relative-paths ...

CarlosHorro:
problems with some search engines using mzml files.

We have experienced with MetaMorpheus search engine (and it may happen with other search engines) that, when loading mzml files created with ThermoRawFileParser galaxy tool, the search engine crashes trying to load from the .mzml this part:

`

      ...
      <sourceFile id="RAW1" name="input" location="./raws_folder/input.raw">
        ...
      </sourceFile>
      ...
`

as it expects in "location" attribute an absolute path; instead, we were writing there a relative path and, even when "./raws_folder/input.raw" is not used at all out of galaxy, just processing the sourceFile element causes this problem.

This modification writes the absolute path of the source raw file to the .mzl, avoiding this problem. 
By the way, I think this is not as misleading as the relative-path method, as it will be obvius when checking the .mzml that the file location attribute will make reference to a galaxy internal folder, whilst having a relative path there may suggest we should have in our computer some "raws_folder/input.raw" file.

Greetings,
Carlos

### DIFF
--- a/tools/ThermoRawFileParser/thermo_converter.xml
+++ b/tools/ThermoRawFileParser/thermo_converter.xml
@@ -7,6 +7,7 @@
 <![CDATA[
 #import re
 
+cwd=`pwd` &&
 mkdir ./raws_folder &&
 mkdir ./output_folder &&
 #for $input_raw in $input:
@@ -19,8 +20,8 @@ mkdir ./output_folder &&
 #end for
 
 ThermoRawFileParser.sh
-    -d=./raws_folder
-    -o=./output_folder
+    -d="\$cwd/raws_folder"
+    -o="\$cwd/output_folder"
     -f=$output_format
     #if $output_metadata_selector != "off":
         --metadata="${output_metadata_selector}"


bgruening:
Welcome back @CarlosHorro!

> just processing the sourceFile element causes this problem.

Why is that? Just because it's relative? If you add a non-existing absolute path it is working?
Do you know what the spec is suggesting?

I'm skeptical if this change, influences upstream tools. The idea was to put relative paths and if a tool needs to RAW file, the tool is liking/copying the RAW file next to the mzml file.

The absolute path is relly not trustable and will mostly never exists. With the relative path, we at least have a chance to make it work.

CarlosHorro:
> Why is that? Just because it's relative? If you add a non-existing absolute path it is working?
> Do you know what the spec is suggesting?

Yep, I've done several tests and a non-existing absolute path works fine as the search engine doesn't actually need to read the source raw file.


The specification at
http://www.peptideatlas.org/tmp/mzML1.1.0.html

CarlosHorro:
Great to see you are still as active as always @bgruening  :-)

> Why is that? Just because it's relative? If you add a non-existing absolute path it is working?
> Do you know what the spec is suggesting?

Yep, I've done several tests and a non-existing absolute path works fine as the search engine doesn't actually need to read the source raw file. The error shown is:

`Exception: Invalid URI: The format of the URI could not be determined. 
`
The specification at
http://www.peptideatlas.org/tmp/mzML1.1.0.html

defines "location" as 

> URI-formatted location where the file was retrieved

with type 

> xs:anyURI


So that is not completely conclusive. Checking some information about the xs:anyURI type:
http://books.xmlschemata.org/relaxng/ch19-77009.html

The "restrictions" topic seems to talk against using relative URIs... but not sure about if I'm interpreting it correctly...

About what you say here:
> The absolute path is relly not trustable and will mostly never exists. With the relative path, we at least have a chance to make it work.

I think neither absolute nor relative paths coming from Galaxy would be reliable... relative paths would work just by chance, which I think is not recommendable to trust on. Once the .mzml is downloaded, the location path would probably have to be changed to adapt to every individual file structure...

bgruening:
> Great to see you are still as active as always @bgruening :-)

Hehe, always!

>   <sourceFile id="RAW1" name="input" location="./raws_folder/input.raw">

Could you please try if this works: `file:///./raws_folder/input.raw`

>  relative paths would work just by chance, which I think is not recommendable to trust on. 

true, but at least we could easily download everything in one folder and then it should work without changing files. Who had the idea to put the path into the XML file ... damn.

CarlosHorro:
> Could you please try if this works: `file:///./raws_folder/input.raw`
> 

Great point, it works when loading by the search engine :-)
Although, when passing "file:///./raws_folder" as an argument to the app via -d, the argument is not properly recognised as a folder, so I think I will have to keep the galaxy tool as it was and modify the app itself introducing the "file:///" stuff when writing the .mzml. 

I will look into that now...

Greetings,
Carlos

PR #460, opened 2020-06-05
Pyprophet export update

foellmelanie:
updated help and updated table in pdf, fixed one typo

### DIFF
--- a/tools/pyprophet/pyprophet_export.xml
+++ b/tools/pyprophet/pyprophet_export.xml
@@ -1,4 +1,4 @@
-<tool id="pyprophet_export" name="PyProphet export" version="@VERSION@.0">
+<tool id="pyprophet_export" name="PyProphet export" version="@VERSION@.1">
     <description>
     Export tabular files, optional swath2stats export
     </description>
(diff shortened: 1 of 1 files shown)


tools/pyprophet/pyprophet_export.xml
```diff
@@ -88,12 +88,32 @@ data.annotated <- sample_annotation(data_me, study_design, column.file = "filena
...
+number_samples = nrow(count_analytes(data.annotated.nodecoy))
+
+    ### for more than 20 annotation groups print only 20 samples per page:
+    if (number_samples <= 20){
+        grid.table(count_analytes(data.annotated.nodecoy), rows= NULL)
+    }else{
+        grid.table(count_analytes(data.annotated.nodecoy)[1:20,], rows= NULL)
+        mincount = 21
+        maxcount = 40
+        for (count15 in 1:(ceiling(number_samples/20)-1)){
```
bgruening, line 106: Meli it is better to add a test for this if at all possible, otherwise such a logic can miss pieces.

foellmelanie: unfortunately no chance to use a test file with > 20 samples

Issue #461, opened 2020-06-11
Percolator 3.1 vs 3.4


Slugger70:
Hi there, we are having issues installing percolator 3.1 at the moment as the conda package will not install. Conda for Percolator 3.4 installs fine though. Are there any plans to update the percolator tool to 3.4?

bgruening:
@Slugger70 no plans afaik, but please try your luck bumping the version.

bernt-matthias:
Seems solved

PR #466, opened 2020-06-23
Update percolator to version 3.5

cat-bro:
Updated percolator to 3.4 and updated test data with planemo.

Also populated the test_data/empty_file*.mzid files for nested_collections.xml tests, as the tests would not run if the uploaded test files were empty.

This addresses https://github.com/galaxyproteomics/tools-galaxyp/issues/461 - we would like to have percolator on Galaxy Australia but version 3.1 is no longer installable with conda.


bernt-matthias:
Percolator 3.5. seems to be available. May you are lucky and bumping the bioconda recipe just works :)

cat-bro:
> Percolator 3.5. seems to be available. May you are lucky and bumping the bioconda recipe just works :)

There is an open pull request in bioconda-recipes for this:  https://github.com/bioconda/bioconda-recipes/pull/23107.

bgruening:
Let's see if we are lucky :)

bernt-matthias:
Seems that allowing 4 lines difference for the percolator output (allowing for version and filename differences) would fix the remaining problems.

cat-bro:
@bernt-matthias I wish this were the case.  For the percolator test that uses all of the option flags, the difference between the expected data and the output is extreme.  I've committed it here but I don't know if this should be merged :/

The unix diff between expected and actual output is 36K lines (for a 22K line file).  I'm not sure if the sorting is important for these output files but even when sorted by psm_id, the diff is about 10K lines.

It looks like all of the options used in the command are still valid in this version of percolator.  Not sure what to make of the output.

bernt-matthias:
Maybe the check for equality is sufficient for one test. The other tests (e.g. the one "that uses all of the option flags") could just check for a specific line (or just a part of it) in the output via an `assert_content` tag.

cat-bro:
@bernt-matthias I don't think the test needs to change.  Since this is the output of percolator v3.5 and it is passing its tests with the updated test-data, it is probably ready.

bgruening:
Thanks @cat-bro!

PR #468, opened 2020-06-26
Mt2mq rowname fix

mariecrane:
Fixed bug of unwanted rownames in output by setting row.names = FALSE in write.table.


bgruening:
Thanks.

PR #471, opened 2020-06-28
sync ci workflows with IUC

bernt-matthias:
- R linting
- updated actions
- planemo test uses --no_dependency_resolution

is ` github.repository_owner == 'galaxyproteomics'` correct?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,14 +1,14 @@
-name: Run Galaxy Tool Linting and Tests
+name: Weekly global Tool Linting and Tests
 on:
   schedule:
     # Run at midnight every monday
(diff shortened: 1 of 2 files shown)


bgruening:
@bernt-matthias is it possible to create an own action based on this? So one single spot that we need to maintain?

bernt-matthias:
Would be nice .. not sure if I can spend time on this now. An alternative might be a git submodule, or?

bernt-matthias:
Also: if we would have an action then all repos still would need to have workflows and update the action's version ...

bgruening:
submodules are usually hard to handle, I think actions and bumping versions are the way to go. Maybe you can open an issue on IUC so we can keep track. Thanks Matthias!

PR #472, opened 2020-06-29
Allow only mgf scans, update latest version of jar

tmcgowan:
Update: mztosqlite executor.awaitTermination flow modified to better handle large data sets processed in low resource environments. Especially when run with only 1 thread. This updates to revised recipe version


bgruening:
> |  /tmp/tmp_82jwx2i/job_working_directory/000/4/tool_script.sh: line 9: mz_to_sqlite: command not found

You are too fast @tmcgowan - amazing. Conda package is not yet ready. We need to restart this one in ~20min.

tmcgowan:
@bgruening 
Yes, I wanted to get this done before I started some other work this morning. I'll get back to this in the late afternoon.

PR #483, opened 2020-08-07
Modify fastg2protlib to suite tool

tmcgowan:
Refactored fastg2protlib as suite of two tools.


bgruening:
@tmcgowan this is really not needed. All you need to do is create a shed.yml file similar to https://github.com/galaxyproteomics/tools-galaxyp/blob/master/tools/cardinal/.shed.yml

Documentation is here: https://planemo.readthedocs.io/en/latest/standards/docs/best_practices/shed_yml.html#advanced-parameters

bgruening:
That is fixed already.

PR #496, opened 2020-09-20
Update cardinal tests so that they can pass in parallel with ephemeris

cat-bro:
These tools pass all of their tests one by one but fail shed-tools parallel testing because a lot of the test inputs have the same value attribute as each other ("") but are different datasets.


bernt-matthias:
This should not be changed, because its actually a Galaxy bug (recently discovered by @mvdbeek: see https://github.com/galaxyproject/galaxy/pull/9885#issuecomment-692818641)

I guess its better to wait for the bug fix...?

cat-bro:
Thanks @bernt-matthias.  I had been almost 50/50 about whether this needed to be fixed here or in Galaxy.

PR #497, opened 2020-09-20
MSI tools: updating to cardinal 2.6

foellmelanie:
resolution (accuracy) argument has changed from half window to full window size, therefore the test for processed files needed doubling the original value to pass


bgruening:
mz_images failed :(

 https://github.com/galaxyproteomics/tools-galaxyp/pull/497/checks?check_run_id=1141515799#step:14:680

Should I restart the tests?

Issue #495, opened 2020-09-19
openms test-parameter


bgruening:
@bernt-matthias I see this parameter in the generated CTD:

>         <ITEM name="test" value="true" type="bool" description="Enables the test mode (needed for internal use only)" required="false" advanced="true"/>

However, it should be __false__ no matter what in real life. The code is here and looks good:

https://github.com/galaxyproteomics/tools-galaxyp/blob/master/tools/openms/OpenSwathWorkflow.xml#L399

Any idea?

bgruening:
Removing this parameter alltogether solves it:

bgruening:
I also found this:

>     <section name="Debugging" title="Debugging" help="" expanded="false"/>

I don't think empty sections should be included.

bernt-matthias:
Empty sections should disappear with https://github.com/WorkflowConversion/CTDConverter/pull/54/commits/346341a19c5b40c61adcfcd35037056698690f1b

bernt-matthias:
Got it (I think). The test parameter is a ("true") boolean parameter in ctd (i.e. one that is not a string with restrictions "true", "false"). In the Galaxy xml  the parameter is rendered as text which is transformed via `bool(...)` which is always `true` :( 

Nasty bug with an easy fix. PR should follow soon.

bgruening:
Would something like that help? https://github.com/galaxyproject/galaxy/blob/2777db0e46f1185e47c41049abdb61f99ac1e2e7/lib/galaxy/util/__init__.py#L939

bernt-matthias:
Actually easier, since we know that there is only "true" / "false" (https://github.com/galaxyproteomics/tools-galaxyp/blob/1d1ee678c35bc262505fde58f50998f7d6586d66/tools/openms/fill_ctd.py#L185) .. started a PR

PR #498, opened 2020-09-21
Topic/openms2.5 fix bool

bernt-matthias:
fixes https://github.com/galaxyproteomics/tools-galaxyp/issues/495

TODO

- [x] check discarded tool tests
- [x] remove 404 links from help (maybe with patch)


bernt-matthias:
I will have a look why the order of the selects changes all the time .. this may be a bit annoying for users.

bgruening:
Looks like there is a dict() involved?

bernt-matthias:
@bgruening .. ready from my side. 

if you like we can change the number of chunks to 10 again for the PR workflow

bgruening:
Is the select order now stable?

tools/openms/macros.xml
```diff
@@ -4,7 +4,7 @@
      add additional macro files using the m/macros parameter -->
 <macros>
   <token name="@TOOL_VERSION@">2.5</token>
-  <token name="@GALAXY_VERSION@">0</token>
+  <token name="@GALAXY_VERSION@">1</token>
```
bgruening, line 7: do we want to increase it or patch the older version?

bernt-matthias: whatever you think is better.

bernt-matthias:
> Is the select order now stable?

I did not find an obvious reason for the change of the order .. unfortunately I do not have time at the moment to work on this...

bgruening:
Merci!

PR #499, opened 2020-09-23
Update metaquantome

jj-umn:
Praveen @pravs3683 modified metaquantome for multiple samples:

stat module can input multiple samples (would require the user to input the control sample name and it will be used as the denominator)
heatmap viz would also generate a tabular file sharing the cluster data
pca viz would also generate the tabular file sharing pc1 and pc2 data
heatmap requires the column name of the corrected p-value when using --filter-to-sig


bgruening (approved):

bgruening:
Thanks @jj-umn!

PR #505, opened 2020-10-19
Update Bumbershoot to 3.0.20290

chambm:


chambm:
@bgruening Where is the conda log? I'm not sure why these commands aren't being found. The tests worked in bioconda.

bgruening:
@chambm have a look under "Actions" here on this page. It will bring you to https://github.com/galaxyproteomics/tools-galaxyp/actions/runs/316310819   than search for `All tool test results` download the tarball and look at the HTML.

chambm:
I had already checked that. That's where I saw the failures due to the commands not being found. I don't see any conda output there. The commands are definitely inside the conda package. The new package size is basically the same as the old one.

But I just noticed I changed the underscore in the package name to a dash inadvertently. That's almost certainly the issue (but it would've been nice to see the conda output to be sure).

chambm:
@bgruening Seems something changed with the environment since the last time I updated this. That idpAssemble command works fine on a local Ubuntu instance, even when I make the filepaths the exact same.

bgruening:
What is different now is that it is running in containers. Can it be that the container is not working?

You can test with `planemo test --docker ...`

chambm:
I don't have planemo handy to run at the moment. Does the container do something different with /tmp?

bgruening:
> Does the container do something different with /tmp?

Oh this can also be -- mh, damn. I will put it on my to do list but no promises :(

bernt-matthias:
Hey @chambm maybe you can cherry-pick this one https://github.com/galaxyproteomics/tools-galaxyp/pull/520

should solve the tmp problem. We did the same for some tools at IUC recently https://github.com/galaxyproject/tools-iuc/pull/3315

PR #507, opened 2020-10-29
Modify tolerance label text per user feedback

tmcgowan:
Modify label text for clarity "Epsilon" -> "Fragment Tolerance"

### DIFF
--- a/tools/psm_validation/macros.xml
+++ b/tools/psm_validation/macros.xml
@@ -1,5 +1,5 @@
 <macros>
-    <token name="@VERSION@">1.0.3</token>
+    <token name="@VERSION@">1.0.3+galaxy1</token>
     <xml name="help-text">
         <help>
             <![CDATA[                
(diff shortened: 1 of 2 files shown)


tools/psm_validation/macros.xml
```diff
@@ -1,5 +1,5 @@
 <macros>
-    <token name="@VERSION@">1.0.3</token>
+    <token name="@VERSION@">1.0.3+galaxy1</token>
```
bgruening, line 2: can you move the galaxy1 to the individual tool? See here for recommendation: https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#tool-versions

The idea is that we can bump tool versions automatically in the future :)

tmcgowan: Will do.

tmcgowan: Added suffix as well:

`<token name="@VERSION_SUFFIX@">1</token>`
`<tool id="psmvalidator" name="PSM-Fragment Validator" version="@VERSION@+galaxy@VERSION_SUFFIX@">`

bgruening (approved):

PR #512, opened 2020-11-06
changes for msconvert to make it runnable via singularity

bernt-matthias:
I successfully got msconvert running via singularity this way (thanks to @sneumann and @bgruening). I'm wondering if there is a way around the sudo stuff. From Galaxy sine one could do this with a call to the external_chown script (https://github.com/galaxyproject/galaxy/blob/dev/scripts/external_chown_script.py) which runs anyway in realuser setups. 

Also I'm not sure if there is a better way than to copy the wineprefix stuff .. 

(with "sucessfuly" I mean that checked the first 2 tests manually :))

Further ideas:

- probably we should clean the created tmpdir after the tool run via a trap
- maybe we use a more standard label for the outputs

### DIFF
--- a/tools/msconvert/msconvert_macros.xml
+++ b/tools/msconvert/msconvert_macros.xml
@@ -1,7 +1,7 @@
 <macros>
   <token name="@VERSION@">3.0.19052</token>
   <token name="@FULL_VERSION@">@VERSION@-089e81090</token>
-  
+
   <xml name="msconvertCommand">
     <command detect_errors="exit_code">
 <![CDATA[
@@ -39,34 +39,20 @@
       ln -s '$data_processing.precursor_refinement.input_ident' '$input_ident_name' &&
     #end if
 
-    ## create a writable copy of wine prefix
-    ## (since copying fails for some html files we swallow the exit code)
+    ## create a writable copy of wine prefix (since copying fails for some html
+    ## stderr and exit code is swallowed)
     export WINEPREFIX=\$(mktemp -d) &&
-    (cp -ar /wineprefix64/* \$WINEPREFIX || true) &&
+    (cp -a /wineprefix64/* \$WINEPREFIX 2> /dev/null || true) &&
 
     wine64 msconvert ${inputmask}
     --outdir outputs
     --${output_type}
 
-    #if $general_options.combineIonMobilitySpectra:
-    --combineIonMobilitySpectra
-    #end if
-
-    #if $general_options.simAsSpectra:
-    --simAsSpectra
-    #end if
-
-    #if $general_options.srmAsSpectra:
-    --srmAsSpectra
-    #end if
-
-    #if $general_options.acceptZeroLengthSpectra:
-    --acceptZeroLengthSpectra
-    #end if
-
-    #if $general_options.ignoreUnknownInstrumentError:
-    --ignoreUnknownInstrumentError
-    #end if
+    $general_options.combineIonMobilitySpectra
+    $general_options.simAsSpectra
+    $general_options.srmAsSpectra
+    $general_options.acceptZeroLengthSpectra
+    $general_options.ignoreUnknownInstrumentError
 
     #if $general_options.scan_summing.do_scan_summing:
       --filter "scanSumming precursorTol=$general_options.scan_summing.precursorTol scanTimeTol=$general_options.scan_summing.scanTimeTol ionMobilityTol=$general_options.scan_summing.ionMobilityTol"
@@ -183,8 +169,8 @@
       #if str($filtering.analyzer) != "false"
         --filter "analyzer $filtering.analyzer"
       #end if
-      
-    ## OUTPUT ENCODING
+
+      ## OUTPUT ENCODING
       #set $mz_encoding = str($settings.mz_encoding)
       #set $intensity_encoding = str($settings.intensity_encoding)
       #if $mz_encoding == $intensity_encoding
@@ -244,7 +230,7 @@
       <option value="ms2">ms2</option>
     </param>
 
-    <section name="data_processing" title="Data Processing Filters">  
+    <section name="data_processing" title="Data Processing Filters">
         <conditional name="peak_picking">
             <param type="boolean" name="pick_peaks" label="Apply peak picking?" truevalue="true" falsevalue="false" />
             <when value="false" />
@@ -408,7 +394,7 @@
 
 
     <section name="filtering" title="Scan Inclusion/Exclusion Filters">
-      
+
       <param name="activation" type="select" label="Filter by Activation">
         <option value="false" selected="true">no</option>
         <option value="ETD">ETD</option>
@@ -463,7 +449,7 @@
     </section>
 
     <section name="general_options" title="General Options">
-      <param argument="--combineIonMobilitySpectra" type="boolean" label="Combine ion mobility spectra" help="When false, each mobility scan is written as a separate spectrum. When true, each retention time point will have a single merged scan. For Bruker TIMS spectra, the ion mobilities will be preserved in a separate binaryDataArray, and for TIMS PASEF MS2s, each precursor will be merged separately." />
+      <param argument="--combineIonMobilitySpectra" type="boolean" truevalue="--combineIonMobilitySpectra" falsevalue="" label="Combine ion mobility spectra" help="When false, each mobility scan is written as a separate spectrum. When true, each retention time point will have a single merged scan. For Bruker TIMS spectra, the ion mobilities will be preserved in a separate binaryDataArray, and for TIMS PASEF MS2s, each precursor will be merged separately." />
 
       <conditional name="scan_summing">
         <param name="do_scan_summing" type="boolean" truevalue="true" falsevalue="false" label="Sum adjacent scans" help="Sums MS2 sub-scans whose precursors are similar in the m/z, scan time, and/or ion mobility dimensions. It is useful for some Waters DDA data and Bruker PASEF data, where sub-scans should be summed together to increase the SNR" />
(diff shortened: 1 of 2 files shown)


bgruening:
@chambm would be great if you can look at it! Thanks!!!

chambm:
Won't this change be at the expense of Docker support?

bgruening:
@chambm the Docker tests are now running here. You can check the logs, the Docker container is used and all tests passed.

chambm:
So, uh, how is that possible if it's not using sudo? Is Docker running inside Docker so it can run as root or something?

tools/msconvert/msconvert_macros.xml
```diff
@@ -217,14 +219,13 @@
 
       #if $general_options.multi_run_output.do_multi_run_output == 'false':
         --outfile '${os.path.splitext($basename)[0]}'
-        && sudo mv 'outputs/${os.path.splitext($basename)[0]}.${output_type}' '${output}' && sudo chown \$uid:\$gid '${output}'
+        && mv 'outputs/${os.path.splitext($basename)[0]}.${output_type}' '${output}'
       #else
-        && sudo chown \$uid:\$gid 'outputs' -R
```
bgruening, line 222: @bernt-matthias this might be a problem. Can you run this locally and check what the permissions are of the outputs after the tool has been running?

I assume they are not owned by the Galaxy user. So in production cleanup scripts, etc can not clean them?

tools/msconvert/msconvert_macros.xml
```diff
@@ -39,10 +39,12 @@
       ln -s '$data_processing.precursor_refinement.input_ident' '$input_ident_name' &&
     #end if
 
-    uid=`id -u` &&
-    gid=`id -g` &&
+    ## create a writable copy of wine prefix
+    ## (since copying fails for some html files we swallow the exit code)
+    export WINEPREFIX=\$(mktemp -d) &&
+    (cp -ar /wineprefix64/* \$WINEPREFIX || true) &&
```
bgruening, line 45: I think this is creating a lot of messages on the stderr. Please look at the planemo outputs.

chambm: OK somehow I missed this part. Hmm. So this is going to copy WINEPREFIX for every job even it runs on the same node. Is it feasible to cache it on the instance so multiple jobs run in series don't pay for the copy each time? That's how the mywine script works. I can definitely imagine race condition issues with that though.

sneumann:
Hi, I am not at a keyboard, but my old idea was that to make wine checks that the prefix is owned by the user happy,

the user owned wine prefix can get away with links to the drive c: content, so only the registry files are copied, not the entire windows "system" and msconvert with dependencies, they can even be read only.

If that doesn't make sense, I need to find the examples I have later.

Gruß Steffen


I blame Android for the brevity and typos

bernt-matthias:
> Won't this change be at the expense of Docker support?

Primarily I tried to make it running with singularity. Hoping that we might have a 2nd version of the tool that supports singularity. 

So if I fixed the docker tests .. this was no intended. But I would be very happy with this result. LOL

tools/msconvert/msconvert_macros.xml
bernt-matthias: Will do.

tools/msconvert/msconvert_macros.xml
bernt-matthias: Might be an option. Question if is if actually something is written into `WINEPREFIX` or if the necessary write access is just a technical necessity. 

Also for Galaxy's singularity support I do not like the fact that each user downloads a separate copy of the container. But maybe this is avoidable via job_conf.xml (but probably only for a single container).. lets see.

sneumann:
Iirc the main requirement for wine is who the *owner* of the wine prefix is. Not sure about testing for writable. Yours Steffen


I blame Android for the brevity and typos

tools/msconvert/msconvert_macros.xml
bernt-matthias: @chambm could you provide me a link explaining a bit more on the `mywine` script?

tools/msconvert/msconvert_macros.xml
bernt-matthias: In my setup I run jobs as real user and use the option to output to the wob working dir. This dir is chowned after the run to the galaxy system user...  but I run with singularity anyway .. so this info does not help much .. but as far as I remember real user setups require output to working dir?  

For the `planemo test` run on my local machine also the permissions were correct. Probably this is because `docker run ...  --user 1000:1000` is used (1000 is me). So it seems that single user Galaxy setups are also no problem.

bernt-matthias:
> the user owned wine prefix can get away with links to the drive c: content, so only the registry files are copied, not the entire windows "system" and msconvert with dependencies, they can even be read only.

Not sure about this, since we use `cp -ar`.

But since this happens inside the container (not sure how `/tmp/` is treated there) its gone after the job anyway.

tools/msconvert/msconvert_macros.xml
bernt-matthias: btw. on my system cpopying needs ~14s

bernt-matthias: > I think this is creating a lot of messages on the stderr. Please look at the planemo outputs.

So I redirect to `/dev/null` now. But maybe its better for future debugging to redirect to some file?

bernt-matthias:
Any ideas how to continue here?

chambm (changes requested):
Can we make the approach used (sudo vs. copy wineprefix) dependent on an environment variable? Maybe there's even a pre-existing singularity variable we can check for?

bernt-matthias:
Sounds fine for me. Would you agree to use the modified version the default and the sudo behaviour dependent on a variable? Since the changes fixed the testing...

bernt-matthias:
Or we try to fix the tests (but I have no idea how) and then make the new behavior depend on the variable.

chambm:
I was thinking the variable would be based on whether it's Singularity or Docker running, so if the tests are run with Docker then it should use the old behavior. If that's broken then I'd like to see why: it was working last time I made a change IIRC.

bernt-matthias:
It failed for the last PR https://github.com/galaxyproteomics/tools-galaxyp/pull/470 and its failing weekly CI with the error `sudo: unknown uid 1001: who are you?`. 

I think this is since we switched to github actions.

Oddly it runs the same `planemo test` command locally, lets try in a PR: https://github.com/galaxyproteomics/tools-galaxyp/pull/513

chambm:
The output from the GitHub actions is inscrutable to me. :/ Where do I see the actual commands that were run? Can we also check for an environment variable in this GitHub environment and condition on that as well as singularity? (where do I even see the environment variables that github is using?)

bernt-matthias:
I wrote you a message on gitter. If you don't read over there I would like to suggest to have a short VC (~30min). I could introduce you to the github actions stuff and we could discuss this PR?

bernt-matthias:
So, how about this https://github.com/galaxyproteomics/tools-galaxyp/pull/512/commits/1153416a4f2cbc389b0920ccc6b2bc277c8b41b4 

- test if sudo is possible (e.g. in dockers running with sudo)
- if so use old sudo based implementation
- if not use the new implementation based on copying the wineprefix

In the discussion with @chambm we found out that the container has three users corresponding to the usegalaxy.eu and two other use cases (travis I guess). So systems that use another user (like github actions) wont work with the sudo approach. Real user Galaxies with strict security policies (ie no sudo) -- like the the one I'm running -- will never work with the sudo approach (since in addition to the sudo problem also user mapping will be impossible). 

In the discussion we also thought about checking if we are actually inside a container. There seem to be two variables that might be useful 

```
SINGULARITY_CONTAINER=/global/apps/galaxy/containers/msconvert/pwiz-skyline-i-agree-to-the-vendor-licenses:3.0.19052-089e81090.sif
SINGULARITY_NAME=pwiz-skyline-i-agree-to-the-vendor-licenses:3.0.19052-089e81090.sif
```

but I'm not sure and I think the current solution is preferable.

On the long run I may suggest that systems using the old approach might switch and also the container might be simplified (users could be removed). But we should see if it really works in practice I guess.

TODOs:
- [ ] add a little bit of documentation to README / sources
- [ ] anyone objecting to relabeling the outputs in a more standard way?

chambm:
Sounds good. What do you mean by "relabeling the outputs"?

bernt-matthias:
> What do you mean by "relabeling the outputs"?

`label="${tool.name} on ${on_string}"` or maybe something like `label="${tool.name} on ${on_string}: OUTPUTTYPE"`

which would be what most tools do.

chambm:
TBH that's one of my least favorite things about Galaxy: default dataset labels that don't tell me what the dataset is without a few clicks or a lot of history scrolling. Also too many proteomics tools rely on filenames for the "more standard" way to be practical. Many proteomic tools expect the filename to be the MS run name which gets carried through to downstream analysis.

bernt-matthias:
I agree to some extent. But see it this way currently msconvert results look (without additional clicks) like uploaded data sets. With the alternative labeling one sees immediately that they are processed and what was the input. 

Both have pros and cons. 

Do you have an example of a downstream tool that creates problems?

chambm:
I expect SearchGUI/PeptideShaker won't show the run names in the report correctly if the msconvert output doesn't preserve the MS run name in the MGF dataset labels.

bernt-matthias:
I think the trick in this case is always collections. The collection name will be equal to the tool output and the element names are equal to the names in the input collection.

Matthias313:
Hi everyone,

I just noticed that you seem to activily working on some changes to the msconvert tool and I was wondering if along those changes it would be also possible to update msconvert to a newer version?

With the currently newest version I observed some strange effects while converting from .raw to .mzML format. 
Somehow an "inf" value was introduced.
Furthermore, newer devices (such as the Orbitrap Eclipse) are unknown in that version, which might be implemented in the later versions of msconvert. 

This completely off from the actual topic in the ongoing dicussion here, i just thought that maybe a version update could be done togetehr with those other changes. 

Of course I fully understand if this is not feasable or not managable in addition with the fixes mentioned in the discussion above! =)

chambm:
@bernt-matthias I think that can be done separately if users actually complain about it. Myself I don't like the inconsistency between the dataset/element names when run as a collection vs. run as one or more datasets. This PR seems ready to merge.

@Matthias313 I'll do that in a separate PR. Thanks for the reminder.

PR #514, opened 2020-11-16
Update msconvert to 3.0.20287-769529fa4

chambm:


bgruening:
I think the test data needs to be updated and checked. Thanks @chambm

chambm:
Ugh. :( What's the fastest way to get a planemo container up and running these days? The changes all look legit, I just need to tell it to overwrite the test data with the new output.

bgruening:
Try `planemo test --docker --update_test_data ...`

chambm:
Is there a Docker one-liner to get a working planemo environment with git?

bernt-matthias:
We should reset the tool version to `.0` or remove this part completely. 

You get a planemo env with 

```
virtualenv .venv -p python3
. .venv/bin/activate
pip install planemo
```

test with 

```
planemo test --galaxy_python_version 3.7 --galaxy_branch release_20.09 --biocontainers --no_conda_auto_init
```
optionally appending `--update_test_data`

bernt-matthias:
If you like you may cherry pick the commit from https://github.com/galaxyproteomics/tools-galaxyp/pull/515

bernt-matthias:
Can I help here somehow?

chambm:
I just haven't had time to get a planemo environment running (i.e. working on other stuff) in order to update the test data. I run/develop on Windows mostly, but I have Linux containers set up for my Windows 10 machine. But I guess Docker in Docker would be a pain in the butt there.

bernt-matthias:
Test data has been updated. Can you check the validity? 

Do you know the `ftype` of the input of the 2nd test?

tools/msconvert/test-data/small-peakpicking-vendor-allMS.mzML
```diff
@@ -79,15 +84,15 @@
     </dataProcessingList>
     <run id="small" defaultInstrumentConfigurationRef="IC1" startTimeStamp="2005-07-20T14:44:22Z" defaultSourceFileRef="RAW1">
       <spectrumList count="48" defaultDataProcessingRef="pwiz_Reader_Thermo_conversion">
-        <spectrum index="0" id="controllerType=0 controllerNumber=1 scan=1" defaultArrayLength="1750">
+        <spectrum index="0" id="controllerType=0 controllerNumber=1 scan=1" defaultArrayLength="1810">
```
chambm, line 87: For future reference, this is changing because pwiz is now including exception/reference peaks from the Thermo reader.

tools/msconvert/test-data/Rpal_01.pepXML.mzRefinement.tsv
```diff
@@ -1,2 +0,0 @@
-ThresholdScore	ThresholdValue	Excluded (score)	Excluded (mass error)	MS1 Included	MS1 Shift method	MS1 Final stDev	MS1 Tolerance for 99%	MS1 Final MAD	MS1 MAD Tolerance for 99%	MS2 Included	MS2 Shift method	MS2 Final stDev	MS2 Tolerance for 99%	MS2 Final MAD	MS2 MAD Tolerance for 99%
```
chambm, line 1: I wouldn't expect this change. I'll check out the test output.

tools/msconvert/msconvert_macros.xml
```diff
@@ -558,115 +556,115 @@
       <param name="pick_peaks" value="true" />
       <param name="pick_peaks_algorithm" value="vendor" />
       <param name="pick_peaks_ms_levels" value="1-" />
-      <output name="output" file="small-peakpicking-vendor-allMS.mzML" lines_diff="4" />
+      <output name="output" file="small-peakpicking-vendor-allMS.mzML" ftype="mzml" lines_diff="4" />
     </test>
     <test>
       <param name="input" value="ThyroglobMRM000003.d.tar" />
```
chambm, line 562: brukertdf.d.tar

tools/msconvert/msconvert_macros.xml
```diff
@@ -558,115 +556,115 @@
...
-      <output name="output" file="MassLynxTest.mzML" lines_diff="4" />
+      <output name="output" file="MassLynxTest.mzML" ftype="mzml" lines_diff="4" />
     </test>
     <test>
-      <param name="input" value="AgilentMassHunterTest.d.tar" />
+      <param name="input" value="AgilentMassHunterTest.d.tar" ftype="agilentmasshunter.d.tar" />
       <param name="license_agreement" value="true" />
       <param name="output_type" value="mzXML" />
-      <output name="output" file="AgilentMassHunterTest.mzXML" lines_diff="4" />
+      <output name="output" file="AgilentMassHunterTest.mzXML" ftype="mzml" lines_diff="4" />
```
chambm, line 579: mzxml

bernt-matthias: done

chambm:
Thanks for updating the test files. I think they should all be updated though? Test 15-33 are failing because of pwiz and CV version updates.

bernt-matthias:
> I think they should all be updated though? 

The `--update_test_data` only updates test files for failing tests. We may enforce this by temporarily removing/replacing the allowed `lines_diff` or `delta` from the xml. Should I?

bgruening:
> The --update_test_data only updates test files for failing tests. We may enforce this by temporarily removing/replacing the allowed lines_diff or delta from the xml. Should I?

I don't think this is true. It should update all files. At least it did. Are you maybe running with `--failed` as well?

bernt-matthias:
> Are you maybe running with --failed as well?

Ups

chambm:
I don't understand how that would be a problem. Unless those tests aren't failing for locally but they're failing on GitHub?

chambm:
The mzRefinement test failure is a bug in msconvert. I'm going to have to fix and push a new version.

bernt-matthias:
> The mzRefinement test failure is a bug in msconvert. I'm going to have to fix and push a new version.

cool. then I have some time to understand if / if not the sudo hack works :) .. I observed some problems.

chambm:
Hmm. The error message is actually just a warning (I get it locally), but the mzRefinement.tsv output seems to be created as it should be. Can you try reverting the mzRefinement.tsv file? In the GitHub check showing a diff, which one is the "reference" (checked into git), the "local file" or the "history file"?

bernt-matthias:
> Hmm. The error message is actually just a warning (I get it locally), but the mzRefinement.tsv output seems to be created as it should be. Can you try reverting the mzRefinement.tsv file? In the GitHub check showing a diff, which one is the "reference" (checked into git), the "local file" or the "history file"?

I hope that this diff is fixed with the latest commit. 

```
--- local_file
+++ history_data
@@ -27,12 +27,13 @@
       <processingOperation name="Conversion to mzML"/>
     </dataProcessing>
     <dataProcessing>
-      <software type="conversion" name="ProteoWizard software" version="3.0.19038"/>
+      <software type="conversion" name="ProteoWizard software" version="3.0.20287"/>
```

`local_file` is the one in the `test-data` folder and `history_data` is the one generated in the planemo run. So in the above example the difference is because the file in `test-data` was created with the old version.

bernt-matthias:
Tests pass :) 

One more thing to do: we see some `Datatype class not found for extension: mzML` to fix this we need to rename mzML and mzXML files to use lower case extensions. Working on it.

tools/msconvert/msconvert_macros.xml
```diff
@@ -219,6 +219,13 @@
       fi
       #if $general_options.multi_run_output.do_multi_run_output == 'false':
         && mv 'outputs/${os.path.splitext($basename)[0]}.${output_type}' '${output}'
+      #else
+        ## make mzML and mzXML extensions lower case (the Galaxy data type is ) otherwise detetion of the file
+        #if $output_type == 'mzML' or $output_type == 'mzML'
```
bgruening, line 224: could that work `for i in $( ls | grep [A-Z] ); do mv -i "$i" "`echo $i | tr 'A-Z' 'a-z'`"; done`

chambm: mzML is here twice, I guess the second should be mzXML

bernt-matthias: > could that work `for i in $( ls | grep [A-Z] ); do mv -i "$i" "`echo $i | tr 'A-Z' 'a-z'`"; done`

This would also make the non-extension part of the file name lower case.

bernt-matthias: > mzML is here twice, I guess the second should be mzXML

Should be fixed

bgruening (approved):

bernt-matthias:
> The mzRefinement test failure is a bug in msconvert. I'm going to have to fix and push a new version.

Did this happen already? Or is it still necessary?

bgruening:
Ready from my side, thanks a lot @chambm and @bernt-matthias

PR #518, opened 2020-11-22
try to fix weekly CI fonts problems

bernt-matthias:


tools/encyclopedia/macros.xml
```diff
@@ -3,6 +3,7 @@
     <xml name="requirements">
         <requirements>
             <requirement type="package" version="@VERSION@">encyclopedia</requirement>
+            <requirement type="package" version="1">fonts-conda-ecosystem</requirement>
```
bgruening, line 6: we should add this to the conda package, isn't it?

bernt-matthias: yep. just want to test if it helps. one of the tools needs massive amounts of memory for the test .. crashing my laptop.

bernt-matthias: I also needed to add openjdk 11 .. by default 10 was installed which misses some packages, e.g. fontconfig. tested successfully locally for encyclopedia

not sure whats the best way to fix it, just update the bioconda recipes in the same way?

bgruening: yes, I think that should work

bernt-matthias: https://github.com/bioconda/bioconda-recipes/pull/25508

tools/meta_proteome_analyzer/meta_proteome_analyzer.xml
```diff
@@ -4,7 +4,6 @@
     </description>
     <requirements>
         <requirement type="package" version="1.4.1">mpa-portable</requirement>
-        <requirement type="package" version="1">fonts-conda-ecosystem</requirement>
```
bernt-matthias: Unfortunately this is needed (openjdk 11 only brings fontconfig). Can you restore it.

bernt-matthias:
Who could tell us something on the `encyclopedia_searchtolib` error? @jj-umn ?

Issue #503, opened 2020-10-16
openms: wishlist


bernt-matthias:
- [ ] Add tool category to help text
- [x] SiriusAdapter, AssayGeneratorMetabo:
  - [x] hardcode out_workspace_directory
  - [x] Sirius Workspace Successful initialized at: /home/berntm/.sirius
- [x] PSMFeatureExtractor: needs two inputs in and in_list

PR #526, opened 2020-12-01
WIP: OpenMS 2.6 improvements

bernt-matthias:
fixes https://github.com/galaxyproteomics/tools-galaxyp/issues/503


tools/openms/DatabaseFilter.xml
```diff
@@ -75,6 +75,8 @@ python3 '$__tool_directory__/fill_ctd.py' '@EXECUTABLE@.ctd' '$args_json' '$hard
   <help><![CDATA[Filters a protein database (FASTA format) based on identified proteins
 
 
-For more information, visit http://www.openms.de/doxygen/release/2.6.0/html/UTILS_DatabaseFilter.html]]></help>
+For more information, visit http://www.openms.de/doxygen/release/2.6.0/html/UTILS_DatabaseFilter.html
+
+tool catecory: Utilities]]></help>
```
bgruening, line 80: I don't think we should do this. This should be really on the Admin level or part of the EDAM ontology, no?

tools/openms/DatabaseSuitability.xml
```diff
@@ -94,6 +94,8 @@ python3 '$__tool_directory__/fill_ctd.py' '@EXECUTABLE@.ctd' '$args_json' '$hard
   </tests>
   <help><![CDATA[Computes a suitability score for a database which was used for a peptide identification search. Also reports the quality of LC-MS spectra.
 
-]]></help>
+
+
+tool catecory: ]]></help>
```
bgruening, line 99: the category is missing here, maybe a bug

tools/openms/DatabaseFilter.xml
bernt-matthias: > I don't think we should do this.

Maybe :) Its easy to remove. 

> This should be really on the Admin level or part of the EDAM ontology, no?

Do you have something specific in mind. The idea was to (kind of) simulate what you have in knime where you have these categories of tools. Don't think that the tools can easily be mapped to EDAM, or?

bgruening: If would not mirror KNIME just because ... ideally we group tools in meaningful categories and if EDAM offers those we have a nice ontology that we can use. I would imagine that most of them can be mapped to EDAM ... but this will still take time.

bernt-matthias: Agreed. I undid the changes.

bernt-matthias:
Do you think I should test a deployment to TTS? In order to see if the `exclude` works?

bgruening:
> Do you think I should test a deployment to TTS? In order to see if the exclude works?

Maybe for 1 or 2 tools. But not all we can save some space.

tools/openms/macros.xml
```diff
@@ -4,7 +4,7 @@
      add additional macro files using the m/macros parameter -->
 <macros>
   <token name="@TOOL_VERSION@">2.6</token>
-  <token name="@GALAXY_VERSION@">0</token>
+  <token name="@GALAXY_VERSION@">1</token>
```
bgruening, line 7: can we just bump the tools that need to be bumped and not all 120 tools. I'm really worried about our toolbox and the parsing/loading/caching involved.

bernt-matthias:
> > Do you think I should test a deployment to TTS? In order to see if the exclude works?
> 
> Maybe for 1 or 2 tools. But not all we can save some space.

Does this make sense given that we use `auto_tool_repositories`.

tools/openms/macros.xml
bernt-matthias: Sure. But I also have no feedback yet from the OpenMS community if they prefer to continue with updated tools or to collect more issues which I could solve in this PR. Will ask again.

bernt-matthias: Just asking myself how to manage this with automatic tool building. One would need to remove the central `@GALAXY_VERSION@` and bump only tools that really change, i.e. compare new and old xml.

bgruening:
> > > Do you think I should test a deployment to TTS? In order to see if the exclude works?
> > 
> > 
> > Maybe for 1 or 2 tools. But not all we can save some space.
> 
> Does this make sense given that we use `auto_tool_repositories`.

Afaik if you change the macro all tools will be updated. If you only change one tool, only one repo will be updated as all other files have not changed. At least this is what I would aim for.

bernt-matthias:
> Afaik if you change the macro all tools will be updated. If you only change one tool, only one repo will be updated as all other files have not changed. At least this is what I would aim for.

Likely. Would this also work without bumping the tool version? Probably not, or? 

Problem is, if we bump the tool version manually, then this is overwritten the next time we autogenerate the tools.

bernt-matthias:
OK Idea:

- We add a little yaml/tsv config file which saves the tool versions (i.e. the `X` in `+galaxyX`)  
- The converter reads the files and sets the tool versions accordingly

This would mean to remove the `GALAXY_VERSION` token.

The config file could be handled manually or I write a script that bumps versions based on `git diff --name-status master`. 

When upgrading to the next OpenMS version we just need to empty the file.

Sounds like a plan?

bernt-matthias:
OK. Maybe like so... Hope that it works this way. Even if the `version` attribute changed it essentially is the same for all but the tools affected by the tools: i.e. all java wrappers that might be affected by the extendend stdio definition, the sirius based tools that are  affected by the new hardcoded variable, and the PSM feature extractor. 

Should we try to deploy this first manually to the TTS to see what happens? Should be done by someone who knows the API key.

bernt-matthias:
Its amazing to see that the complete suite can now be tested with only 4 chunks instead of 40. Seems that the overhead caused by Galaxy has really been dramatically reduced in 20.09... and maybe also the PR workflow got a bit more efficient :)

bgruening:
> Should we try to deploy this first manually to the TTS to see what happens? Should be done by someone who knows the API key.

Yes. Do you want to have the creds? Ping me on gitter please.

bernt-matthias:
Hi @bgruening I think I'm finished here... assuming that the tests pass. 

Because of the changes to the macros the github workflow will update everything. So I guess we need to push this manually, or? If so, how? my original idea would be to run shed_update on a copy of the openms folder where all but the changed tools are deleted.

bgruening:
Yes, that's what I did last time as well. Get rid of all tools that you don't want to update. Thanks

bgruening (approved):

bernt-matthias:
OK. Merged and cancelled the workflow. Will now update the tools manually

PR #530, opened 2020-12-17
Update Mycoplasma ids

gallardoalba:
The ids of _Mycoplasma arginini_ and _M. fermentans_ have been updated by alternative ones corresponding to different strains. The previous ids pointed to defective Uniprot entries which caused the Protein database downloader doesn't yield any result.


bgruening:
@bernt-matthias does that look good to you?

bernt-matthias:
Hey @gallardoalba this is a good start. 

The problem is that `uniprot_taxons.loc.sample` is just an example that has no effect in any installation of the tool.  Missing is also a `tool_data_table_conf.xml.sample`. 

Currently also missing is a test that actually uses the uniprot db. For this one would need a `tool_data_table_conf.xml.test` and a `uniprot_taxons.loc` in `test-data` (that you refer to in the tool_data_table_conf.xml.test). 

Maybe you find some inspiration over here: https://github.com/galaxyproject/tools-iuc/tree/master/tools/mitos

PR #533, opened 2020-12-23
Add DIAlignR

npinter:
DIAlignR is an R package for retention time alignment of targeted mass spectrometric data, including DIA and SWATH-MS data.


tools/dialignr/.shed.yml
```diff
@@ -0,0 +1,11 @@
+name: dialignr
+owner: iuc
```
bgruening, line 2:
```suggestion
owner: galaxyp
```

tools/dialignr/.shed.yml
```diff
@@ -0,0 +1,11 @@
+name: dialignr
+owner: galaxyp
+remote_repository_url: "https://github.com/galaxyproject/tools-iuc/tree/master/tools/dialignr"
+homepage_url: "https://github.com/shubham1637/DIAlignR"
+long_description: |
+ DIAlignR is an R package for retention time alignment of targeted mass spectrometric data, including DIA and SWATH-MS data. This tool works with MS2 chromatograms directly and uses dynamic programming for alignment of raw chromatographic traces. DIAlignR uses a hybrid approach of global (feature-based) and local (raw data-based) alignment to establish correspondence between peaks.
+categories:
+- "DIA Analysis"
```
bernt-matthias, line 8: You need to choose an existing category: `Proteomics`?

tools/dialignr/.shed.yml
```diff
@@ -0,0 +1,11 @@
+name: dialignr
+owner: galaxyp
+remote_repository_url: "https://github.com/galaxyproject/tools-iuc/tree/master/tools/dialignr"
```
bernt-matthias, line 3: repo url needs to change

tools/dialignr/.shed.yml
npinter: Thanks!!

tools/dialignr/.shed.yml
npinter: Thanks!!

tools/dialignr/dialignr.xml
```diff
@@ -0,0 +1,51 @@
...
+    </macros>
+    <requirements>
+        <requirement type="package" version="@TOOL_VERSION@">bioconductor-dialignr</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+    mkdir -p ./data/mzml &&
+    mkdir -p ./data/osw &&
+
+    #for $mzml in $input_mzml_files:
+        #set $file_name_mzml = $mzml.element_identifier.split(".")
```
bernt-matthias, line 14: you will need to sanitize the identifier like here: https://github.com/galaxyproject/tools-iuc/blob/2808c3c3e47e43eaf835d725fa64c63ee6dc0913/tools/chromeister/chromeister.xml#L13 otherwise its not safe to use for the link name. 

Since in Galaxy users can chose to have any characters in the identifiers (including `'.'`) `split(".")` might not be a good idea. Anyway in real life there will not be an extension in the identifier (its only in the tests where the identifier is equal to the uploaded file name).

npinter: Thanks for pointing that out! For the sake of simplicity, I just removed the extension of the test files.

bernt-matthias:
Looks good to me now. 

Do you plan to extend this by: 
- some of the parameters as shown here: https://www.bioconductor.org/packages/release/bioc/manuals/DIAlignR/man/DIAlignR.pdf#page.7
- the plotting functionality as outlined here https://www.bioconductor.org/packages/release/bioc/vignettes/DIAlignR/inst/doc/DIAlignR-vignette.html? Would be quite fancy.

tools/dialignr/dialignr.xml
```diff
@@ -0,0 +1,53 @@
...
+    ln -s '$input_osw_file' './data/osw/merged.osw' &&
+
+    Rscript --vanilla '${__tool_directory__}/dialignr.R' `pwd`
+    ]]></command>
+    <inputs>
+        <param name="input_mzml_files" type="data" format="mzml" multiple="true" label="MZML file(s)" help="Sample file(s) in .MZML format." />
+        <param name="input_osw_file" type="data" format="osw" multiple="false" label="OSW file" help="Sample file in .OSW format." />
+    </inputs>
+    <outputs>
+        <data name="out_aligned_targeted_runs" format="csv" label="${tool.name} on ${on_string}: alignedTargetedRuns.csv" from_work_dir="alignedTargetedRuns.csv" />
```
bgruening, line 30: can we convert this file to tsv in R ... CSV are not that useful in Galaxy

npinter: Done!

npinter:
> 
> 
> Looks good to me now.
> 
> Do you plan to extend this by:
> 
>     * some of the parameters as shown here: https://www.bioconductor.org/packages/release/bioc/manuals/DIAlignR/man/DIAlignR.pdf#page.7
> 
>     * the plotting functionality as outlined here https://www.bioconductor.org/packages/release/bioc/vignettes/DIAlignR/inst/doc/DIAlignR-vignette.html? Would be quite fancy.

I'm going to add an advanced section for the additional parameters of the `alignTargetedRuns` function. The plotting fuctionality is indeed fancy, however this seems to be a bigger code change and might be something for a future update of the wrapper. I've discussed that with @Matthias313 earlier and decided to provide only the basic functionality of DIAlignR in first place.

bgruening (approved):
Looks good to me, but we wait for more parameters?

tools/dialignr/dialignr.R
```diff
@@ -0,0 +1,13 @@
+library(DIAlignR)
+
+args <- commandArgs(trailingOnly = FALSE)
+
+data_path <- paste0(args[7], "/data/")
+runs_file_path <- paste0(args[7], "/runs.txt")
```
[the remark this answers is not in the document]
bernt-matthias: or just feed the path strings directly to the function .. there is no need to define extra variables.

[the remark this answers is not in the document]
npinter: The data folder is job_working_directory and the R script is in the tool directory. I've tried different things like symlink the R script in the job_working directory, but that didn't work either..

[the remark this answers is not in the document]
bernt-matthias: > The data folder is in the job_working_directory and the R script is in the tool directory. I've tried different things like symlink the R script in the job_working directory, but that didn't work either..

True. But the suggested change should work, because the the current working dir (`pwd`) will be the job working dir.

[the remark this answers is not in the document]
bgruening: @npinter you could pass the `pwd` into the R script.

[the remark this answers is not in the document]
npinter: > 
> 
> @npinter you could pass the `pwd` into the R script.

That's what I did!

> True. But the suggested change should work, because the the current working dir (`pwd`) will be the job working dir.

That was to easy... changed it 👍 Thanks!

tools/dialignr/dialignr.xml
```diff
@@ -0,0 +1,155 @@
...
+                <option value="simpson">simpson</option>
+            </param>
+            <param name="fitEMG" type="boolean" truevalue="TRUE" falsevalue="FALSE" label="Fit EMG" help="Enable/disable exponentially modified gaussian peak model fitting." />
+            <param name="recalIntensity" type="boolean" truevalue="TRUE" falsevalue="FALSE" label="Recal intensity" help="Recalculate intensity for all analytes." />
+            <param name="fillMissing" type="boolean" checked="true" truevalue="TRUE" falsevalue="FALSE" label="Fill missing" help="Calculate intensity for analytes for which features are not found." />
+            <param name="smoothPeakArea" type="boolean" truevalue="TRUE" falsevalue="FALSE" label="Smooth peak area" help="No: Raw chromatograms will be used for quantification. Yes: Smoothed chromatograms will be used for quantification." />
+        </section>
+    </inputs>
+    <outputs>
+        <data name="out_aligned_targeted_runs" format="tsv" label="${tool.name} on ${on_string}: alignedTargetedRuns.tsv" from_work_dir="alignedTargetedRuns.tsv" />
```
bgruening, line 132: please use `tabular` instead of tsv

tools/dialignr/dialignr.xml
```diff
@@ -0,0 +1,155 @@
...
+    </outputs>
+    <tests>
+        <test>
+            <param name="input_mzml_files" ftype="mzml" value="hroest_K120809_Strep0_PlasmaBiolRepl2_R04_SW_filt,hroest_K120809_Strep10_PlasmaBiolRepl2_R04_SW_filt" />
+            <param name="input_osw_file" ftype="osw" value="merged.osw" />
+            <output name="out_aligned_targeted_runs" file="alignedTargetedRuns.tsv" />
```
bgruening, line 138: Nico could we add one test that uses some advanced parameters? The rest is awesome! Thanks.

bgruening (approved):

PR #534, opened 2021-01-04
OpenMS improvement / update

bernt-matthias:
PR is intended to collect ideas / improvements for OpenMS wrappers. We can merge it as soon as there is a significant improvement.

- simplify int & float list validators


bernt-matthias:
Message to myself: this does not work because: the expression raises an exception in case of wrong input, but validator needs true or false

PR #539, opened 2021-01-15
add mztab2tsv

bernt-matthias:
using pyteomics

### DIFF
--- a/tools/pyteomics/mztab2tsv.xml
+++ b/tools/pyteomics/mztab2tsv.xml
@@ -1,0 +1,98 @@
+<tool id="mztab2tsv" name="pyteomics" version="@TOOL_VERSION@" profile="20.01" license="MIT">
+    <description>convert mztab to tabular</description>
+    <macros>
+        <token name="@TOOL_VERSION@">4.4.1</token>
+        <xml name="output" token_type="" token_label="">
+            <data name="out_@TYPE@" format="tabular" from_work_dir="@TYPE@.tsv" label="${tool.name} on ${on_string}: @LABEL@">
+                <filter>"@TYPE@" in out_select</filter>
+            </data>
+        </xml>
+    </macros>
+    <xrefs>
+        <xref type="bio.tools">pyteomics</xref>
+    </xrefs>
+    <edam_topics>
+        <edam_topic>topic_0121</edam_topic><!-- proteomics -->
+        <edam_topic>topic_3520</edam_topic><!-- proteomics experiment-->
+    </edam_topics>
+    <edam_operations>
+        <edam_operation>operation_3434</edam_operation><!--	Convert a data set from one form to another -->
+    </edam_operations>
+    <requirements>
+        <requirement type="package" version="@TOOL_VERSION@">pyteomics</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+    ## make sure that selected outputs exist even if the tool does not generate them
+    #for o in $out_select
+        touch '$o'.tsv &&
+    #end for
+    '$__tool_directory__/mztab_reader.py' --path_in '$path_in'
+    ]]></command>
+    <inputs>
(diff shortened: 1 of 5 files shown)


bernt-matthias:
python script by @annalefarova

tools/pyteomics/mztab2tsv.xml
```diff
@@ -0,0 +1,98 @@
+<tool id="mztab2tsv" name="pyteomics: convert mztab to tabular" version="@TOOL_VERSION@" profile="20.01" license="TODO_license_in_spdx_format">
```
bgruening, line 1: maybe the name just "pyteomics" and the rest as description?

license is missing ... its the license of the python script

tools/pyteomics/mztab2tsv.xml
```diff
@@ -0,0 +1,98 @@
...
+    <edam_operations>
+        <edam_operation>operation_3434</edam_operation><!--	Convert a data set from one form to another -->
+    </edam_operations>
+    <requirements>
+        <requirement type="package" version="@TOOL_VERSION@">pyteomics</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+    ## make sure that selected outputs exist even if the tool does not generate them
+    #for o in $out_select
+        touch "$o".tsv &&
```
bgruening, line 27: single quotes

tools/pyteomics/mztab2tsv.xml
```diff
@@ -0,0 +1,98 @@
...
+            <option value="smf">Small molecule feature table (v2)</option>
+            <option value="sme">Small molecule evidence table (v2)</option>
+        </param>
+    </inputs>
+    <outputs>
+        <expand macro="output" type="mtd" label="Metadata"/>
+        <expand macro="output" type="prt" label="Proteins"/>
+        <expand macro="output" type="pep" label="Peptides"/>
+        <expand macro="output" type="psm" label="Peptide spectrum matches"/>
+        <expand macro="output" type="sml" label="Small molecules"/>
```
bgruening, line 48: are all filetypes available in Galaxy?

bgruening (commented):
Very nice, only some small suggestions.

bgruening:
@annalefarova that looks great! Thanks a lot.

tools/pyteomics/mztab2tsv.xml
bernt-matthias: these are all tabular files .. see the macro?

tools/pyteomics/mztab2tsv.xml
bernt-matthias: done

tools/pyteomics/mztab2tsv.xml
bernt-matthias: preferences anyone .. ? The license question is always a bit annoying to me .. I would like https://en.wikipedia.org/wiki/WTFPL :)

bgruening (approved):

bgruening:
Merci! @annalefarova do you want to do step 3 as well :) https://galaxyproject.eu/posts/2020/08/22/three-steps-to-galaxify-your-tool/

annalefarova:
@bgruening thank you, I can try to do that :)
Besides I've also updated my function to write an empty `tsv` file instead of file with an empty string when some table is missing (@bernt-matthias asked about that). Can I create PR with an update to [this file](https://github.com/bernt-matthias/tools-galaxyp/blob/cef579d20013d6b67ac4f682191276fe426c78bc/tools/pyteomics/mztab_reader.py).  Only internal implementation has changed the interface remained the same.

bgruening:
Yes, this would be the way. Just PR this file and tests will run.

bernt-matthias:
> For step 3, which tool_panel_section_label should I write? 

Metabolomics or proteomics? I guess Massspec.. does not exist. 

> And the name is read_mztab, right?

Should be mztab2tsv

PR #545, opened 2021-01-22
update peptide_genomic_coordinate

jj-umn:
Find peptide sequence in the protein using a regex pattern.
This is to eliminate issues with leucine/isoleucine interchangeability.


tools/peptide_genomic_coordinate/peptide_genomic_coordinate.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="peptide_genomic_coordinate" name="Peptide Genomic Coodinate" version="0.1.1">
+<tool id="peptide_genomic_coordinate" name="Peptide Genomic Coordinate" version="0.1.2">
     <description>Get Peptide's genomic coordinate using mzsqlite DB and genomic mapping sqlite DB</description>
     <requirements>
         <requirement type="package" version="3.7.1">python</requirement>
```
bgruening, line 4: you could increase the python version if you are on it, maybe 3.9 is enough?

jj-umn:
There is more problems in this code than just using a String find function to locate the peptide.   
I'm going to inspect this code more thoroughly.

PR #555, opened 2021-02-06
OpenMS place each in/output symlink in unique dir

bernt-matthias:
before each symlink was placed in a dir named by the name of the corresponding parameter. this still allows for conflicts if several
data sets have the same identifier.

solution: place symlinks in directory parameter_name/galaxy-dataset-id

usecase: ConsensusID takes the result of different search engines for the same sample. then the identifiers of the inputs are all the sample name which leads to a conflict.

question is how we can manage the tool version bump since we manually bumped some of the tools (bump.json). we could continue to bump manually, or just use the macro token and increase it to max bump + 1.

For most tools the change is just an additional import.. I could also change the converter such that the import is only added if needed.


tools/openms/AccurateMassSearch.xml
```diff
@@ -14,6 +14,7 @@
   <command detect_errors="exit_code"><![CDATA[@QUOTE_FOO@
 @EXT_FOO@
 #import re
+#import os.path
```
bgruening, line 17: just import os?

bernt-matthias: If you like I can change this. Since only functionality from os.path is needed I wanted to be specific.

tools/openms/MSstatsConverter.xml
```diff
@@ -24,7 +24,8 @@ mkdir out &&
 #if $adv_opts_cond.adv_opts_selector=='advanced':
   #if $adv_opts_cond.reannotate_filenames:
     mkdir adv_opts_cond.reannotate_filenames &&
-    ${ ' '.join(["ln -s '%s' 'adv_opts_cond.reannotate_filenames/%s.%s' &&" % (_, re.sub('[^\w\-_]', '_', _.element_identifier), $gxy2omsext(_.ext)) for _ in $adv_opts_cond.reannotate_filenames if _]) }
+    mkdir ${' '.join(["'adv_opts_cond.reannotate_filenames/%s'" % (i) for i, f in enumerate($adv_opts_cond.reannotate_filenames) if f])} && 
+    ${' '.join(["ln -s '%s' 'adv_opts_cond.reannotate_filenames/%s/%s.%s' && " % (f, i, re.sub('[^\w\-_]', '_', f.element_identifier), $gxy2omsext(f.ext)) for i, f in enumerate($adv_opts_cond.reannotate_filenames) if f])}
```
bgruening, line 28: @bernt-matthias a very general question, should we maybe have `f.save_element_identifier` that is doing this inside Galaxy, s that we can avoid all this sed/replace stuff in the wrappers?

bernt-matthias: Yeah! This is used in so many tools. Started almost 2 years ago with this https://github.com/galaxyproject/galaxy/pull/7937 .. somehow got distracted. We should try to get this in 21.05.

bgruening: Oh gosh, complete forget about it ... :100:

bernt-matthias:
Hi @bgruening , do you think we can merge soon (and if you like selectively push at least ConsensusID to the TS)?

The latest test fails for the online mascot online adapter (maybe we should set this on the skip list .. which is not possible yet https://github.com/galaxyproject/planemo-ci-action/issues/13). I guess this is transient, but I could ask at the OpenMS gitter channel.

For an overview let me summarize the changes (The first can be critical - I need it in particular for ConsensusID. The last two are cosmetic):

1. The symlink for each input dataset is placed in an individual directory: `PARAMETER_NAME/INDEX`, where the `INDEX` is a unique integer. This avoids clashes of inputs with the same element identifier. Apart from ConsensusID I do not know if this is relevant in realistic use cases .. but who knows. 

The complete list of 35 affected tools

```
AccurateMassSearch.xml
AssayGeneratorMetabo.xml
CVInspector.xml
ConsensusID.xml
DecoyDatabase.xml
EICExtractor.xml
Epifany.xml
FeatureLinkerUnlabeled.xml
FeatureLinkerUnlabeledKD.xml
FeatureLinkerUnlabeledQT.xml
FileMerger.xml
GNPSExport.xml
IDMassAccuracy.xml
IDMerger.xml
MSSimulator.xml
MSstatsConverter.xml
MaRaClusterAdapter.xml
MapAlignerIdentification.xml
MapAlignerPoseClustering.xml
MapAlignerSpectrum.xml
MapAlignerTreeGuided.xml
OpenSwathAnalyzer.xml
OpenSwathChromatogramExtractor.xml
OpenSwathDIAPreScoring.xml
OpenSwathFeatureXMLToTSV.xml
OpenSwathRTNormalizer.xml
PSMFeatureExtractor.xml
PercolatorAdapter.xml
ProteinInference.xml
ProteinResolver.xml
QCMerger.xml
QualityControl.xml
SemanticValidator.xml
SpecLibSearcher.xml
SpectraSTSearchAdapter.xml
```

2. The change from `optional="true"` to `false` for many parameters. The reason is: 

OpenMS sets text, int, select, bool parameters that have a default as optional (required=False), the default value is set implicitly if no value is given. This is reasonable for the CLI because one certainly does not want the user to specify the default manually for all parameters. 

For Galaxy tools setting these parameters as required leads to the equivalent behavior. Assuming required is better because it makes the implicit setting of parameters more transparent to the user (in Galaxy the default would be prefilled in the form and at least one option needs to be selected). 

3. I added a token to the validator macro such that the output also shows which parameter violates the validator: `<expand macro="list_string_san" name="matrix"/>`

bgruening (approved):

bgruening:
That looks all pretty great! I would prefer if you only push ConsensusID to the TS if that is possible and wait for the other enhancements for a new OpenMS release. If possible.

bgruening:
Thanks a lot!!!

bernt-matthias:
> I would prefer if you only push ConsensusID to the TS

Will do.

bernt-matthias:
I cancelled the CI workflow and uploaded ConsensusID manuallyto the TS.

PR #556, opened 2021-02-10
fix link to thermorawfileparser

martenson:
it is case sensitive


bernt-matthias:
Hi @martenson I cherry picked your commit to there https://github.com/galaxyproteomics/tools-galaxyp/pull/560

Issue #544, opened 2021-01-21
Error running maxquant_mqpar on Galaxy Australia


cat-bro:
I’m not 100% sure, but I think that an error with maxquant_mqpar on Galaxy Australia is due to mqpar being used as an input name and an output name.
https://github.com/galaxyproteomics/tools-galaxyp/blob/master/tools/maxquant_mqpar/maxquant_mqpar.xml#L105
https://github.com/galaxyproteomics/tools-galaxyp/blob/master/tools/maxquant/macros.xml#L159

I cannot reproduce the error on Galaxy Europe.

The command looks like so:

```
ln -s '/mnt/user-data/003/422/dataset_3422791.dat' 'ABCDE.mzML' &&  python3 '/mnt/tools-indices/shed_tools/toolshed.g2.bx.psu.edu/repos/galaxyp/maxquant_mqpar/2d67fb758956/maxquant_mqpar/modify_mqpar.py' --infiles='ABCDE.mzML' --version=1.6.10.43 --num_threads=${GALAXY_SLOTS:-1} --substitution_rx='[^\w\-\s\.]' --fasta_files='/mnt/user-data/003/422/dataset_3422983.dat' --description_parse_rule='^>.*\|.*\|[^ ]+ (.*) OS.*$' --identifier_parse_rule='^>.*\|(.*)\|.*$' '/mnt/user-data/003/435/dataset_3435253.dat'  && maxquant mqpar.xml >> '/mnt/user-data/003/435/dataset_3435269.dat' && mv '/mnt/user-data/003/435/dataset_3435253.dat' combined/txt/mqpar.xml && tar -zcf '/mnt/user-data/003/435/dataset_3435268.dat' ./combined/txt
```

The input `‘$mqpar’` parameter ought to correspond to `'/mnt/user-data/003/435/dataset_3435253.dat'` in the above command, but this is not the input mqpar.xml file path, instead it is an empty file created at runtime, so an error is thrown from xml.etree trying to parse an empty file.

The correct input file is used only when mqpar.xml is excluded from the outputs.  The tool passes its test because the mqpar.xml output is not listed as an output for the test case.

The solution could be to either (a) rename the input parameter from mqpar to mqpar_input or (b) not have mqpar.xml as an output option.  I’ve tried (a) on a dev server and it works.

[...]

PR #559, opened 2021-02-11
fix maxquant_mqpar variable names

cat-bro:
For issue https://github.com/galaxyproteomics/tools-galaxyp/issues/544

maxquant_mqpar jobs are failing when mqpar is included in the output because mqpar is also an input variable.

The first commit adds mqpar to the test output and breaks the test. The second commit updates the name of the input variable so that it's different from the output variable and test passes.


cat-bro:
This fails linting but the files are in the maxquant directory, not the maxquant_mqpar directory.  Should they be linted as part of this PR?

bernt-matthias:
The linting errors are for the python scripts:

```
tools/maxquant_mqpar/mqparam.py:9:1: I201 Missing newline between import groups. 'import yaml' is identified as Third Party and 'import re' is identified as Stdlib.
tools/maxquant_mqpar/mqparam.py:10:1: I100 Import statements are in the wrong order. 'import xml.etree.ElementTree' should be before 'import yaml' and in a different group.
tools/maxquant_mqpar/mqparam.py:10:1: I201 Missing newline between import groups. 'import xml.etree.ElementTree' is identified as Stdlib and 'import yaml' is identified as Third Party.
```

you just need to fix the import order.

bernt-matthias (approved):

bgruening:
Thanks @cat-bro and @bernt-matthias for the review!

Issue #558, opened 2021-02-10
Run thermorawfileparser in parallel on collections


martenson:
The current version puts all input files in a folder and invokes a single instance of the converter binary, which processes the files from the folder sequentially. This is not optimal for larger batches and could be heavily improved with parallelism.

local issue xref: https://github.com/RECETOX/galaxytools/issues/39

bernt-matthias:
@martenson after our discussion at the RECETOX issue I also checked here. I guess the reason is that the ThermoRawFileParser also gets metadata (txt or json). So the main question seems to be if this can be dropped and metadata can be derived afterwards.

Issue #557, opened 2021-02-10
thermorawfileparser successfully creates empty collections


martenson:
We run the tool at version `(Galaxy Version 1.2.3+galaxy0)` and for some inputs, often instrumental quality control but also some common runs, the tool outputs an empty collection without yielding any error. What could be causing this? Is this expected for possibly invalid inputs?

example history: https://umsa.cerit-sc.cz/u/recebot_converter/h/manual-test

![Galaxy](https://user-images.githubusercontent.com/1814954/107519320-710c9c00-6bb0-11eb-85c8-3932cf2e6484.png)

bernt-matthias:
@martenson Can you enable `Output metadata`? Then you will also get a txt/json file from which you can access stdout and stderr. Maybe this sheds some light.

martenson:
I tried all option combinations. It ends the same. The screenshot is with included metadata as json, that's why there are two collections made (one is for the metadata).

bgruening:
@martenson do you have access to the cluster? Can you have a look at the jwd? If not can you reproduce that at EU?

bernt-matthias:
Error message is `2021-02-11 10:24:23 ERROR RAW file still being acquired - ./raws_folder/19_meox_tms_NP100.raw` whatever this means. 

I guess we should extend stdio error detection.

PR #560, opened 2021-02-11
make ThermoRawFileParser run parallel

bernt-matthias:
and 

- add stdio (script returns always exit code 0)
- fix citation

fixes https://github.com/galaxyproteomics/tools-galaxyp/issues/558
should fix https://github.com/galaxyproteomics/tools-galaxyp/issues/557

we could also use the opportunity and update to 1.3.2 ... but I do not have the time to bump the bioconda recipe.


martenson:
We could postpone merging this after https://github.com/bioconda/bioconda-recipes/pull/26729 - the bioconda build is passing now.

bernt-matthias:
Yep. Lets do this. Can you check here https://github.com/compomics/ThermoRawFileParser/blob/master/README.md if there are new parameters that we should add?

martenson:
Maybe the exception data? 

```diff
--- from_file
+++ ThermoRawFileParser.exe --help
@@ -28,10 +28,17 @@
   -z, --noZlibCompression    Don't use zlib compression for the m/z ratios and
                                intensities. By default zlib compression is
                                enabled.
+  -a, --allDetectors         Extract additional detector data: UV/PDA etc
   -l, --logging=VALUE        Optional logging level: 0 for silent, 1 for
                                verbose.
   -e, --ignoreInstrumentErrors
                              Ignore missing properties by the instrument.
+  -x, --includeExceptionData Include reference and exception data
+  -L, --msLevel=VALUE        Select MS levels (MS1, MS2, etc) included in the
+                               output, should be a comma-separated list of
+                               integers ( 1,2,3 ) and/or intervals ( 1-3 ),
+                               open-end intervals ( 1- ) are allowed
+  -P, --mgfPrecursor         Include precursor scan number in MGF file TITLE
   -u, --s3_url[=VALUE]       Optional property to write directly the data into
                                S3 Storage.
   -k, --s3_accesskeyid[=VALUE]
```

martenson:
1.3.2 is ready https://anaconda.org/bioconda/thermorawfileparser/files

bernt-matthias:
@martenson the last test should fail (parquet generation) .. and succeed with https://github.com/bioconda/bioconda-recipes/pull/26756 (I tested this manually as described here https://planemo.readthedocs.io/en/latest/writing_how_do_i.html#test-tools-against-a-package-or-container-in-a-bioconda-pull-request)

bernt-matthias:
Tests do not run at all since we need release_21.01 (for parquet) which would come with https://github.com/galaxyproteomics/tools-galaxyp/pull/561

But we can discuss if we want such a recent profile or if its preferable to omit parquet for now.

Also I'm not 100% sure about the output format specific options which I implemented in the conditional.

tools/ThermoRawFileParser/thermo_converter.xml
```diff
@@ -1,144 +1,201 @@
...
-            label="Ignore missing instrument properties" help="If false, it stops the conversion if instrument properties are missing" />
-
-        <param name="output_metadata_selector" type="select" label="Output metadata" >
+        <param argument="--input" name="input_file" type="data" format="thermo.raw" label="Thermo RAW file" help="" optional="False"/>
+        <conditional name="format_cond">
+            <param argument="--format" name="output_format" type="select" label="Output format">
+                <option value="0">mgf</option>
+                <option value="1" selected="True">mzml</option>
+                <option value="2">Indexed mzml</option>
+                <option value="3">Parquet</option>
```
bgruening, line 50: Is there a more specific name for it? Parquet is like hdf5, a general concept/format.

bernt-matthias: Readme says _Apache Parquet: under development_ .. see also https://github.com/compomics/ThermoRawFileParser/issues/19 :) .. so maybe we skip this and choose profile 20.05 or 20.09?

martenson:
xref https://github.com/compomics/ThermoRawFileParser/issues/105

tools/ThermoRawFileParser/thermo_converter.xml
martenson: @bernt-matthias That seems like the better choice imho, wrapping unfinished features is tricky.

martenson:
The `tools/ThermoRawFileParser/test-data/really_small.parquet` file displays as having 0 bytes to me.

tools/ThermoRawFileParser/thermo_converter.xml
bernt-matthias: I disabled parquet output for now.

martenson (approved):

tools/ThermoRawFileParser/thermo_converter.xml
```diff
@@ -1,145 +1,202 @@
-<tool id="thermo_raw_file_converter" name="Thermo" version="1.2.3+galaxy0">
+<tool id="thermo_raw_file_converter" name="Thermo" version="@TOOL_VERSION@+galaxy0" profile="21.01">
```
martenson, line 1: Can we drop the profile a bit down then? Is there a required target?

bgruening: :+1:

tools/ThermoRawFileParser/thermo_converter.xml
```diff
@@ -1,145 +1,202 @@
...
+                <option value="0">mgf</option>
+                <option value="1" selected="True">mzml</option>
+                <option value="2">Indexed mzml</option>
+                <!-- <option value="3">Parquet</option> -->
+            </param>
+            <when value="0">
+                <param argument="--mgfPrecursor" type="boolean" truevalue="--mgfPrecursor" falsevalue="" label="Include precursor scan number in MGF file" help=""/>
+            </when>
+            <when value="1">
+                <param argument="--noZlibCompression" name="zlib_boolean" type="boolean" truevalue="" falsevalue="-z" checked="true"
```
bgruening, line 56: I'm wondering what this implies? Is there a use-case for our users to disable it?

tools/ThermoRawFileParser/thermo_converter.xml
```diff
@@ -1,145 +1,202 @@
...
-        <collection name="output_indexedmzml_collection" type="list" label="${tool.name} on ${on_string}: Indexed mzML collection">
-            <filter>output_format == "2"</filter>
-            <discover_datasets pattern="(?P&lt;designation&gt;.+)\.mzML" ext="mzml" directory="output_folder"/>
-        </collection>
-
-        <collection name="output_metadata_collection" type="list" label="${tool.name} on ${on_string}: metadata collection">
+        <data name="output" format="mzml" label="${tool.name} on ${on_string}">
+            <change_format>
+                <when input="format_cond.output_format" value="0" format="mgf"/>
+                <when input="format_cond.output_format" value="3" format="parquet"/>
```
bgruening, line 91: drop for now?

tools/ThermoRawFileParser/thermo_converter.xml
martenson: For future re-enabling the format they used seems to be implemented here: https://github.com/compomics/ThermoRawFileParser/blob/c2d4e16347262074c8c998768f306f1ed7043191/Writer/ParquetSpectrumWriter.cs#L134-L153

tools/ThermoRawFileParser/thermo_converter.xml
bernt-matthias: As far as I know different compression schemes can be used for numerical values. Maybe one or the other downstream tool is incompatible with the default ..

tools/ThermoRawFileParser/thermo_converter.xml
bernt-matthias: doen

tools/ThermoRawFileParser/thermo_converter.xml
bernt-matthias: > Can we drop the profile a bit down then?

Sure. I went for 20.05

> Is there a required target?

Not really.

bgruening (approved):

martenson:
Thanks a lot @bernt-matthias! And for the review @bgruening 👍

PR #562, opened 2021-02-13
fix openms test data workaround

bernt-matthias:
The workflow sync messed up the test data generation (needed for OpenMS). The workaround is to call test-data.sh for all repos (and not only those in the chunk) .. which makes no difference as long as OpenMS is the only tool using this. 

I guess it would be good if I could find time to work on something like this https://github.com/galaxyproject/tools-iuc/pull/2908


bgruening (approved):

bgruening:
@bernt-matthias how stable are the test file URLs. I would really like to enable Galaxy to upload test data from a URL.

This way we don't need any hacks and Galaxy could pull the datasets whenever we need it.

bernt-matthias:
The input file should be part of the openms release + 2 external sources https://github.com/galaxyproteomics/tools-galaxyp/blob/5e4f8567c0145de8c6f9344fe4ee4c3bf2a81e59/tools/openms/test-data.sh#L105  https://github.com/galaxyproteomics/tools-galaxyp/blob/5e4f8567c0145de8c6f9344fe4ee4c3bf2a81e59/tools/openms/test-data.sh#L113. But outputs and parameter files are re-computed https://github.com/galaxyproteomics/tools-galaxyp/blob/5e4f8567c0145de8c6f9344fe4ee4c3bf2a81e59/tools/openms/test-data.sh#L203 (the script is dynamically generated and outputs might differ with improvements to the generation an ctdopts/ctdconverter).

PR #569, opened 2021-03-10
Update MaxQuant

npinter:
This PR adds the "Match between runs" parameters as well as the quantification method for "Reporter ion MS3".


tools/maxquant/maxquant.xml
```diff
@@ -654,6 +875,29 @@ short peptides are usually not unique in the protein database and therefore not
...
+                <param name="files" value="BSA_min_22"/>
+                <param name="maxMissedCleavages" value="1"/>
+                <param name="variableModifications" value="Oxidation (M)" />
+                <conditional name="quant_method">
+                    <param name="select_quant_method" value="lfq" />
+                </conditional>
+            </repeat>
+            <param name="dry_run" value="True" />
+            <param name="output" value="config,mqpar" />
+            <output name="config" file="03/config.yml" lines_diff="9" />
```
bgruening, line 898: @npinter I think we should use asserts here very aggressively in addition.

tools/maxquant/maxquant.xml
```diff
@@ -632,6 +784,75 @@ short peptides are usually not unique in the protein database and therefore not
...
+                        <param name="filter_pif" value="True"/>
+                        <param name="reporter_pif" value="0.8"/> <!-- as filter_pif is True this value is used -->
+                    </conditional>
+                </conditional>
+            </repeat>
+            <param name="dry_run" value="True" />
+            <param name="output" value="config,mqpar,mzTab" />
+            <output name="config" file="02/config.yml" lines_diff="2" />
+            <!-- high difference due to unconsistant xml formatting in MQ -->
+            <output name="mqpar" file="02/mqpar.xml" lines_diff="24" />
```
bgruening, line 854: asserts could also work here

tools/maxquant/maxquant.xml
```diff
@@ -89,7 +89,16 @@
...
-            matchBetweenRuns: ${search_opts.match_between_runs}
+            #if $search_opts.mbr.match_between_runs == 'True':
+            matchBetweenRuns: ${search_opts.mbr.match_between_runs}
+            matchingTimeWindow: ${search_opts.mbr.matching_time_window}
+            matchingIonMobilityWindow: ${search_opts.mbr.matching_ion_mobility_window}
+            alignmentTimeWindow: ${search_opts.mbr.alignment_time_window}
+            alignmentIonMobilityWindow: ${search_opts.mbr.alignment_ion_mobility_window}
+            matchUnidentifiedFeatures: ${search_opts.mbr.match_unidentified_features}
+            #else:
+            matchBetweenRuns: ${search_opts.mbr.match_between_runs}
```
bgruening, line 100: why not taking it out of the if, as before?

npinter: `matchBetweenRuns: False` is missing in the config.yml file where it was before the update and therefore Planemo raises a lines_diff error.

bgruening: What I mean is that this line is both in the if **and** else.

npinter: True! Sorry, it's way too late.... ;)

foellmelanie:
Thank you so much @npinter!!! It looks good to me.

bgruening (approved):

bgruening:
Gracias!

bernt-matthias:
This PR makes maxquant_mqpar fail, because it uses the same macros, e.g. `NameMapper.NotFound: cannot find 'search_opts'` . This became apparent in https://github.com/galaxyproteomics/tools-galaxyp/pull/575

bgruening:
@bernt-matthias the TS was not updated for the mqpar wrapper: https://toolshed.g2.bx.psu.edu/view/galaxyp/maxquant_mqpar/9cb7dcc07dae

And this is like it should be. The symlinking is not optimal, we should move them in one folder so test gets not triggered when we update only one tool. But at least the second tool is not updated in the TS - so it can not break. What do I miss?

bernt-matthias:
TS is fine. Its just that `planemo test` fails for maxquant_mqpar since the change .. you can see this in the weekly CI and also in my PR. 

+1 for moving the tools in one dir .. or https://github.com/galaxyproject/planemo/issues/1150

PR #570, opened 2021-03-11
PS updated to 2.0.15 & SG updated to 4.0.22

CarlosHorro:


bernt-matthias:
Hey @CarlosHorro do you like to integrate my changes from here: https://github.com/galaxyproteomics/tools-galaxyp/pull/572? Or should be merge them separately?

For this PR it seems that only an assertion needs to be adapted.

CarlosHorro:
Hi @bernt-matthias We may integrate them here, how do I do that?

> Hey @CarlosHorro do you like to integrate my changes from here: #572? Or should be merge them separately?
> 
> For this PR it seems that only an assertion needs to be adapted.

bernt-matthias:
> We may integrate them here, how do I do that?

Should be simply: `git cherry-pick edef57be5ff2a0a1c93d3ee4858eab01516fa88c`

tools/peptideshaker/peptide_shaker.xml
```diff
@@ -12,8 +12,7 @@
     </requirements>
     <expand macro="stdio" />
 
-    <command><![CDATA[
-    ## When supporting more advanced Galaxy versions:    command use_shared_home="false"
+    <command use_shared_home="false"><![CDATA[
```
bernt-matthias, line 15: I think it would be better to increase the `profile` of the tools to something greater than or equal to 18.01. 

See https://github.com/galaxyproject/galaxy/blob/ce23f963e036722865ecd27fda509f173406ab61/lib/galaxy/tool_util/parser/xml.py#L169

I would suggest at least 20.01

CarlosHorro: Yes, you are right @bernt-matthias, thank you. That was something I wanted to do some time ago but I didn't remember to include it...

CarlosHorro:
Great, I think we have included enough stuff here :-)

bernt-matthias:
> Great, I think we have included enough stuff here :-)

I think we should append `log_folder/*.log ` to the stderr. See here https://github.com/compomics/searchgui/issues/288#issuecomment-798934981

bernt-matthias:
Maybe like this: If you include this in the first line of the shell script:

```
trap ">&2 cat log_folder/*.log" EXIT
```

CarlosHorro:
Um... @bernt-matthias  do you think it is a good idea to include (long) html reports into that output? may it be enough to include PeptideShaker.log for PeptideShaker executions and SearchGUI.log for SG?

> > Great, I think we have included enough stuff here :-)
> 
> I think we should append `log_folder/*.log ` to the stderr. See here [compomics/searchgui#288 (comment)](https://github.com/compomics/searchgui/issues/288#issuecomment-798934981)

CarlosHorro:
> Maybe like this: If you include this in the first line of the shell script:
> 
> ```
> trap ">&2 cat log_folder/*.log" EXIT;
> ```

Yep, it seems to work nicely, I will add it :-)

bernt-matthias:
One more thing: https://github.com/CarlosHorro/tools-galaxyp/pull/2

bernt-matthias:
Interesting new tool errors...

CarlosHorro:
> Interesting new tool errors...

Well, the error seems related to trying to write into user's folder:

_
java.lang.RuntimeException: java.io.FileNotFoundException: /usr/local/share/peptide-shaker-2.0.15-0/resources/conf/paths.txt (Permission denied)
	at com.compomics.util.io.flat.SimpleFileWriter.<init>(SimpleFileWriter.java:59)
	at eu.isas.peptideshaker.preferences.PeptideShakerPathParameters.writeConfigurationToFile(PeptideShakerPathParameters.java:391)
	at eu.isas.peptideshaker.cmd.PathSettingsCLI.setPathSettings(PathSettingsCLI.java:110)
	at eu.isas.peptideshaker.cmd.PathSettingsCLI.extractAndUpdatePathOptions(PathSettingsCLI.java:269)
	at eu.isas.peptideshaker.cmd.PeptideShakerCLI.main(PeptideShakerCLI.java:1366)
Caused by: java.io.FileNotFoundException: /usr/local/share/peptide-shaker-2.0.15-0/resources/conf/paths.txt (Permission denied)
	at java.base/java.io.FileOutputStream.open0(Native Method)
	at java.base/java.io.FileOutputStream.open(FileOutputStream.java:298)
	at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:237)
	at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:187)
	at java.base/java.io.FileWriter.<init>(FileWriter.java:96)
	at com.compomics.util.io.flat.SimpleFileWriter.<init>(SimpleFileWriter.java:53)
	... 4 more
_

But I don't understand why this happens now and not just right after I included "command use_shared_home="false" in PS...

bernt-matthias:
> But I don't understand why this happens now and not just right after I included "command use_shared_home="false" in PS...

Maybe because we did not see the exception until the trap has been included.

You can simply try to revert the profile and set `command use_shared_home="false"` again

CarlosHorro:
> > But I don't understand why this happens now and not just right after I included "command use_shared_home="false" in PS...
> 
> Maybe because we did not see the exception until the trap has been included.
> 
> You can simply try to revert the profile and set `command use_shared_home="false"` again

I will check if it is feasible to write that file in the temp folder. I will undo the profile otherwise...

CarlosHorro:
> > But I don't understand why this happens now and not just right after I included "command use_shared_home="false" in PS...
> 
> Maybe because we did not see the exception until the trap has been included.
> 
> You can simply try to revert the profile and set `command use_shared_home="false"` again

I've seen that this is not immediate to solve. There is even another problem :

```
java.io.FileNotFoundException: /tmp/tmprrb5ssna/job_working_directory/000/6/working/PeptideShakerCLI/exportFactory.json (No such file or directory)
	at java.base/java.io.FileInputStream.open0(Native Method)
	at java.base/java.io.FileInputStream.open(FileInputStream.java:219)
	at java.base/java.io.FileInputStream.<init>(FileInputStream.java:157)
	at com.compomics.util.io.json.JsonMarshaller.getJsonStringFromFile(JsonMarshaller.java:227)
	at com.compomics.util.io.json.JsonMarshaller.fromJson(JsonMarshaller.java:142)
	at eu.isas.peptideshaker.export.PSExportFactory.loadFromFile(PSExportFactory.java:158)
	at eu.isas.peptideshaker.export.PSExportFactory.getInstance(PSExportFactory.java:99)
	at eu.isas.peptideshaker.cmd.ReportCLIParams.<clinit>(ReportCLIParams.java:17)
	at eu.isas.peptideshaker.cmd.PeptideShakerCLIParams.createOptionsCLI(PeptideShakerCLIParams.java:76)
	at eu.isas.peptideshaker.cmd.PeptideShakerCLI.main(PeptideShakerCLI.java:1370)
```

So I will undo the last changes until these issues are solved in the app itself.

CarlosHorro:
So, I guess we can merge this like it is and afterwards when PS app itself manages the paths in a better way we can create another PR.

bernt-matthias:
> So, I guess we can merge this like it is and afterwards when PS app itself manages the paths in a better way we can create another PR.

Agreed. Was just thinking if we can have `profile="20.01"` for fasta_cli and ident_params?

Have you opened an issue at the PS repo?

bernt-matthias:
I still do not understand how the two error relate to the use of a shared home / a more recent profile.

Or are the two exceptions non-critical? Then we could be more specific in the stdio regular expressions...

CarlosHorro:
> > So, I guess we can merge this like it is and afterwards when PS app itself manages the paths in a better way we can create another PR.
> 
> Agreed. Was just thinking if we can have `profile="20.01"` for fasta_cli and ident_params?

Trying it...

> 
> Have you opened an issue at the PS repo?

Yes, just did it:
https://github.com/compomics/peptide-shaker/issues/445

CarlosHorro:
> I still do not understand how the two error relate to the use of a shared home / a more recent profile.
> 
> Or are the two exceptions non-critical? Then we could be more specific in the stdio regular expressions...

Not sure either; however, 1) they don't look critical as PS have been working well without taking care of them, and 2) I think we need to move this forward and manage that all in the future PR.

bernt-matthias:
> I think we need to move this forward and manage that all in the future PR.

Agreed. But please also enable the newer profile for peptide_shaker. The profile is not the source of the problem, but the trap which appends the exception traces from the log files to stderr which is then caught by our stdio regular expressions, in particular https://github.com/galaxyproteomics/tools-galaxyp/blob/bbe7a74e22521fb8d0c395368d3a91ff26955dc8/tools/peptideshaker/macros_basic.xml#L6 

I already verified this locally.

bernt-matthias (approved):
I will merge as soon as the CI passes

PR #571, opened 2021-03-11
Fix MSstatsTMT

npinter:
Selecting "allonly" at "Select protein IDs to draw plots" in the "Plot Output Options" together with "ProfilePlot" as an output, crashes the R script. This PR fixes this by setting the default value for "which.Protein" (="all"), when the above is being selected. A test for this case was added.


foellmelanie:
looks good :)

bgruening (approved):

PR #572, opened 2021-03-12
some fixes for searchgui and peptideshaker

bernt-matthias:
1. remove stderr redirection

in case of

```
command 2>> $temp_stderr &&
.... &&
cat $temp_stderr 2>&1;
```

stderr is swallowed if command fails.
this was the case for PS, FC, and IP

in case of

```
command 2>> $temp_stderr &&
.... ;
cat $temp_stderr 2>&1;
```

the last cat is executed even if the command fails.
the exit code of the command line is then the exit
code of the cat, i.e. the tool will always be counted
successful, even if everything fails.

2. fix peptideshaker find exec constructs

Galaxy removes the last ';' from the generated command line.
thus we need to make sure that the last `find ... -exec ... \;`
is not the last command.

3. remove some unnecessay parentheses

TODO

- [ ] make IP and FC versions equal to SG version (except for tool version suffix `+galaxyX`)?

### DIFF
--- a/tools/peptideshaker/fasta_cli.xml
+++ b/tools/peptideshaker/fasta_cli.xml
@@ -11,7 +11,6 @@
     <expand macro="stdio" />
     <command>
 <![CDATA[
-        #set $temp_stderr = "fasta_cli_stderr"
         #set $output_db_name = $input_database.display_name.replace(".fasta", "") + $database_processing_options.decoy_file_tag.value + ".fasta"
 
         mkdir output &&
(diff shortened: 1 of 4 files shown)


bernt-matthias:
Wonder why the test is failing .. was running locally. 

On the positive side it seems to be the same error as in https://github.com/galaxyproteomics/tools-galaxyp/pull/570 that should be fixed easily.

bernt-matthias:
Closed in favor of https://github.com/galaxyproteomics/tools-galaxyp/pull/570

PR #573, opened 2021-03-13
update pepquery to v1.6.2

jj-umn:
Add support for -tag option for immunopeptidomics


jj-umn:
The pepquery bioconda recipe uses the same shell script to exec the pepquery  java application that the varscan tool uses. 
This is running into a locale issue:
/usr/local/bin/pepquery: line 8: warning: setlocale: **LC_ALL**: cannot change locale (en_US.UTF-8): No such file or directory

tools/pepquery/pepquery.xml
```diff
@@ -143,39 +155,63 @@
...
+                </param>
+                <when value="None"/>
+                <when value="PepQuery">
+                    <param name="tag_file" argument="-tag" type="data" format="txt" label="PepPuery Tag File"/>
+                    <param name="tagIndexType" argument="-tagIndexType" type="select" label="Tag File Index Type" help="pFind uses the spectrum title" >
+                        <option value="1">index (1-based) in MGF</option>
+                        <option value="2" selected="true">spectrum title in MGF</option>
+                    </param>
+                </when>
+                <when value="pFind">
```
bgruening, line 178: What is this pFind?

tools/pepquery/pepquery.xml
```diff
@@ -19,6 +19,11 @@
             #set $db_file = $re.sub('\s','_',$re.sub('[.][^.]*$','',$req_inputs.db_file.display_name.split('/')[-1])) + ".fa"
             ln -s '$req_inputs.spectrum_file' '$spectrum_file' &&
             ln -s '$req_inputs.db_file' '$db_file' &&
+            #if $req_inputs.tags.tagType == 'pFind':
+                pepquery main.java.util.generateTags $req_inputs.tags.qry_res > qry.tags &&
+                ls -l $req_inputs.tags.qry_res qry.tags &&
```
bgruening: is that ls and cat still needed?

tools/pepquery/pepquery.xml
jj-umn: @bgruening   http://pepquery.org/data/PepQuery_for_immunopeptidomics_data.pdf 
Open-pFind can be downloaded from http://pfind.ict.ac.cn/software/pFind3/. It can only be run on Windows system.  
pFind generates peptide tags that can be used to greatly reduce the search space for pepquery when searching without an enzyme cleavage.   Someone suggested that MsFragger might also help in that regard, but I don't have the knowledge to evaluate that proposition.

tools/pepquery/pepquery.xml
jj-umn: @bgruening   No.  That was just debugging.  I'll get that removed. Thanks.

jj-umn:
@bgruening  A little more background about our need for pFind.   We're working on a cancer vaccine project.   We'll be trying to verify predicted HLA presented neoantigens by using Mass Spec.   Since the HLA bound peptide won't be subjected to enzyme digestion, it is very computationally intense to search the Mass Spec data.

bgruening:
Sounds awesome! Please keep us in the loop! :)

PR #577, opened 2021-03-19
update searchgui, peptideshaker and thermorawfileparser

bernt-matthias:
- [x]  update SG, PS, and TRFP
- [x] bind fasta_cli and ident_param versions to searchgui: in the past updates the versions of FC and IP did not change, which seemed reasonable because the tools did "not change". But actually the requirement (SG) was updated. This may result in multiple installations of FC and IP with different
requirements. When installed successively the problem is that the dependency reolution can only resolve one of the envs so at least one tool won't have a env and will fail therefore.
- [x] improve SG/PS help https://github.com/compomics/peptide-shaker/issues/447
- [ ] is there a better way to check the trfp outputs, e.g. assertions


bernt-matthias:
Hey @CarlosHorro would you like to review this?

CarlosHorro:
Hi @bernt-matthias Everything looks good, but I'm not sure about why has been necessary to update test files...?

bernt-matthias:
> I'm not sure about why has been necessary to update test files...?

The mzml files for instance contain some version numbers which changed also some rounding problems seem to be fixed now. 

For tabular files one column was added + some small numerical differences.

Was just wondering if there is a better way to test the outputs, e.g. by using assertions.

martenson (approved):
The Thermoparser version update is much appreciated. Thank you. :+1:

The tests pass for me locally.

bernt-matthias:
Is this ready to go from your side @CarlosHorro and @martenson ?

martenson:
Looks good to me, would it be useful if I reran the tests?

bernt-matthias:
> Looks good to me, would it be useful if I reran the tests?
> editL they pass on MacOS

Excellent

bgruening (approved):

martenson (approved):

PR #578, opened 2021-03-25
ThermoRawFileParser: try to fix reading error

bernt-matthias:
```
2021-03-25 19:35:42 ERROR Error opening (ThermoFisher.CommonCore.RawFileReader.Facade.RawFileLoader) - 180714_NJ_Bastida_ALP176.raw
```

Apparently stopped working with https://github.com/bioconda/bioconda-recipes/pull/26756

### DIFF
--- a/tools/ThermoRawFileParser/thermo_converter.xml
+++ b/tools/ThermoRawFileParser/thermo_converter.xml
@@ -1,4 +1,4 @@
-<tool id="thermo_raw_file_converter" name="Thermo" version="@TOOL_VERSION@+galaxy0" profile="20.05">
+<tool id="thermo_raw_file_converter" name="Thermo" version="@TOOL_VERSION@+galaxy1" profile="20.05">
     <description>RAW file converter</description>
     <macros>
         <token name="@TOOL_VERSION@">1.3.2</token>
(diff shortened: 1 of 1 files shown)


CarlosHorro:
> Ping @CarlosHorro do you have any idea why symlinking fails?
> 
> * 1st commit shows the problem
> * 2nd "solves" it

Hi @bernt-matthias The stacktrace is not very detailed so I don't know (permission issue on the source file?), but I don't understand 1) why ln was previously working and now it doesn't, and 2) why ln -s -f / ln -f -s made a difference here...

bernt-matthias:
> The stacktrace is not very detailed so I don't know (permission issue on the source file?), 

Unlikely in a planemo run. It fails with docker and conda.

> but I don't understand 1) why ln was previously working and now it doesn't,

My best guess is the change in the conda recipe. 

>  and 2) why ln -s -f / ln -f -s made a difference here...

It doesn't. This is just an artificial change .. because I was playing around a bit.

bernt-matthias:
Apparently its due to `mono 6`: https://github.com/compomics/ThermoRawFileParser/issues/108#issuecomment-793079885

In the bioconda recipe we enforce `mono >=5` https://github.com/bernt-matthias/bioconda-recipes/blob/1184915b6296b26dcb7a202c199fce02ec080c94/recipes/thermorawfileparser/meta.yaml#L28

I guess in the past this was resolved to mono 5 and now to mono 6.

bernt-matthias:
With https://github.com/bioconda/bioconda-recipes/pull/27618 it seems to work again. I also checked with conda .. .

So we can close this one.

PR #584, opened 2021-04-22
Bump SearchGUI and PeptideShaker

bernt-matthias:
depends on https://github.com/bioconda/bioconda-recipes/pull/28155

- [x] I will tests this locally before merging the bioconda recipe
- [ ] we could enable metamorpheus https://github.com/galaxyproteomics/tools-galaxyp/blob/d8d9887e20120bb93f49de9e3460f95d385575bb/tools/peptideshaker/searchgui.xml#L190 but I would suggest to install it via a requirement in the recipe and the remove the packaged MM

### DIFF
--- a/tools/peptideshaker/fasta_cli.xml
+++ b/tools/peptideshaker/fasta_cli.xml
@@ -28,7 +28,15 @@
         ###########################################
 
         echo 'Creating decoy database.' &&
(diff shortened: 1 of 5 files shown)


bernt-matthias:
Closed in favor of https://github.com/galaxyproteomics/tools-galaxyp/pull/594

PR #585, opened 2021-05-02
New eggnog release

bgruening:
grafik

This seems to be a confusing release as well. Now version `Version 5.0.2 of the eggnog.db sqlite DB.` is needed?

Just putting out this PR, to see what happens and if anyone is interested to work on it.


abretaud:
FYI I'm working on updating to 2.1.6

PR #602, opened 2021-06-01
[WIP] id converter proteore tool

combesf:
Proteore id_converter tool 
version 20210601
for review, thank you


combesf:
I still have an error in the tool test here. 
The error is : 
`parameter 'mapping_file': requires a value, but no legal values defined`

But in the xml file, in the test lines, mapping_file is affected, so I don't understand ... 

If someone could help here please ??? 
thansk a lot

bgruening:
@combesf can you have a look at this tool: https://github.com/bgruening/galaxytools/tree/master/tools/general_mapping

Its more generic I think and maybe you can use that tool instead?

yvandenb:
>can you have a look at this tool: https://github.com/bgruening/galaxytools/tree/master/tools/general_mapping 
@bgruening : not sure that these tools are exactly the same; btw I was unable to find it in the TS so that we can install it on our dev-instance for evaluation purpose, could provide us with the exact name or a link ? is this tool deployed in usegalaxy.eu ?

foellmelanie:
Hi @yvandenb I am not sure where the general mapping tool is but I think the 'uniprot_rest_interface' Uniprot tool can do the ID conversions for many different IDs and organisms: https://toolshed.g2.bx.psu.edu/repository?repository_id=c8774310981b07c5 It's also installed on the EU server.

bgruening:
@combesf, @yvandenb  I never published the tool in the toolshed, but at this time I wanted to solve the exact same problem as you, but in a more general way: https://github.com/bgruening/galaxytools/blob/master/tools/general_mapping/readme.md

In essence, an admin can create config files, of arbitrary mapping files. One tool that can map IDs for proteomics and physics and so on...

PR #604, opened 2021-06-14
Fix for crash of MaxQuant wrapper when using TMT data and PTXQC report

npinter:
This PR fixes the PTXQC report generation when using TMT input data. In addition, PTXQC was updated to the newest version and isobaric labels were moved into tokens. A test for PTXQC was added.


bgruening (approved):

bgruening:
Merci!

PR #605, opened 2021-06-15
Fix MS level filter

chambm:
The option parser has a bug if there's a space in the MS level range, so `[2, 3]` is interpreted as only MS level 3, but `[2,3]` is levels 2 and 3. This is technically a bug in msconvert but it's easier to fix here.


chambm:
Being able to update test files with CI would be a big time saver.

PR #609, opened 2021-06-20
diffacto

jj-umn:
Initial tool for Diffacto.
Diffacto quantifies comparative protein abundance from the covariation of peptide abundance.
Diffacto applies factor analysis to extract the covariation of peptides' abundances. The method enables a weighted geometrical average summarization and automatic elimination of incoherent peptides, which may result from suboptimal digestion or being partially modified, and are not representative of the protein concentration.


tools/diffacto/diffacto.xml
```diff
@@ -0,0 +1,230 @@
...
+        -out '$output'
+        #if $mcfdr
+            -mc_out '$mc_out'
+        #end if
+        #if $loadings
+            -loadings_out '$loadings_out'
+        #end if
+    ]]></command>
+    <inputs>
+        <param name="input" argument="-i" type="data" format="csv" label="Peptides abundances">
```
bgruening, line 47: where is this file coming from? Can we make it a TSV?

tools/diffacto/diffacto.xml
```diff
@@ -0,0 +1,230 @@
...
+                The first column should contain unique peptide sequences. 
+                Missing values should be empty instead of zeros.
+            ]]></help>
+        </param>
+        <param argument="-db" type="data" format="fasta" label="Protein database" optional="true"
+               help="Optional if the peptide file has protein IDs in the second column"/> 
+        <param argument="-samples" type="data" format="tabular" label="Sample list" optional="true">
+            <help><![CDATA[
+               One run and its sample group per line, separated by tab. 
+               If None, read from peptide file headings,
```
bgruening, line 60: which header?

tools/diffacto/diffacto.xml
```diff
@@ -0,0 +1,230 @@
...
+        <param argument="-db" type="data" format="fasta" label="Protein database" optional="true"
+               help="Optional if the peptide file has protein IDs in the second column"/> 
+        <param argument="-samples" type="data" format="tabular" label="Sample list" optional="true">
+            <help><![CDATA[
+               One run and its sample group per line, separated by tab. 
+               If None, read from peptide file headings, 
+               then each run will be summarized as a group.
+            ]]></help>
+        </param>
+        <param argument="-reference" type="text" value="" label="reference sample groups"
```
bgruening, line 64: do we need to add a validator?

tools/diffacto/diffacto.xml
```diff
@@ -0,0 +1,230 @@
...
+        <param argument="-cutoff_weight" type="float" value="0.5" min="0." max="1.0" label="Peptide cutoff weight"
+            help="Peptides weighted lower than the cutoff will be excluded."/>
+        <param argument="-use_unique" type="boolean" truevalue="-use_unique True" falsevalue="" checked="false" label="Use unique peptides only"/>
+        <param argument="-fast" type="boolean" truevalue="-fast True" falsevalue="" checked="false" label="Allow early termination in EM calculation when noise is sufficiently small."/>
+        <param name="mcfdr" argument="-mc_out" type="boolean" truevalue="True" falsevalue="False" checked="false" label="Perform Monte Carlo FDR simulation"/>
+        <param name="loadings" argument="-loadings_out" type="boolean" truevalue="True" falsevalue="False" checked="false" label="Output Protein Peptide loadings file"/>
+    </inputs>
+    <outputs>
+        <data name="output" format="tabular" label="${input.name.replace('.csv','')}.tsv">
+            <actions>
```
bgruening, line 93: Oh this is gold! We should add this to IUC as well to as many tool as possible. DESEQ and EDGER for example.

tools/diffacto/diffacto.xml
```diff
@@ -0,0 +1,230 @@
...
+                <action name="comment_lines" type="metadata" default="1" />
+                <action name="column_names" type="metadata" default="Protein,N.Pept,Q.Pept,S/N,P(PECA)" />
+            </actions>
+        </data>
+        <data name="mc_out" format="tabular" label="${input.name.replace('.csv','')}.FDR">
```
bgruening, line 98: users can change that name as they like.

What about: `${tool.name} on ${on_string}: FDR` or `${tool.name} on ${on_string}: Tabular` above.

tools/diffacto/diffacto.xml
jj-umn: @bgruening   pjagtap@umn.edu asked for this tool.   I don't have a sample workflow from him at this time.

diffacto is using pandas.read_csv() for this input.   I tried using  a tabular input, but got an error.  
However, galaxy was able to autoconvert a tabular to csv for the input, so I think that will work for most cases.  
I will add a comment in the help about the auto conversion.

bgruening: could we just do something like `tr '\t' ',' < file.tsv > file.csv` as a first step and avoid this extra job?

jj-umn: @bgruening  I incorporated your suggestion

tools/diffacto/diffacto.xml
jj-umn: I improved the help text, so I hope that makes this more clear.

tools/diffacto/diffacto.xml
jj-umn: @bgruening I couldn't think of a way to incorporate validators for the possible combinations of inputs.  I hope the expanded help information will guide the user.

tools/diffacto/diffacto.xml
jj-umn: @bgruening I reverted to the standard method you suggested.

bgruening (commented):
.shed.yml file is missing I think.

bgruening (approved):

PR #613, opened 2021-06-30
Add ProteomIQon PeptideDB

ZimmerA:
The tool ProteomIQon PeptideDB creates a peptide database in the SQLite format.


tools/proteomiqon_peptidedb/proteomiqon_peptidedb.xml
```diff
@@ -0,0 +1,206 @@
+<tool id="proteomiqon_peptidedb" name="ProteomIQon PeptideDB" version="@VERSION@" python_template_version="20.05">
+    <description>
+		creates a peptide database in the SQLite format.
+	</description>
+    <macros>
+		<token name="@VERSION@">0.0.5</token>
```
bgruening, line 6: indentaion, please use 4 spaces everywhere instead of a tab

tools/proteomiqon_peptidedb/proteomiqon_peptidedb.xml
```diff
@@ -0,0 +1,206 @@
...
+            #end for
+            ],
+            "VarModThreshold": ${VarModThreshold}
+            }
+            ]]>
+        </configfile>
+    </configfiles>
+    <inputs>
+        <param name="fasta" type="data" format="fasta" label="Fasta file" help="Please specify your protein sequences in the .fasta format. If you want to search for contaminants you have to include the sequences in your file. You do not have to supply decoy sequences, those are generated for you."/>
+        <param name="Name" type="text" value="" label="Database name" />
```
bgruening, line 71: is that really needed? can we not use a fixed one? the user will not see it? Something like galaxy.db?

tools/proteomiqon_peptidedb/proteomiqon_peptidedb.xml
```diff
@@ -0,0 +1,206 @@
...
+        <param name="ParseProteinIDRegexPattern" type="text" value="id" label="Regex pattern to parse protein ID from .fasta header" help="Fasta headers do often contain additional information in addition to your protein identifier, by specifying a regex pattern the tool can extract the protein IDs. If you fasta headers are already cleaned you can leave this field empty.">
+        <sanitizer sanitize="false" />
```
bgruening, line 73: this can be dangerous, it means you allow the user to maybe break out of the CMD or add some wired stuff to your config file.

Maybe you can use that sanatizer? https://github.com/galaxyproteomics/tools-galaxyp/blob/64740168b7816f0fe38f6f0629cb7e618cd42878/tools/regex_find_replace/regex_tabular.xml#L21

tools/proteomiqon_peptidedb/proteomiqon_peptidedb.xml
```diff
@@ -0,0 +1,206 @@
...
+        </param>
+        <param name="Protease" type="select" label="Protease" help="Select a protease the tool can use when performing in silico digestion.">
+            <option value="Trypsin">Trypsin</option>
```
bgruening, line 76: A select with only one option?

tools/proteomiqon_peptidedb/proteomiqon_peptidedb.xml
```diff
@@ -0,0 +1,206 @@
...
+        <param name="MaxMissedCleavages" type="integer" value="2" label="Max missed cleavages" help="Select the maximum amount of miss cleavages in a peptide sequence."/>
+        <param name="MaxMass" type="float" value="15000.0" label="Max mass" help="Select the maximum mass of peptides included in the data base."/>
+        <param name="MinPepLength" type="integer" value="4" label="Min peptide length" help="Select the minimum amino acid  length of peptides included in the data base."/>
+        <param name="MaxPepLength" type="integer" value="65" label="Max peptide length" help="Select the minimum amino acid  length of peptides included in the data base."/>
+        <param name="IsotopicMod" type="select" label="Isotopic mod" help="Select isotopic amino acid modifications. These modifications change the natural abundance of isotopes of single elements. Each peptide sequence is duplicated and included in the data base at natural isotopic abundances and using the modified isotopic abundances." multiple="true">
+            <option value="N15" selected="true">N15</option>
```
bgruening, line 84: only one option?

tools/proteomiqon_peptidedb/proteomiqon_peptidedb.xml
```diff
@@ -0,0 +1,206 @@
...
+------------
+The PeptideDB tool helps to create peptide databases by in silico digestion given proteome information in the FASTA format and a set of parameters 
+that allow the user to mimic conditions of their specific experiment. 
+The created database stores peptide protein relationships in a SQLite database which can then be supplied to other ProteomIQon tools.
+
+Further Reading
+---------------
+Additional information about the tool can be found in the `documentation <https://csbiology.github.io/ProteomIQon/tools/PeptideDB.html>`_.
+    ]]>
+    </help>
```
bgruening, line 205: add ciations?

tools/proteomiqon_peptidedb/proteomiqon_peptidedb.xml
bgruening: btw. maybe you want to offer some standard regex? as a select?

tools/proteomiqon_peptidedb/proteomiqon_peptidedb.xml
```diff
@@ -0,0 +1,206 @@
+<tool id="proteomiqon_peptidedb" name="ProteomIQon PeptideDB" version="@VERSION@" python_template_version="20.05">
```
bgruening, line 1: remove python_template_version and add profile="20.05"

tools/proteomiqon_peptidedb/proteomiqon_peptidedb.xml
ZimmerA: Indeed, will change with next commits!

tools/proteomiqon_peptidedb/proteomiqon_peptidedb.xml
ZimmerA: Oh, I didn't know that! I will try this one out.

ZimmerA: The sanitizer you provided escapes backslashes and apostrophes. Is it possible that we remove the escaping of these characters? We don't use this parameter in the command tag. Just the config file. It should be fine to have those characters there, since it's just a JSON file that gets parsed by our tool?

bgruening: In this case leave your solution I don't think I have a better idea. And let's hope it can not be exploited.

tools/proteomiqon_peptidedb/proteomiqon_peptidedb.xml
ZimmerA: Added more options.

tools/proteomiqon_peptidedb/proteomiqon_peptidedb.xml
ZimmerA: Added more options.

tools/proteomiqon_peptidedb/proteomiqon_peptidedb.xml
ZimmerA: Is in preparation and will be added upon publication.

bgruening (approved):

bgruening:
Cool danke!

PR #622, opened 2021-07-23
Fix image not showing in proteomiqon_peptidespectrummatching

ZimmerA:
This PR attempts to fix the image not showing in the help section of the Proteomiqon PeptideSpectrumMatching tool, found here: https://usegalaxy.eu/root?tool_id=proteomiqon_peptidespectrummatching

To solve the issue, I took a look at https://github.com/galaxyproject/tools-iuc/blob/master/tools/bedtools/slopBed.xml and saw that they prefixed images with $PATH_TO_IMAGES/ so I followed that practice.

If this is not the cause of the image not showing up, please let me know.


bgruening:
I'm not sure this will fix it on EU. We seem to have a general problem with those images on our deployment ... we will investigate that as soon as possible.

PR #623, opened 2021-07-24
Add ProteomIQon ProteinInference

ZimmerA:
MS-based shotgun proteomics estimates protein abundances using a proxy: peptides. The process of 'ProteinInference' is concerned with the mapping of identified peptides to the proteins they putatively originated from.


tools/proteomiqon_proteininference/proteomiqon_proteininference.xml
```diff
@@ -0,0 +1,267 @@
+<tool id="proteomiqon_proteininference" name="ProteomIQon ProteinInference" version="@VERSION@" profile="20.05">
+    <description>
+        uses a list of identified peptides to deduce proteins that explain their presence in the sample.
```
bgruening, line 3: can you make this more crisp/shorter?

tools/proteomiqon_proteininference/proteomiqon_proteininference.xml
```diff
@@ -0,0 +1,267 @@
...
+                    "Case": "${GetQValueCond.GetQValue}",
+                }
+                #end if
+            }
+            ]]>
+        </configfile>
+    </configfiles>
+    <inputs>
+        <param name="peptideDB" type="data" format="sqlite" label="Peptide database" help="Specify the peptide data base."/>
+        <conditional name="PSMInputModeCond">
```
bgruening, line 60: why is this difference important?

a multiple=true datatype can also just have one?

tools/proteomiqon_proteininference/proteomiqon_proteininference.xml
```diff
@@ -0,0 +1,267 @@
...
+                <option value="LogisticRegression">Logistic Regression</option>
+                <option value="NoQValue">NoQValue</option>
+            </param>
+            <when value="LogisticRegression">
+                <param name="LogisticRegressionType" type="select" label="Logistic regression type" help="">
+                    <option value="Conservative">Conservative</option>
+                    <option value="MAYU" selected="true">MAYU</option>
+                    <option value="DecoyTargetRatio">DecoyTargetRatio</option>
+                </param>
+            </when>
```
bgruening, line 97: the other "whens" are missing here

bgruening: mh, why is out linter not complaining, wired...

bgruening (commented):
Pretty cool, just a few small comments.

tools/proteomiqon_proteininference/proteomiqon_proteininference.xml
ZimmerA: multiple=false allows to start batch jobs from the UI without any extra setup, I liked the possibility of being able to do that.

tools/proteomiqon_proteininference/proteomiqon_proteininference.xml
ZimmerA: These are union cases, the other whens are not needed, it's just the LogisticRegression that follows up with extra possibilities.

tools/proteomiqon_proteininference/proteomiqon_proteininference.xml
ZimmerA: I'll take a look!

tools/proteomiqon_proteininference/proteomiqon_proteininference.xml
bgruening: Can you please explain that? You mean you would like to start N jobs for N inputs? If so what is the advantage of multiple=true?

tools/proteomiqon_proteininference/proteomiqon_proteininference.xml
bgruening: That is fine, just from a formal point of view we need `<when value="Storey"/>` and  `<when value="NoQValue"/>`

ZimmerA: Understood!

tools/proteomiqon_proteininference/proteomiqon_proteininference.xml
ZimmerA: From my understanding, in order to use multiple files for one job, I need multiple=true. 
But if I do so, I lose the ability to start batch jobs in the way I can do it with multiple=false?
So I did this to have maximum flexibility.

To start N jobs for N inputs, I probably need to use collections, or is there another way?

tools/proteomiqon_proteininference/proteomiqon_proteininference.xml
ZimmerA: How about: "uses identified peptides to deduce proteins explaining their presence in the sample."

tools/proteomiqon_proteininference/proteomiqon_proteininference.xml
bgruening: > So I did this to have maximum flexibility.

That is true.

> To start N jobs for N inputs, I probably need to use collections, or is there another way?

You don't need to, but it makes it easier. Your argumentation is completly correct. I guess what I'm asking is what is the difference for the user to:

1) run this tool on 2 datasets one ofter the other and concatenating the outputs
2) run this tool only once with 2 inputs

Is there a functional difference?

tools/proteomiqon_proteininference/proteomiqon_proteininference.xml
bgruening: "uses identified peptides to infere proteins" ?

tools/proteomiqon_proteininference/proteomiqon_proteininference.xml
ZimmerA: Yes, if you run the tool with multiple files and groupFiles=true, then the tool executes the inference step using all files from all runs which can lead to peptides being included in the result which would otherwise get a bad score and get rejected.

So to put it short, the N files do have a dependency on each other when groupFiles=true.
I hope this answers your question?

bgruening (approved):

bgruening:
Yes, that is a fine use-case!

PR #624, opened 2021-08-04
Add TMTpro18plex isobaric labels to MaxQuant 1.6.17.0

npinter:
This PR adds TMTpro18plex isobaric labels to the MaxQuant 1.6.17.0 wrapper which were introduced with MaxQuant bioconda package `1.6.17.0-4` (test added). In addition, help text was updated since there was an issue with Thermo RAW files and the experimental design and finally, a test for Thermo RAW files was added.

MaxQuant mqpar was updated and moved into the regular maxquant wrapper.


bgruening:
Lets see this should fail, since we need to update the container I think for the new `-4` version.

npinter:
> 
> 
> Lets see this should fail, since we need to update the container I think for the new `-4` version.

Is there a way to force use a build number, let's say 4, in the galaxy wrapper?

sth like:
`<token name="@VERSION@">1.6.17.0-4</token>`

bgruening:
Yes, something like that should work. The conda syntax should work, like you would do if you specify a build-number on you CMD

npinter:
It does not work the conda cmd line way. Can you update the container?

bgruening:
That was a coincidence ... I was just fixing the container: https://github.com/BioContainers/multi-package-containers/blob/master/combinations/mulled-v2-795058a223bc38fcee39b97525846c52cbd08579:836dcc5480c9c7e02b70a54568a699027e24aa12-0.tsv

npinter:
Fixed and moved the MaxQuant mqpar.xml tool wrapper as discussed here https://github.com/galaxyproteomics/tools-galaxyp/pull/569#issuecomment-800616286

bgruening:
The failures look legitimate ... left?

tools/maxquant/maxquant.xml
```diff
@@ -1012,6 +1012,15 @@ This tool is a wrapper for MaxQuant v@VERSION@. The current version of the wrapp
...
                                                           (File5 missing)
+        - However, “.raw” (Thermo) is considered a part of the filename:
+
+                ::
+
+                    Name         Fraction    Experiment   PTM
+                    File1.raw        1          E1       False
+                    File2.raw        2          E1       False
+                    File3.raw        3          E1       False
+                    File4.raw        4          E2       False
```
foellmelanie, line 1068: just realised that the fraction for a new experiment would start at 1 again. Name: File4.raw; Fraction: 1; Experiment E2, PTM: False. Maybe to make it less confusing add File5.raw,2,E2, False and File6.raw,3,E2,False

bgruening (approved):

PR #629, opened 2021-08-25
Thermo raw file parser outputname

jj-umn:
@subinamehta requested the the input spectrum name be preserved when using ThermoRawConverter.  
This will remove a .raw or .RAW extension from the input name and append either a .mgf or .mzml as the output name.


tools/ThermoRawFileParser/thermo_converter.xml
```diff
@@ -85,7 +85,7 @@ ThermoRawFileParser.sh
         </param>
     </inputs>
     <outputs>
-        <data name="output" format="mzml" label="${tool.name} on ${on_string}">
+        <data name="output" format="mzml" label="${($input_file.name.rsplit('.',1)[0] if $input_file.name.lower().endswith('.raw') else $input_file.name)}.${('mgf' if $format_cond.output_format == '0' else 'mzml')}">
```
bgruening, line 88: I don't think we should do such magic. Its a Galaxy issue, that we should not fix per tool. The way we recommend today is using tags. Do tags help here? @subinamehta ?

PR #632, opened 2021-08-29
Add proteomiqon_psmbasedquantification

ZimmerA:
The PSMBasedQuantification tool was designed to allow label-free quantification as well as quantification of full metabolic labeled samples.


tools/proteomiqon_psmbasedquantification/proteomiqon_psmbasedquantification.xml
```diff
@@ -0,0 +1,449 @@
...
+    <macros>
+        <token name="@VERSION@">0.0.7</token>
+    </macros>
+    <requirements>
+        <requirement type="package" version="@VERSION@">proteomiqon-psmbasedquantification</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        #import re
+        #set instrumentOutput_basename = $re.sub(r'[^\w ,.\-+]','_',$instrumentOutput.element_identifier)
+        #set scoredPSMs_basename = $re.sub(r'[^\w ,.\-+]','_',$scoredPSMs.element_identifier)
```
bgruening, line 14: Since you are in control of the program, do you have any smart idea how we can avoid encoding metadata in file names that we need to preserve?

tools/proteomiqon_psmbasedquantification/proteomiqon_psmbasedquantification.xml
```diff
@@ -0,0 +1,449 @@
...
+                }
+                #else
+                "BaseLineCorrection": null
+                #end if
+            }
+            ]]>
+        </configfile>
+    </configfiles>
+    <inputs>
+        <param name="instrumentOutput" type="data" format="sqlite" label="Instrument output" help="Specify mass spectrometry data you want to analyze."/>
```
bgruening, line 138: Just a hint, the sqlite datatype does support metadata, so you can say here "I only take sqlite dbs with a column count of X". Like here: https://github.com/galaxyproject/tools-iuc/blob/e22821f51ae326b3696a9456cbc44f9982512a52/tools/gemini/gemini_macros.xml#L46

tools/proteomiqon_psmbasedquantification/proteomiqon_psmbasedquantification.xml
```diff
@@ -0,0 +1,449 @@
...
+        <param name="PerformLabeledQuantification" type="boolean" checked="true" label="Perform labeled quantification" help="If checked, a quantification of peptides which contain an isotopic modification is performed. Additionally, for every identified peptide we check if the mass spectrometry run contains a peak at the m/z of the labeled/unlabeled counterpart and quantify it."/>
+        <section name="XicExtraction" title="XIC extraction">
+            <param name="ScanTimeWindow" type="float" value="2.0" label="Scan timewindow" help="Specify the length of the scan time window used for XIC creation."/>
+            <conditional name="MzWindow_DaTypeCond">
+                <param name="MzWindow_DaType" type="select" label="MzWindow_Da type" help="Specify the MzWindow Da Type">
+                    <option value="Fixed" selected="true">Fixed</option>
+                    <option value="Estimate">Estimate</option>
+                </param>
+                <when value="Fixed">
+                    <param name="MzWindow_Da" type="float" value="0.07" label="Value" help="Specify the m/z window set centered on the target m/z used for XIC creation" />
```
bgruening, line 150: label value is a little bit strange :) Please explain which value

tools/proteomiqon_psmbasedquantification/proteomiqon_psmbasedquantification.xml
```diff
@@ -0,0 +1,449 @@
...
+                            <when value="no">
+                            </when>
+                        </conditional>
+                        <conditional name="MinPeakLengthCond">
+                            <param name="MinPeakLengthEnabled" type="select" label="Use min peak length">
+                                <option value="yes" selected="true">Yes</option>
+                                <option value="no">No</option>
+                            </param> 
+                            <when value="yes">
+                                <param name="MinPeakLength" type="float" value="0.1" label="Min peak length" />
```
bgruening, line 225: All those conditionals with just one float you could remove and simply make the param optional if you like.

tools/proteomiqon_psmbasedquantification/proteomiqon_psmbasedquantification.xml
ZimmerA: Tell me if I am missing something, but would I not have to do this  specific task with a validator as opposed to options+filters? https://docs.galaxyproject.org/en/latest/dev/schema.html#validators-for-data-and-data-collection-parameters

tools/proteomiqon_psmbasedquantification/proteomiqon_psmbasedquantification.xml
ZimmerA: Good point! Will change :)

tools/proteomiqon_psmbasedquantification/proteomiqon_psmbasedquantification.xml
ZimmerA: Absolutely! Thanks :)

tools/proteomiqon_psmbasedquantification/proteomiqon_psmbasedquantification.xml
ZimmerA: The tool doesn't actually need this. It is just a convenience for the user to see the original file names and improve readability in the output

tools/proteomiqon_psmbasedquantification/proteomiqon_psmbasedquantification.xml
```diff
@@ -0,0 +1,385 @@
...
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        #import re
+        #set instrumentOutput_basename = $re.sub(r'[^\w ,.\-+]','_',$instrumentOutput.element_identifier)
+        #set scoredPSMs_basename = $re.sub(r'[^\w ,.\-+]','_',$scoredPSMs.element_identifier)
+        #set peptideDB_basename = $re.sub(r'[^\w ,.\-+]','_',$peptideDB.element_identifier)
+        #if $outputParamfile:
+            cat '$paramfile' >> '$out_paramfile' &&
+        #end if
+        ln -s $instrumentOutput '${instrumentOutput_basename}.mzlite' &&
```
bgruening, line 19: single quotes are missing

bgruening (approved):

bgruening:
Merci!

PR #636, opened 2021-09-27
Added hitmap wrapper

GraceAHall:
HiT-MaP is a tool for analysis of high-resolution MALDI imaging samples. 

To pass the single test in the wrapper, a node with 64Gb mem is required. 
Attempts have been made to reduce this using an even smaller dataset, but it seems to have little impact on memory usage.

### DIFF
--- a/tools/hitmap/hitmap.xml
+++ b/tools/hitmap/hitmap.xml
@@ -1,0 +1,297 @@
+
+<tool id="hitmap" name="HiT-MaP" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" license="MIT">
+    <description>High-resolution Maldi-imaging Proteomics Analysis</description>
+
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+
+    <edam_topics>
+        <edam_topic>topic_0121</edam_topic>
+    </edam_topics>
+
+    <edam_operations>
+        <edam_operation>operation_3630</edam_operation>
+    </edam_operations>
+
+    <requirements>
+        <container type="docker">mashuoa/hitmap@sha256:97f2f853362b56be6c2b0fb1507e0288667e7fb6ad0eda3ff1d761dd72dfaad0</container>   
+    </requirements>
+
+    <command detect_errors="exit_code"><![CDATA[
+        mkdir expdata &&
+        cp '${file_inputs.imzml_file}' expdata/sample.imzML &&
+        cp '${file_inputs.ibd_file}' expdata/sample.ibd &&
+        cp '${file_inputs.fasta_db}' expdata/database.fasta &&
+        Rscript '${hitmap_script}' &&
+        python3 $__tool_directory__/hitmap_gen_html.py expdata &&
+        zip -r hitmap_output.zip expdata
+    ]]></command>
+
+    <configfiles>   
+        <expand macro="hitmap_config_script" />
+    </configfiles>
+
+    <inputs>
+        <section name="file_inputs" title="File Inputs" expanded="True">
+            <param  name="imzml_file" 
+                    type="data" 
+                    format="mzml" 
+                    label="MALDI-imaging imzML file" 
+                    help="Input *.imzML file. Must have same basename as counterpart *.ibd. " />
+    
+            <param  name="ibd_file" 
+                    type="data" 
+                    format="binary" 
+                    label="MALDI-imaging binary ibd file" 
+                    help="Input *.ibd file. Must have same basename as *.imzML. " />
+    
+            <param  name="fasta_db"
+                    type="data"
+                    format="fasta"
+                    label="Proteome"
+                    help="Proteome of organism in fasta format. The proteome for a particular organism can be downloaded from https://www.uniprot.org/proteomes/. file must be fasta format. "/>
+        </section>
+
+        
+        <section name="peptide_identification" title="Peptide Identification" expanded="True">
+            <param  name="digestion_site" 
+                    type="select" 
+                    label="Digestion site" 
+                    help="Digestion enzyme used in sample preparation">
+                        <option value="arg-c proteinase">arg-c proteinase</option>
+                        <option value="asp-n endopeptidase">asp-n endopeptidase</option>
+                        <option value="bnps-skatole">bnps-skatole</option>
+                        <option value="caspase1">caspase1</option>
+                        <option value="caspase2">caspase2</option>
+                        <option value="caspase3">caspase3</option>
+                        <option value="caspase4">caspase4</option>
+                        <option value="caspase5">caspase5</option>
+                        <option value="caspase6">caspase6</option>
+                        <option value="caspase7">caspase7</option>
+                        <option value="caspase8">caspase8</option>
+                        <option value="caspase9">caspase9</option>
+                        <option value="caspase10">caspase10</option>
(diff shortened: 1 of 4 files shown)


bgruening:
@GraceAHall thanks a lot!

Do you know already this project: https://github.com/galaxyproject/galaxy-language-server

This might help you a bit with formatting. Given an input section, it can bootstrap you a test section and a command section. It will show you nice hits and has auto-completion of parameters etc. Moreover, it has common snippets for conditionals, integers ...

tools/hitmap/hitmap.xml
```diff
@@ -0,0 +1,297 @@
...
+        <edam_operation>operation_3630</edam_operation>
+    </edam_operations>
+
+    <requirements>
+        <container type="docker">mashuoa/hitmap@sha256:97f2f853362b56be6c2b0fb1507e0288667e7fb6ad0eda3ff1d761dd72dfaad0</container>   
+    </requirements>
+
+    <command detect_errors="exit_code"><![CDATA[
+        mkdir expdata &&
+        cp '${file_inputs.imzml_file}' expdata/sample.imzML &&
```
bgruening, line 23: is `ln` enough?

tools/hitmap/hitmap.xml
```diff
@@ -0,0 +1,297 @@
+
+<tool id="hitmap" name="HiT-MaP" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" license="MIT">
```
bgruening, line 2: the license refers, I think, to code that you ship with this tool not the license of the tool

Can you please also include a profile="20.05"

tools/hitmap/hitmap.xml
```diff
@@ -0,0 +1,297 @@
...
+    <description>High-resolution Maldi-imaging Proteomics Analysis</description>
+
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+
+    <edam_topics>
```
bgruening, line 9: if you like please include a bio.tools identifier. The edam ontology is not strictly needed anymore, we will get those with the bio.tools IDs. See the xref tag.

tools/hitmap/hitmap.xml
```diff
@@ -0,0 +1,297 @@
...
+        <container type="docker">mashuoa/hitmap@sha256:97f2f853362b56be6c2b0fb1507e0288667e7fb6ad0eda3ff1d761dd72dfaad0</container>   
+    </requirements>
+
+    <command detect_errors="exit_code"><![CDATA[
+        mkdir expdata &&
+        cp '${file_inputs.imzml_file}' expdata/sample.imzML &&
+        cp '${file_inputs.ibd_file}' expdata/sample.ibd &&
+        cp '${file_inputs.fasta_db}' expdata/database.fasta &&
+        Rscript '${hitmap_script}' &&
+        python3 $__tool_directory__/hitmap_gen_html.py expdata &&
```
bgruening, line 27: single quote for the path

tools/hitmap/hitmap.xml
```diff
@@ -0,0 +1,297 @@
...
+        zip -r hitmap_output.zip expdata
+    ]]></command>
+
+    <configfiles>   
+        <expand macro="hitmap_config_script" />
+    </configfiles>
+
+    <inputs>
+        <section name="file_inputs" title="File Inputs" expanded="True">
+            <param  name="imzml_file"
```
bgruening, line 37: The Galaxy datatype imzml is a composite datatype, which means it contains 2 files. You can see how this can be used with the Cardinal tools.

https://github.com/galaxyproteomics/tools-galaxyp/blob/master/tools/cardinal/macros.xml#L18

bgruening (commented):
The test data is too big, unfortunately. We try to be below 1MB. Can you maybe use the ibd/imzml files from the Cardinal tools?

tools/hitmap/hitmap.xml
GraceAHall: Unfortunatenly not. Hitmap needs the file extension to be intact, but all galaxy data has extension '.dat'

tools/hitmap/hitmap.xml
GraceAHall: do you mean $__tool_directory__/hitmap_gen_html.py -> '$__tool_directory__/hitmap_gen_html.py'?

tools/hitmap/hitmap.xml
GraceAHall: I see.  I originally tried to do it that way, but couldn't work out how to upload two files (.imzml + .ibd) together as a composite type in the galaxy uploader. 

I just went with the approach that the users upload the imzml and ibd as seperate files into their history, then select those when performing the run. 

What do you think about that approach? 

Also thank you again so much for looking over these submissions and helping me out. I really appreciate it!

tools/hitmap/hitmap.xml
GraceAHall: HiT-MaP doesn't have a listing in bio.tools. Should I leave this as edam topics?

tools/hitmap/hitmap.xml
bgruening: yes :) the reason is that you might have "spaces" in your tool_directory path .. in this case the tool breaks.

tools/hitmap/hitmap.xml
bgruening: > What do you think about that approach?

I think its can result in more errors from the user side, picking the wrong combination of files etc.
Do you know why this did not work for you? Can you try this on EU. I think its working on EU.

grafik

tools/hitmap/hitmap.xml
bgruening: In this case, leave it! We could also add it to https://github.com/bio-tools/content, but this repo is not yet in production.

tools/hitmap/hitmap.xml
bgruening: But instead of using `cp` you can do `ln -l '${file_inputs.imzml_file}' expdata/sample.imzML`

That saves a copy operation. And works in 95% of all cases. In some cases, the tool does not follow symlinks ... I hope this is not the case here.

tools/hitmap/hitmap.xml
GraceAHall: haha oh yeah that's a great point. Can happen! 
Added to the next commit :)

tools/hitmap/hitmap.xml
GraceAHall: Ohhhh I see! 
Yes this is working fine, I just wasn't aware of that option. Very neat! 
Added to the next commit

tools/hitmap/hitmap.xml
GraceAHall: Mmm gave this a shot, but since it's a container, it was having difficulty creating the link. 

I think this is an improvement which could appear in a later version of the wrapper. 

HiTMaP has been very temperamental and hard to get working, so I'd rather just keep as is rather that spending time on this optimization.

bgruening:
Hi @GraceAHall! Do we want to work on this one in 2 week? There is a new version on dockerhub: `968278fe1272192cff0f598df5981fd7301d8a4d394370765033409eb3a5053e`

bgruening:
@GraceAHall I guess we forgot this. Do we still want to do it?

PR #641, opened 2021-11-02
'mzxml' and 'mzml' filetypes must be specified in lower case

eschen42:
'mzxml' and 'mzml' filetypes must be specified in lower case for inputs to function properly in the workflow editor; I got orange input nodes rather than green when I tried to link an mzxml output to the inputs of `maxquant` or `maxquant_mqpar`.

To circumvent this issue, I had to hack the tool xml files in a local Galaxy; then the editor worked, and (luckily) I could upload the workflow to production Galaxies (e.g., UseGalaxy.eu) and they can be run successfully (they can even be edited so long as the connection made in an "offline Galaxy" is not severed).  This PR makes the case of the inputs correct, in my view.

Thank you.

attn: @npinter

### DIFF
--- a/tools/maxquant/maxquant.xml
+++ b/tools/maxquant/maxquant.xml
@@ -431,7 +431,7 @@ short peptides are usually not unique in the protein database and therefore not 
 
 
         <repeat name="paramGroups" title="Parameter Group" min="1" default="1">
-            <param type="data" format="thermo.raw,mzXML,mzML" name="files" label="Infiles" multiple="true"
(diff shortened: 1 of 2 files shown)


npinter:
Thanks. I just changed the labels to restart the test. Looks like the test command failed which has nothing to do with your PR.

bgruening:
Working on the tests, some fallouts from the Galaxy release yesterday.

bgruening:
So the test is green, but since the linting is more strict since the new version we have a few lint errors. Does anyone has time to look at those?

eschen42:
@bgruening Yes, I will look at it tomorrow

PR #644, opened 2021-12-12
Add required_files for maxquant an maxquant_mqpar

cat-bro:
Both maxquant and maxquant_mqpar call on the python scripts create_mqpar.py and modify_mqpar.py.  Both scripts require another file, mqparam.py but pulsar will not copy this file because it is not referred to by name in the tool command.

### DIFF
--- a/tools/maxquant/macros.xml
+++ b/tools/maxquant/macros.xml
@@ -96,6 +96,11 @@
             <requirement type="package" version="1.32">tar</requirement>
         </requirements>
     </xml>
+    <xml name="required_files">
+        <required_files>
+            <include path="mqparam.py" />
+        </required_files>
+    </xml>
     <xml name="ptxqc">
         <configfile name="qr_yaml">
             PTXQC:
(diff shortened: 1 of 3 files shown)


tools/maxquant/macros.xml
```diff
@@ -96,6 +96,11 @@
             <requirement type="package" version="1.32">tar</requirement>
         </requirements>
     </xml>
+    <xml name="required_files">
+        <required_files>
+            <include path="mqparam.py" />
```
bgruening, line 101: @cat-bro I think we should add all file necessary. Maybe all python files?

cat-bro: I've added create_mqpar.py (required by maxquant) and modify_mqpar.py (required by maxquant_mqpar).  I'm not sure if it is possible to say <include path="*.py"> here instead

cat-bro:
ref https://github.com/galaxyproteomics/tools-galaxyp/issues/671

tools/maxquant/macros.xml
bgruening: I think this should be fine.

bgruening (approved):

PR #649, opened 2022-02-07
make regex example useful

bernt-matthias:


tools/regex_find_replace/regex.py
```diff
@@ -27,12 +27,16 @@ def main():
             if line == "":
                 break
             for (pattern, replacement) in zip(options.patterns, options.replacements):
+                print(f"{pattern} {replacement}")
```
bgruening, line 30: are those prints needed?

bernt-matthias: No. My mistake.

bgruening (approved):

tools/regex_find_replace/regex.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="regex1" name="Regex Find And Replace" version="1.0.1">
+<tool id="regex1" name="Regex Find And Replace" version="1.0.1" profile="21.01">
```
bernt-matthias, line 1: Bump

Issue #566, opened 2021-02-22
Idea collection for the next OpenMS regenerate/merge


bernt-matthias:
- [x] for some tools users should be able to choose between batch mode and joint processing -> Just implemented for all `multiple="true"` data parameters https://github.com/WorkflowConversion/CTDConverter/pull/63/commits/0756b08eb8c4a3098080031a210ddbac4559857b
  - [x]  PSMFeatureExctractor (currently solved via patch)
  - [x] PercolatorAdaptor
- [x] implement advanced options as section (folded and maybe unfold all others)
  - for instance in the OpenSwathWorkflow the `out_chrom` output can be enabled (`out_chrom_FLAG`) but the `out_chrom_type` is not known in cheetah if the advanced options are disabled.
- [x] fix selects https://github.com/galaxyproteomics/tools-galaxyp/issues/576
  - fixed in https://github.com/WorkflowConversion/CTDConverter/pull/63/commits/8332c7d3418e904a0fef2f337a9059bab02e1050
  - see also  https://github.com/galaxyproject/galaxy/pull/11727

PR #653, opened 2022-03-01
openms 2.8

bernt-matthias:
Significant changes:

- All data parameters with `multiple="true"` are now in a conditional that allows to switch between `true` and `false` (default for `true`)
  - So far this was solved with a patch for `PSMFeatureExctractor`
  - But was requested also for `PercolatorAdaptor` https://github.com/galaxyproteomics/tools-galaxyp/issues/566#issue-813752018
- select parameters only use compatible display attributes (radio / checkboxes)
- Tests are now written directly to the tool's tests node. This should speed up tool loading quite a bit because the large test macro files don't need to be (xml-)parsed for each tool. 
- Advanced options are now implemented as (folded) section instead of a conditional (see https://github.com/galaxyproteomics/tools-galaxyp/issues/566#issue-813752018). Advantage: for instance in the OpenSwathWorkflow the `out_chrom` output can be enabled (`out_chrom_FLAG`) but the `out_chrom_type` is not known in cheetah if the advanced options are disabled.
- containerized tests

TODOs/Ideas

- [x] Move additional files (shell scripts etc) to subdir and exclude from shed upload 
- [x] add mzqc datatype to Galaxy as subclass of Json https://github.com/galaxyproject/galaxy/pull/13564
  - [x] implies profile 22.01
- [x] some booleans are not rendered as booleans but string parameters https://github.com/OpenMS/OpenMS/issues/5880
  - [x] maybe hardcode as bool for those fixed in the corresponding PR
- [x] use `re_match` for comparisons .. for this all (input) file names in output data would need to be replaced by `.*` (or something more specific ..  `[a-zA-Z0-9._-]+`?) and the other file content needs to be regex quoted  [re.escape]
(https://docs.python.org/3/library/re.html#re.escape)
  - seems too difficult at the moment (since the same file may be used as in and output) .. 
- [x] Fix failing tools
  - [x] Decharger test failing because of https://github.com/OpenMS/OpenMS/pull/5920 ?
  - [x] DecoyDatabase (symlink generation seems faulty)
  - [x] 2 FalseDiscoveryRate tests fail if `-force` is blacklisted https://github.com/OpenMS/OpenMS/issues/5921
    - readd `-force` 
  - [x] FeatureFinderIdentification test failing https://github.com/OpenMS/OpenMS/pull/5917 .. maybe blacklist parameter?
  - [x] ProteinQuantifier .. just needs larger `delta_frac` test file is small and the file name is included


tools/openms/AccurateMassSearch.xml
```diff
@@ -142,6 +156,6 @@ python3 '$__tool_directory__/fill_ctd.py' '@EXECUTABLE@.ctd' '$args_json' '$hard
   <help><![CDATA[Match MS signals to molecules from a database by mass.
 
 
-For more information, visit http://www.openms.de/doxygen/release/2.6.0/html/UTILS_AccurateMassSearch.html]]></help>
+For more information, visit http://www.openms.de/doxygen/nightly/html/UTILS_AccurateMassSearch.html]]></help>
```
bgruening, line 159: that looks wrong ..

tools/openms/BaselineFilter.xml
```diff
@@ -41,14 +41,14 @@ python3 '$__tool_directory__/fill_ctd.py' '@EXECUTABLE@.ctd' '$args_json' '$hard
...
-    <param name="struc_elem_unit" argument="-struc_elem_unit" display="checkboxes" type="select" optional="false" label="Unit of 'struc_elem_length' paramete" help="">
+    <param argument="-in" type="data" format="mzml" optional="false" label="input raw data file" help=" select mzml data sets(s)"/>
+    <param argument="-struc_elem_length" type="float" optional="false" value="3.0" label="Length of the structuring element (should be wider than maximal peak width - see documentation)" help=""/>
+    <param argument="-struc_elem_unit" type="text" optional="false" label="Unit of 'struc_elem_length' paramete" help="">
       <option value="Thomson" selected="true">Thomson</option>
       <option value="DataPoints">DataPoints</option>
       <expand macro="list_string_san" name="struc_elem_unit"/>
     </param>
-    <param name="method" argument="-method" type="select" optional="false" label="The name of the morphological filter to be applied" help="If you are unsure, use the default">
+    <param argument="-method" type="text" optional="false" label="The name of the morphological filter to be applied" help="If you are unsure, use the default">
```
bgruening, line 51: this looks also wrong

tools/openms/ClusterMassTraces.xml
```diff
@@ -41,16 +41,16 @@ python3 '$__tool_directory__/fill_ctd.py' '@EXECUTABLE@.ctd' '$args_json' '$hard
...
+    <param argument="-max_lag" type="integer" optional="false" value="1" label="Maximal lag" help=""/>
+    <param argument="-max_rt_apex_difference" type="float" optional="false" value="5.0" label="Maximal difference of the apex in retention time" help=""/>
+    <param argument="-max_intensity_cutoff" type="float" optional="false" value="0.0" label="Maximal intensity to be added to a spectrum" help=""/>
+    <param argument="-add_precursor" type="float" optional="false" value="0.0" label="Add a precursor mass" help=""/>
     <expand macro="adv_opts_macro">
-      <param name="force" argument="-force" type="boolean" truevalue="true" falsevalue="false" checked="false" label="Overrides tool-specific checks" help=""/>
-      <param name="test" argument="-test" type="hidden" optional="false" value="False" label="Enables the test mode (needed for internal use only)" help="">
+      <param argument="-force" type="boolean" truevalue="true" falsevalue="false" checked="false" label="Overrides tool-specific checks" help=""/>
+      <param argument="-test" type="hidden" optional="false" value="False" label="Enables the test mode (needed for internal use only)" help="">
```
bgruening, line 53: test and force should probably not be exposed to the user?

tools/openms/CompNovo.xml
```diff
@@ -41,28 +41,28 @@ python3 '$__tool_directory__/fill_ctd.py' '@EXECUTABLE@.ctd' '$args_json' '$hard
...
   </configfiles>
   <inputs>
-    <param name="in" argument="-in" type="data" format="mzml" optional="false" label="input file in mzML format" help=" select mzml data sets(s)"/>
+    <param argument="-in" type="data" format="mzml" optional="false" label="input file in mzML format" help=" select mzml data sets(s)"/>
     <section name="algorithm" title="Algorithm section" help="" expanded="false">
-      <param name="max_number_aa_per_decomp" argument="-algorithm:max_number_aa_per_decomp" type="integer" optional="false" value="4" label="maximal amino acid frequency per decomposition" help=""/>
-      <param name="tryptic_only" argument="-algorithm:tryptic_only" type="text" optional="false" value="true" label="if set to true only tryptic peptides are reported" help="">
+      <param argument="-algorithm:max_number_aa_per_decomp" type="integer" optional="false" value="4" label="maximal amino acid frequency per decomposition" help=""/>
+      <param argument="-algorithm:tryptic_only" type="text" optional="false" value="true" label="if set to true only tryptic peptides are reported" help="">
```
bgruening, line 47: looks like a boolean and not like a text

tools/openms/CruxAdapter.xml
```diff
@@ -45,94 +45,94 @@ python3 '$__tool_directory__/fill_ctd.py' '@EXECUTABLE@.ctd' '$args_json' '$hard
...
     </param>
-    <param name="extra_percolator_args" argument="-extra_percolator_args" type="text" optional="true" value="" label="Extra arguments to be passed to percolato" help="">
+    <param argument="-extra_percolator_args" type="text" optional="true" value="" label="Extra arguments to be passed to percolato" help="">
       <expand macro="list_string_san" name="extra_percolator_args"/>
     </param>
-    <param name="precursor_mass_tolerance" argument="-precursor_mass_tolerance" type="float" optional="false" value="10.0" label="Precursor monoisotopic mass tolerance (Crux parameter: peptide_mass_tolerance)" help=""/>
-    <param name="precursor_mass_units" argument="-precursor_mass_units" display="checkboxes" type="select" optional="false" label="Unit of precursor mass tolerance (amu, m/z or ppm)" help="">
+    <param argument="-precursor_mass_tolerance" type="float" optional="false" value="10.0" label="Precursor monoisotopic mass tolerance (Crux parameter: peptide_mass_tolerance)" help=""/>
+    <param argument="-precursor_mass_units" type="text" optional="false" label="Unit of precursor mass tolerance (amu, m/z or ppm)" help="">
```
bgruening, line 60: select not text

tools/openms/CruxAdapter.xml
```diff
@@ -45,94 +45,94 @@ python3 '$__tool_directory__/fill_ctd.py' '@EXECUTABLE@.ctd' '$args_json' '$hard
...
+    <param argument="-isotope_error" type="text" optional="true" value="" label="List of positive, non-zero integers" help="">
       <expand macro="list_string_san" name="isotope_error"/>
     </param>
-    <param name="run_percolator" argument="-run_percolator" type="boolean" truevalue="true" falsevalue="false" checked="true" label="Whether to run percolator after tide-search" help=""/>
-    <param name="enzyme" argument="-enzyme" type="select" optional="false" label="The enzyme used for peptide digestion" help="">
+    <param argument="-run_percolator" type="boolean" truevalue="true" falsevalue="false" checked="true" label="Whether to run percolator after tide-search" help=""/>
+    <param argument="-enzyme" type="text" optional="false" label="The enzyme used for peptide digestion" help="">
```
bgruening, line 72: select

tools/openms/CompNovo.xml
bernt-matthias: https://github.com/OpenMS/OpenMS/issues/5880

tools/openms/CruxAdapter.xml
bernt-matthias: Consider the selects fixed.

tools/openms/ClusterMassTraces.xml
bernt-matthias: You are right with `force`. `test` is hidden and needed for the tests passing.

bernt-matthias: Actually it seems that `-force` is needed https://github.com/OpenMS/OpenMS/issues/5921

bgruening:
restarted the tests

bernt-matthias:
No idea why the test fails. Can't reproduce this locally.

bernt-matthias:
All, but linting succeeded. The linter errors should be fixed when the 22.01 galaxy python packages are available. I can wait for this I guess.

Not sure if I should increase the profile, for the `mzqc` datatype that became available in 22.01...

bernt-matthias:
@bgruening I think its good as long as you agree with keeping the profile (correct would be 22.01 because of the mzqc datatype .. but maybe that is not important enough .. and 22.01 is maybe too recent).

Since linting will fail we will need to push to the TS manually, or?

bgruening (approved):
What a marvelous work!

PR #660, opened 2022-04-06
metanovo: do not let user decide on cores

bernt-matthias:
Just wanted to try this tool and have observed some things that should be changed:

1. Users should/must not be allowed to change the number of cores and the memory. That is the job of Galaxy and the Galaxy admin.

I disabled the possibility to set the core and set the `Xmx` and `Xms` to an empty strings. The number of available cores is available via the `$GALAXY_SLOTS` environment variable (I subtract 1 as mentioned in the docs).

The memory options need to be set by Galaxy / Galaxy admin via the  `_JAVA_OPTIONS` environment variable (https://stackoverflow.com/questions/28327620/difference-between-java-options-java-tool-options-and-java-opts) which will overwrite these variables.

2. Tests are needed @reid-wagner  .. can you help here?

Finally I think that the bioconda recipe is should be improved. Instead of downloading (precompiled?) versions of SearchGUI and other software this should be specified as requirements in the recipe. I think most/all of the dependencies are available in bioconda.

PR #661, opened 2022-04-19
metanovo: Add tests

reid-wagner:
Two tests are added, one for the single MGF input case, and one for the collection input case. A separate input FASTA file  and an expected output FASTA are included for each case.


bernt-matthias:
Thanks, looks good to me. I included the commit into https://github.com/galaxyproteomics/tools-galaxyp/pull/660

PR #662, opened 2022-04-28
Fix multi file mode in proteomiqon_proteininference

ZimmerA:
This PR fixes a bug that caused no collection being returned in multi file mode.


ZimmerA:
Do the tools update on the instance if I do not increase the version number?

PR #668, opened 2022-05-26
Update pep_pointer to use python3

cat-bro:
I've converted pep_pointer to python3 because the previous version's dependency (python 2.7.9) is no longer available with conda 4.9.* and above.  <10 lines needed to be changed for the conversion from python2 to python3, the rest of the changes are for flake8.


tools/pep_pointer/pep_pointer.xml
```diff
@@ -1,7 +1,7 @@
-<tool id="pep_pointer" name="PepPointer" version="0.1.3">
+<tool id="pep_pointer" name="PepPointer" version="0.1.3+galaxy1">
```
bernt-matthias, line 1: I guess here we can just increase the version to 0.1.4, because it does not depend on external software ..

PR #675, opened 2022-07-21
Xml lower

eschen42:
mzml and mzxml must be lower case for wrapped tool to accept them correctly.


eschen42:
@bgruening Everything works; should we increment the version or just the build suffix?

bgruening:
Thanks @eschen42

PR #676, opened 2022-07-27
eggnog mapper: implement search and annotation phase as additional separate tools

bernt-matthias:
for rationale see README

TODO:

- [x] separate output options that apply only to annotation / search phase -> macros

### DIFF
--- a/tools/eggnog_mapper/eggnog_macros.xml
+++ b/tools/eggnog_mapper/eggnog_macros.xml
@@ -3,6 +3,7 @@
    <token name="@TOOL_VERSION@">2.1.8</token>
    <token name="@VERSION_SUFFIX@">3</token>
    <token name="@EGGNOG_DB_VERSION@">5.0.2</token>
+   <token name="@PROFILE@">22.01</token>
     <!--
     # DB versionning was super confusing for eggnog-mapper 2.0.x:
     # eggnog-mapper 1.* needed a db v4.5 (based on eggnog v4.5)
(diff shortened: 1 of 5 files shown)


bgruening:
Having those 3 steps in one tool is not an option?

bernt-matthias:
> Having those 3 steps in one tool is not an option?

I do not see how to map the steps to different destinations if implemented in one tool (dynamic destinations might be an option, but many sites use static destinations.. Like mine ;))

bernt-matthias:
Btw. Linting failures are due to the missing updates of the tool-util package for 21.01. Any idea how to get this on the way?

bernt-matthias:
> Having those 3 steps in one tool is not an option?

- eggnog_mapper.xml is still the monolithic tool 
- I just added two tools separating the search and annotation phase (as suggested by the docs https://github.com/eggnogdb/eggnog-mapper/wiki/eggNOG-mapper-v2.1.5-to-v2.1.12#setting-up-large-annotation-jobs), which allows setting different number of CPUs and a memory for the two stages. It also allows for a better parallelization of the search phase by splitting the input into chunks (+ the parallelization provided by the underlying search tool).

Let my just add https://github.com/galaxyproteomics/tools-galaxyp/pull/727 to this PR ..

bernt-matthias:
> Let my just add https://github.com/galaxyproteomics/tools-galaxyp/pull/727 to this PR ..

I reconsidered (because `--dbmem` was already handled here) and just left it as is. I'm happy to change to an entirely admin-controlled behavior, i.e. to use `--dbmem` only if the admin sets it as in https://github.com/galaxyproteomics/tools-galaxyp/pull/727

bgruening (approved):

PR #677, opened 2022-07-28
Update requirements in proteogenomics macros

chambm:
My conda couldn't resolve the current dependency set, probably due to biopython, but it couldn't resolve even when biopython's version was unspecified.


bernt-matthias (commented):
You need to bump all tools depending on this macro.

PR #679, opened 2022-08-18
regex find/replace: use format_source for output

jancrichter:
Version 1.0.2 of the tool uses format="input" in its output declaration. This creates an output with data type "input". I figure, the intention is instead to create an output with the same data type as the input. This should fix it, right?

Incidentally, this same error can currently be found in the Tool XML documentation.

### DIFF
--- a/tools/regex_find_replace/regex.xml
+++ b/tools/regex_find_replace/regex.xml
@@ -48,26 +48,26 @@
   </outputs>
   <tests>
     <test>
-      <param name="input" value="find1.txt"/>
+      <param name="input" value="find1.txt" ftype="txt"/>
       <param name="pattern" value="(T\w+)"/>
       <param name="replacement" value="\1 \1" />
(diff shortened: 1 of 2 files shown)


bgruening (approved):
Looks good to me! 

@bernt-matthias good to go?

bernt-matthias (commented):
`format_source` is more explicit (I remember that `format="input"` implies that the format is taken from the 1st input).

Can you add the `ftype` attribute to the in- and outputs in the tests.

bernt-matthias:
Got it in the [last PR](https://github.com/galaxyproteomics/tools-galaxyp/pull/649) we set a profile. Problem is that `format="input"` was disabled for non legacy tools (ie tools which specify any profile): https://docs.galaxyproject.org/en/master/dev/schema.html#id1. This was unnoticed because the tests did not include `ftype`

Using format_source is the correct fix here.

bernt-matthias (approved):

PR #680, opened 2022-08-28
fix the uniprot URL

bgruening:
@jj-umn @bernt-matthias we have a problem here. The uniprot API has changed. 

I could not quickly find a way to download everything in one go. It seems the UniProt API is limited to 500 entries per request.
And other things have also changed.


jj-umn:
@reid-wagner @bgruening   I probably won't be able to look at this until after I get back on Sept. 6.  I'll have minimal internet access until then.

PR #682, opened 2022-09-26
Pr 680

jj-umn:


jj-umn:
@bgruening  So I obviously didn't figure out how to do a pull request to a pull request.

bgruening:
very cool. Have you tried downloading many 1000 datasets? Uniprot is limiting the number of datasets now and we need to paginate over the result list? Have you encountered that?

jj-umn:
@bgruening I used the pagination script first, but it was considerably slower for the fasta downloads I tried.   I haven't tried retrieving multiple datasets in a query.   
Do the requirements in the tool wrapper look OK, or is there a better way to do that?

bgruening:
The requirements look ok. Let's get thhiis in and ask Melii to test.

Thanks a lot.

foellmelanie:
Hi @jj-umn thank you so much for this quick fix! It workes for me now. 

There is one strange thing: when I set "proteome set" to "reference proteome set" I obtain more proteins than when I set it to "any". Typically it was the other way around; is there maybe an accidental invert? 

Best, Melanie

jj-umn:
@foellmelanie I saw that too and was a bit confused by that.   But that seems to be the way the Uniprot API does that now.  If you can see something different than that, let me know.  Thanks.

bgruening:
@jj-umn I think we need to twek the URLs.

This one works for me:

https://rest.uniprot.org/uniprotkb/stream?compressed=true&format=fasta&query=taxonomy_id:9606+AND+reviewed:true' 

Can it be that we need to add the `AND` and also the `:` everywhere?
I have found that here as description: https://www.uniprot.org/help/api_queries

jj-umn:
@bgruening @foellmelanie   So that's what it was.    I'll try to get that fixed.

foellmelanie:
Hi @jj-umn thanks for your efforts! 
I am not sure how is the current status, but I just tested again the 0.3.4 version: https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/galaxyp/dbbuilder/dbbuilder/0.3.4 and found that for the parameter "reviewed" the two options "UniProtKB" and "reviewed only" can be selected but when I click "execute" there is a blue error message popping up and I cannot start the tool:  parameter 'reviewed': an invalid option ('reviewed:true') was selected (valid options: ,+reviewed%3Atrue,+reviewed%3Afalse). 
Best, Melanie

jj-umn:
@foellmelanie  I'll check on that.

jj-umn:
@foellmelanie @bgruening   You may need to uninstall earlier versions, and just keep version 0.3.4.   It looks like it might be defaulting to version 0.3.3   I'm not having that issue with either MSI or laptop installed dbbuilder 0.3.4

PR #685, opened 2022-10-31
Pepquery2

jj-umn:
PepQuery version 2 greatly expands the capabilities of the PepQuery application and the command line options are considerably changed.   The 20 demos in the documentation:  http://pepquery.org/document.html   we used to test the options during development of the pepquery2 galaxy tool.


tools/pepquery2/pepquery2.xml
```diff
@@ -0,0 +1,809 @@
...
+                 <param name="decoy" argument="-decoy" type="boolean" truevalue="-decoy" falsevalue="" checked="false" label="Identity the decoy version of the selected target protein"/>
+            </when>
+        </conditional>
+        <section name="req_inputs" title="Input Data" expanded="true">
+            <conditional name="input_type">
+                <param name="input_type_selector" argument="-t" type="select" label="Input Type" help="" >
+                    <option value="peptide">peptide</option>
+                    <option value="protein">protein</option>
+                    <option value="DNA">DNA (translate to protein sequences)</option>
+                    <!-- these are not working with pepquery version 1.6
```
bgruening, line 168: But this is version 2? Should we drop this?

tools/pepquery2/pepquery2.xml
```diff
@@ -0,0 +1,809 @@
...
+                    <param name="frame" argument="-f" type="select" label="Frame(s) for DNA translation" multiple="true" help="The frame(s) to translate DNA sequence to protein. Selecting nothing (default) keeps the longest frame">
+                        <option value="1">1</option>
+                        <option value="2">2</option>
+                        <option value="3">3</option>
+                        <option value="4">4</option>
+                        <option value="5">5</option>
+                        <option value="6">6</option>
+                    </param>
+                </when>
+                <!-- these are not working with pepquery version 1.6
```
bgruening, line 234: Verion 2.0 drop?

tools/pepquery2/pepquery2.xml
```diff
@@ -0,0 +1,809 @@
...
+                    <param name="min_length" value="7"/>
+                    <param name="max_length" value="45"/>
+                    <param name="num_random_peptides" value="1000"/>
+                </section>
+            </section>
+            <output name="psm_rank_txt">
+                <assert_contents>
+                    <has_text text="LVVVGADGVGK" />
+                    <not_has_text text="AHSSMVGVNLPQK" />
+                    <has_text text="02CPTAC_LUAD_W_BI_20180518_KR_f15:25149:2" />
```
bgruening, line 515: It could be useful to check for the number of columns in those outputs. This is a general comment as you are dealing with column headers etc.

tools/pepquery2/pepquery2.xml
```diff
@@ -0,0 +1,809 @@
...
+           <actions>
+                <action name="comment_lines" type="metadata" default="1" />
+                <action name="column_names" type="metadata" default="peptide,modification,n,spectrum_title,charge,exp_mass,tol_ppm,tol_da,isotope_error,pep_mass,mz,score,n_db,total_db,n_random,total_random,pvalue,rank,ptm_spectrum_title,ptm_peptide,ptm_charge,ptm_exp_mass,ptm_pep_mass,ptm_tol_ppm,ptm_tol_da,ptm_isotope_error,ptm_modification,ptm_score" />
+            </actions>
+        </data> 
+
+    </outputs>
+    <tests>
+        <!-- Test-1  PepQueryDB peptide gencode:human -->
+        <test>
```
bgruening, line 457: please add https://docs.galaxyproject.org/en/latest/dev/schema.html#id92 expect_num_outputs

bgruening (commented):
Only some general comments, very nice work @jj-umn

Do you think we should remove version 1? Maybe not remove but overwrite the old version with this new version in the TS?

tools/pepquery2/pepquery2.xml
jj-umn: @bgruening  They are valid forms of input, but I have not implemented them.  I probably left that commented code in there in case someone wanted to make this complete.   I'll try to address your comments later in the week.  Once again thanks for the careful review.

jj-umn:
@bgruening I think I addressed the suggestions in your review.   Let me know if you see anything else.  Thanks.

bgruening (approved):

bgruening:
Thanks a lot!

PR #687, opened 2022-11-03
Update uniprotxml_downloader to use new Uniprot API

reid-wagner:
- Updates script to use newer uniprot API.
- Adds field option and parameter to python script and Galaxy tool that specifies if IDs or names are provided for query.

### DIFF
--- a/tools/uniprotxml_downloader/uniprotxml_downloader.xml
+++ b/tools/uniprotxml_downloader/uniprotxml_downloader.xml
@@ -1,5 +1,8 @@
-<tool id="uniprotxml_downloader" name="UniProt" version="2.2.0" profile="21.01">
+<tool id="uniprotxml_downloader" name="UniProt" version="2.3.0" profile="21.01">
     <description>download proteome as XML or fasta</description>
+    <macros>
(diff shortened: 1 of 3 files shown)


tools/uniprotxml_downloader/uniprotxml_downloader.xml
```diff
@@ -1,5 +1,8 @@
-<tool id="uniprotxml_downloader" name="UniProt" version="2.2.0" profile="21.01">
+<tool id="uniprotxml_downloader" name="UniProt" version="2.3.0" profile="21.01">
     <description>download proteome as XML or fasta</description>
+    <macros>
+        <import>macros.xml</import>
```
bgruening, line 4: small macros you could include directly here

bgruening (approved):
Thanks @reid-wagner

tools/uniprotxml_downloader/uniprotxml_downloader.xml
reid-wagner: Thanks, I considered it but arbitrarily chose macros.xml.

PR #698, opened 2023-01-31
Add new reference for MaxQuant, MSStats and MSStatsTMT

npinter:
I've added https://pubs.acs.org/doi/10.1021/acs.jproteome.2c00051 to the references (and did some small changes regarding the version suffix in MSStats and MSStatsTMT).

### DIFF
--- a/tools/maxquant/macros.xml
+++ b/tools/maxquant/macros.xml
@@ -882,6 +882,7 @@
                 }
             </citation>
             <citation type="doi">10.1021/acs.jproteome.5b00780</citation>
+            <citation type="doi">10.1021/acs.jproteome.2c00051</citation>
         </citations>
     </xml>
 </macros>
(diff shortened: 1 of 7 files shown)


tools/maxquant/macros.xml
```diff
@@ -882,6 +882,7 @@
                 }
             </citation>
             <citation type="doi">10.1021/acs.jproteome.5b00780</citation>
```
bgruening, line 884: should the old one be dropped?

npinter: @foellmelanie what do you think? keep both?

npinter: @foellmelanie has agreed to leave both references in!

bgruening (approved):

bgruening:
Thanks.

PR #701, opened 2023-02-14
corrected sorting issue and non-useful parameter setting

eschen42:
Sort substrates by quality.
Do not filter out correlations less than one.


eschen42:
@bgruening this is a bug fix thank you.

bgruening (approved):

PR #702, opened 2023-02-17
merged redundant parameters

eschen42:
One parameter that applied to two stages of this tool was repeated for each stage, yet it should have the same value for both. This submission addresses this issue.


bgruening (approved):

eschen42:
@bgruening Awesome speed! Thank you.

PR #710, opened 2023-04-12
Add QuPath ROI splitter

npinter:
This tool processes QuPath cell detection-based annotations extracted from TMA (Tissue Microarray) projects. Users must first run the provided QuPath script locally to generate GeoJSON files containing the annotations. The generated GeoJSON files are then uploaded to Galaxy as a collection. The tool processes these GeoJSON files and splits the annotations by ROI, outputting the results as a collection of text files, each containing x and y coordinates.

The QuPath (Groovy) script can be found in this repo https://github.com/npinter/ROIsplitter. I'm not sure if it's better to have the Python script in the tool directory (as it is now) or to have some sort of a direkt link to my repo. 

Thanks :)


bgruening:
I think the script here is ok. The alternative is to create a cond package of your scripts.

tools/qupath_roi_splitter/qupath_roi_splitter.xml
```diff
@@ -3,10 +3,13 @@
     <requirements>
         <requirement type="package" version="3.0.1">geojson</requirement>
         <requirement type="package" version="1.24.2">numpy</requirement>
-        <requirement type="package" version="4.7.0">libopencv</requirement>
+        <requirement type="package" version="4.7.0">opencv</requirement>
         <requirement type="package" version="2.0.0">pandas</requirement>
     </requirements>
      <command detect_errors="exit_code"><![CDATA[
+        export LIBGL_PATH=$(find /usr -name libGL.so.1) &&
```
bgruening, line 10: Its running in a container, and this container probably has no libgl.
Can you create a container here please: https://github.com/BioContainers/multi-package-containers/

With this workarund we can create a libgl container.

npinter: thanks!

npinter: libGL isn't availiable as a conda package, so I assume this will not work?

there is a headless opencv package availiable, which should work for use cases like servers and docker containers. 
It's availiable via fastai (not cia conda-forge).

https://github.com/BioContainers/multi-package-containers/pull/2585

bgruening:
@npinter a few thoughts.

* we can merge the Biocontainers PR, it will create a "mulled" container
* we can then explicitly require this container, but then it will not work with conda
* as it is now, it also will not work on most of the Galaxy instances because this new channel is not enabled in most instances
* enabling this channel in Galaxy will potentially cause a lot of trouble, so I would not even recommend it. There are reasons why we only support bioconda and conda-forge
* we could try to move this one package to conda-forge
* we could use opencv and build an OpenGL container - similar to your current biocontainers PR

npinter:
> @npinter a few thoughts.
> 
>     * we can merge the Biocontainers PR, it will create a "mulled" container
> 
>     * we can then explicitly require this container, but then it will not work with conda
> 
>     * as it is now, it also will not work on most of the Galaxy instances because this new channel is not enabled in most instances
> 
>     * enabling this channel in Galaxy will potentially cause a lot of trouble, so I would not even recommend it. There are reasons why we only support bioconda and conda-forge
> 
>     * we could try to move this one package to conda-forge
> 
>     * we could use opencv and build an OpenGL container - similar to your current biocontainers PR

Thanks!! Since https://github.com/opencv/opencv-python#installation-and-usage is recommending using the headless variant, I tend to move the headless package to conda-forge.

I don't know how to get OpenGL working in a mulled container, I would need some help here. There is pyopengl availiable as a conda package but I don't think it will be compatible with opencv.

npinter:
Adding a headless version to `conda-forge` is already discussed here https://github.com/conda-forge/opencv-feedstock/issues/293 and her https://github.com/rapidsai/ci-imgs/pull/31.

npinter:
I don't know how to continue with the mulled container and/or if opencv with libGL/openGL is part of the container. The last test raised: `ModuleNotFoundError: No module named 'cv2'`

bgruening:
Läuft! :)

bgruening (approved):

bgruening:
You can add it to EU now :)

PR #714, opened 2023-05-05
Updated output label in ThermoRawFileParser

### LINKED ISSUE #706: thermo raw file converter changes names of files in the collection
While the thermo raw file converter leaves the display labels of samples intact, it renames the files in a collection using the tool name - this makes downstream analysis which relies on filenames (which I know is not best, but common, practice) impossible and breaks quite some downstream tools. Is it possible to keep the filenames intact and not change those? `msconvert` actually leaves those intact - this behaviour should be adopted.

zargham-ahmad:
This PR includes the following:

- Updated output label in ThermoRawFileParser #706 
- bump version

closes #706


bernt-matthias:
Hi @zargham-ahmad thanks for the contribution. I think this is not the correct solution... see my comment in the original issue https://github.com/galaxyproteomics/tools-galaxyp/issues/706#issuecomment-1540075407

PR #717, opened 2023-06-14
Fix QuPath ROI Splitter

npinter:
I accidentally included a bug in the Galaxy version that has already been fixed.


bgruening:
Hi @npinter,

slightly offtopic, but since I'm working on slides currently ... do you know that we can run Desktop applications in Galaxy ... https://galaxyproject.org/news/2023-05-21-fair-ease-euro-science-gateway/ would that be interesting for QuPath?

npinter:
Yes, definitely, so does it run in a VM? Does the VM have GPU(s)?

bgruening (approved):

bgruening:
Thanks. 
Ping me if you want more information about the remote option.

PR #719, opened 2023-06-23
Add DIA-NN 

npinter:
This PR adds the tool wrapper for DIA-NN (https://github.com/vdemichev/DiaNN), a universal software for data-independent acquisition (DIA) proteomics data processing. I tried to create a conda package of DIA-NN (https://github.com/bioconda/bioconda-recipes/pull/41491) but it failed because of dependency issues. 
However, a docker container (https://hub.docker.com/r/biocontainers/diann/tags) was already available of the newest version (v1.8.1). A docker container at quay.io (https://quay.io/repository/bigbio/diann?tab=tags) exists as well, but only in version v1.8.0.


tools/diann/diann.xml
```diff
@@ -0,0 +1,486 @@
...
+            ## if brukertdf.d.tar, extract to ./input_data and append "./input_data/" to infiles_str
+            #if $infile.is_of_type("brukertdf.d.tar"):
+                tar -xf '$infile' -C ./input_data &&
+                #set $infiles_str += '--f ./input_data/' + $infile.name[:-4] + ' '
+            #else
+                #set $infiles_str += '--f ' + $infile.name + ' '
+            #end if
+        #end for
+
+        diann-@TOOL_VERSION@
```
bgruening, line 26: I would need to check the container when I'm home, but are you sure the binary is called after the version number?

npinter: no, just found out it's not the case.........

npinter: the command in the docker is just `diann` and not `diann-1.8.1`

tools/diann/diann.xml
```diff
@@ -0,0 +1,486 @@
...
+
+        ## generate string like "--f file1 --f file2 ..." for each input in $input.f, which are comma separated
+        #set infiles_str = ''
+        #for $infile in $input.f
+            ## if brukertdf.d.tar, extract to ./input_data and append "./input_data/" to infiles_str
+            #if $infile.is_of_type("brukertdf.d.tar"):
+                tar -xf '$infile' -C ./input_data &&
+                #set $infiles_str += '--f ./input_data/' + $infile.name[:-4] + ' '
+            #else
+                #set $infiles_str += '--f ' + $infile.name + ' '
```
bgruening, line 22: Infile.name is the history name and not the path correct? Is that really like it should be? If so it's better to use element_identifier than name

tools/diann/diann.xml
```diff
@@ -0,0 +1,486 @@
...
+            #end if
+
+            #if $input.spectral_lib_options.gen_spec_lib
+            && cp ./report-lib.tsv '$output_report_lib'
+            #end if
+    ]]></command>
+    <inputs>
+        <!-- not used: cfg, convert, dir, ext, no_quant_files, out, out_lib, out_lib_copy, prefix, temp, threads, use-quant -->
+        <section name="input" expanded="true" title="Input files">
+            <param name="f" type="data" format="mzml,dia,wiff,thermo.raw,brukertdf.d.tar" multiple="true" optional="true" label="Input file" help="Specify a run to be analysed"/>
```
bgruening, line 225: How can this be optional?

tools/diann/diann.xml
npinter: bc the name can be changed? changed it!

tools/diann/diann.xml
```diff
@@ -0,0 +1,486 @@
...
+
+        ## generate string like "--f file1 --f file2 ..." for each input in $input.f, which are comma separated
+        #set infiles_str = ''
+        #for $infile in $input.f
+            ## if brukertdf.d.tar, extract to ./input_data and append "./input_data/" to infiles_str
+            #if $infile.is_of_type("brukertdf.d.tar"):
+                tar -xf '$infile' -C ./input_data &&
+                #set $infiles_str += '--f ./input_data/' + $infile.element_identifier [:-4] + ' '
+            #else
+                #set $infiles_str += '--f ' + $infile.element_identifier  + ' '
```
bgruening, line 22: So this will end up as `--f small-peakpicking-cwt-allMS`. But no path is ever passed to the program, or did I miss that? How should the tool know where the file is located?

tools/diann/diann.xml
```diff
@@ -0,0 +1,486 @@
...
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        mkdir ./input_data &&
+        mkdir ./tmp &&
+
+        ## generate string like "--f file1 --f file2 ..." for each input in $input.f, which are comma separated
+        #set infiles_str = ''
+        #for $infile in $input.f
+            ## if brukertdf.d.tar, extract to ./input_data and append "./input_data/" to infiles_str
+            #if $infile.is_of_type("brukertdf.d.tar"):
```
bgruening, line 18: Should we add a bruker test, to be sure this actually works?

bgruening (commented):
I'm just confused with the inputs, otherwise, this looks pretty great. As usual great work @npinter !

tools/diann/diann.xml
npinter: You are right! fixed

tools/diann/diann.xml
npinter: Bruker files are quite large, even if one measures a blank in a very short gradient.. I'll try to generate one manually

bgruening: Yeah, I know. 2MB would be maybe also ok. But if we do not test this code path, it might not work :(

npinter: I found one test file in the msconvert wrapper, but it's not functional with DIA-NN (raises a C++ error). I even tried to fix it... I do not know how to get a smaller file. Even if we measure a blank with a shorter gradient we get files ~500MB.

I will try to run the small file here as well, but I guess it will fail as well. I can confirm it's running with one of those files measured on our machine (timsTOF flex)

npinter: grafik
the variable is not being expanded correctly when wrapped in quotes

bgruening (approved):
Lets ship it, you spent so much time on this tool already ...

PR #720, opened 2023-07-06
Add accession number to Uniprot search options

reid-wagner:
To accommodate searching protein accession numbers, the tool's naming convention was generalized away from referring to taxa specifically.

The tool may be expanded in the future to include more of the possible Uniprot search fields.


bgruening (approved):
Cool, thanks

reid-wagner:
Thanks!

PR #722, opened 2023-07-14
fix PepQuery2 memory

mira-miracoli:
I changed the Java options, because the JVM overhead can lead to failing jobs, when the tool uses the full amount defined in GALAXY_MEMORY_MB as stack.
- https://github.com/galaxyproject/tpv-shared-database/issues/30


tools/pepquery2/pepquery2.xml
```diff
@@ -32,7 +32,7 @@
 #end if
 ## PepQuery command
 pepquery 
-  -Xmx\$[ \${GALAXY_MEMORY_MB:-8192} / 1024 ]g
+  -Xmx\$(( \${GALAXY_MEMORY_MB:-8192} - 256 - (\${GALAXY_SLOTS:-1} / 5) - \${GALAXY_MEMORY_MB:-8192} * (1 / 20) ))m
```
bernt-matthias, line 35: How do you arrive at this formula? 

If you want to make the memory depend on the number of available threads use it may be better to use `\$GALAXY_MEMORY_MB_PER_SLOT`?

Wouldn't it be a simple solution to let the admin just give more memory to the application?

bernt-matthias: Here is what I do on my instance. In each destination of the job_conf I add

```
<env id="_JAVA_OPTIONS">-Xmx@JAVAMEM@ -Xms256m</env>
```

and set `@JAVAMEM@` to something lower than the available memory (roughly 33%). I guess this does not work here because the bioconda recipe seems not to respect `_JAVA_OPTIONS`, but I might be wrong. If I'm right it would be good to adapt the recipe ..  a grep `_JAVA_OPTIONS` in the bioconda recipes should guide you. 

Would be interesting to ask the TPV / admin community how they handle java memory.

mira-miracoli: Wouln't it be cleaner to substract that 33% already in the wrapper and not later. Otherwise all admins would need to think of it (like me for example) and we would need an additional destination and configuration lines

mira-miracoli: Test 2 is failing for me on the main branch with `--docker`, could someone check if it is just for me or in general?

bgruening:
@mira-miracoli we have now something similar in the default tool. Do we need this PR still?

mira-miracoli:
I guess not

PR #727, opened 2023-09-04
eggnog: allow admins to set --dbmem

bernt-matthias:
having the annotation DB on slow memory causes eggnog to be very inefficient in the annotation phase. For larger sequence databases to be processed we need to use more than 1 CPU .. and then the annotation phase is even more inefficient.

I hope that --dbmem can help here.


bernt-matthias:
Dropped in favour of https://github.com/galaxyproteomics/tools-galaxyp/pull/676

PR #730, opened 2023-10-12
Increase default JVM stack size for PepQuery2

reid-wagner:
The "non enzyme" digestion option results in a much larger search space and deeper recursion within PepQuery2, which can result in a stack overflow. This PR sets the JVM per-thread stack size to 2048KB. The default on Linux x86_64 is 1024KB.

A small typo in the format attribute for input files is corrected as well.


jj-umn (approved):
Looks great.

bgruening (approved):
Indeed very good work!

PR #736, opened 2023-11-29
Fix DIA-NN

npinter:
There is a problem when the user wants to use a custom library. DIA-NN does not ~~find~~recognize the speclib file if it is in the temp folder (with a .dat extension). Added a test to cover this issue.


bgruening (approved):

bgruening:
Merci!

PR #741, opened 2024-02-13
Update DIA-NN wrapper

npinter:
- Changed FASTA database input to optional (since search can be done with a predicted library w/o a FASTA db) + fix for having no FASTA db as input 
- Increased BioContainer version
- Removed all experimental features, to be in line with the licence (just commented out)


bgruening (approved):

bgruening:
Thanks!

PR #752, opened 2024-03-11
mzToSQLite biotools

bebatut:
Hi all,

This PR links the tools to its bio.tools entry: https://bio.tools/mztosqlite
It will be useful to get EDAM ontology annotations

Bérénice


bebatut:
Thanks!

PR #760, opened 2024-05-08
Remove raw file option from PepQuery2

reid-wagner:
Raw input files do not work with the current PepQuery galaxy implementation. The `-dc` flag must be used to pass a ThermoRawFileParser executable to pepquery, which will require further modifications to the pepquery recipe.

For the time being, remove this option.


bgruening (approved):

reid-wagner:
All fixed - thanks!

PR #759, opened 2024-05-07
Add FragPipe

reid-wagner:
This PR adds a wrapper for the FragPipe tool suite.

A handful of default FragPipe workflows are provided for the user, while more may be added in the future. Most FragPipe parameters are duplicated in the Galaxy tool, many of which are optional to allow the workflow defaults to be used.

Some components have not yet been implemented due to license restrictions (DIA-NN & MSBooster). Other parameters need further work and testing before implementation, and will default to the workflow defaults. The workflow, along with the log file and some other FragPipe outputs are optionally selected as Galaxy outputs.

License keys are hard-coded into the tool, and a required select option and text warnings are in place to ensure a user understands the licenses, as has been discussed with the developers.

PR #761, opened 2024-05-11
Add FragPipe

reid-wagner:
Re-creating #759

This PR adds a wrapper for the FragPipe tool suite.

A handful of default FragPipe workflows are provided for the user, while more may be added in the future. Most FragPipe parameters are duplicated in the Galaxy tool, many of which are optional to allow the workflow defaults to be used.

Some components have not yet been implemented due to license restrictions (DIA-NN & MSBooster). Other parameters need further work and testing before implementation, and will default to the workflow defaults. The workflow, along with the log file and some other FragPipe outputs are optionally selected as Galaxy outputs.

License keys are hard-coded into the tool, and a required select option and text warnings are in place to ensure a user understands the licenses, as has been discussed with the developers.


bgruening:
You can now see the planemo error reports at this page: https://github.com/galaxyproteomics/tools-galaxyp/actions/runs/9046215197?pr=761 scroll down.

Seems like the mv command has nothing to move. How do you run the tests locally?

reid-wagner:
Yes - it looks like `libz` is a dependency that might need to be added: `libz.so.1: cannot open shared object file`. It was already present on our system so I didn't catch it.

When I try to add it to the recipe with `<requirement type="package" version="1.2.13">zlib</requirement>`, the FragPipe package doesn't seem to be installed correctly (i.e. can't find `fragpipe` binary on path), so I haven't figured that out.

I'll add `zlib` to the bioconda recipe anyways, but I was hope to get it working here to verify that's all we need. I'm trying to run tests with `planemo test --engine docker_galaxy --no_cache_galaxy`, but I'm having separate issues there too. I am able to run this manually on our university Galaxy instance.

bgruening:
> When I try to add it to the recipe with `<requirement type="package" version="1.2.13">zlib</requirement>`, the FragPipe package doesn't seem to be installed correctly (i.e. can't find `fragpipe` binary on path), so I haven't figured that out.

In those cases its probably the wrong, incompatible zlib version. Try on your commandline `conda create -n foo --dry-run fragpipe=x.x.x zlib=y.y.y` you will see that the solver will bail out maybe and then you can play around with the version until it works.

But you are correct, fixing the conda recipe is the correct thing to do.
 
> I'll add `zlib` to the bioconda recipe anyways, but I was hope to get it working here to verify that's all we need. I'm trying to run tests with `planemo test --engine docker_galaxy --no_cache_galaxy`, but I'm having separate issues there too. I am able to run this manually on our university Galaxy instance.

Oh I see, this is wrong :)

Just do `planemo test --biocontainers` this will run your tests using a container (that is created if needed). And it is exactly what is running here on CI.

reid-wagner:
> you can play around with the version until it works

This sort-of how I came to zlib=1.2.13. I built the fragpipe recipe with an unpinned zlib dependency and used the resulting zlib version. Creating a new environment with zlib and fragpipe as separate dependencies also resolved to 1.2.13. I'm wondering if there's a way to find logs of the mulled environment creation?

> Oh I see, this is wrong :)

Okay, thanks for the advice! I'm just getting started with `planemo test` so I have a lot to learn.

reid-wagner:
@bgruening Is there a way to force builds to use a biocontainer with a more recent build of the fragpipe bioconda package? I fixed the zlib issue in the recipe, since I had trouble with it here, but I think it's still using the older version.

reid-wagner:
> @bgruening Is there a way to force builds to use a biocontainer with a more recent build of the fragpipe bioconda package? I fixed the zlib issue in the recipe, since I had trouble with it here, but I think it's still using the older version.

Apologies - I didn't sufficiently fix the dependency. And it is using the most recent biocontainer.

bgruening:
Are you saying the _3 container here: https://quay.io/repository/biocontainers/fragpipe?tab=tags&tag=latest is not working?

bgruening:
Are you saying the _3 container here: https://quay.io/repository/biocontainers/fragpipe?tab=tags&tag=latest is not working?

reid-wagner:
It's not, but I made an obvious mistake with the dependencies so I'm working to fix that in the recipe now.

reid-wagner:
This is ready for review now.

There actually was an issue with the biocontainer - it seems like `ldconfig` is not appropriately setting the library path in the container. I worked around this by adding `/usr/local/lib` to `LD_LIBRARY_PATH` in the recipe.

tools/fragpipe/fragpipe.xml
```diff
@@ -0,0 +1,286 @@
...
+                --contamprefix &&
+                mv *-decoys-*.fas decoys-contam-database.fas &&
+            #end if
+            fragpipe --msfragger_key f28c4be2-759b0a11-6c1ea652-1240565a --ionquant_key 47cb2953-b4bf7726-2c96f5fe-8f1013aa --headless --threads $threads --ram \$[ \${GALAXY_MEMORY_MB:-8192} / 1024 ] --workflow fp.workflow --manifest fp.manifest --workdir outputs
+            --config-msfragger \${CONDA_PREFIX}/share/msfragger-*/MSFragger-*/MSFragger-*.jar --config-philosopher `which philosopher` --config-python `which python` --config-ionquant \${CONDA_PREFIX}/share/ionquant-*/IonQuant.jar
+            && cd outputs
+            && ln -s log_*.txt log.txt
+            #if $output_options and 'combined_outputs' in $output_options
+                && mkdir combined_outputs
+                && for f in combined*tsv; do [[ -e $\{f} || continue ]]; mv \${f} combined_outputs; done
```
bernt-matthias, line 53: `-e $\{f}` looks wrong. Can you add a test covering this piece of code?

maybe a `find . -name "combined*tsv" -exec mv {} combined_outputs/\;`

tools/fragpipe/fragpipe.xml
```diff
@@ -0,0 +1,286 @@
...
+            #end if
+            && for f in psm.tsv ion.tsv peptide.tsv protein.tsv; do n=0; for i in `find * -name \${f}`; do d=\${i%/*}; echo "\${n}  \${d} \${i}"; if [[ \$n == 0 ]]; then awk 'NR==1{printf("Experiment\t%s\n",\$0);}' \$i > concat_\${f}; fi; awk 'NR>1{printf("'"\${d}"'\t%s\n",\$0);}' \$i >> concat_\${f}; t=`echo \${i} | tr '/' '.'`; mv \${i} \${t}; n=\$((n+1)); done; done
```
bernt-matthias, line 55: Can you describe what you are doing here, e.g. in a code comment? My preference would also be to indent the bash loop and if statements to make this readable.

tools/fragpipe/macros.xml
```diff
@@ -0,0 +1,1161 @@
...
+    <token name="@iTRAQ-8_LABELS@">['113', '114', '115', '116', '117', '118', '119', '121']</token>
+
+    <!--
+      User specifies input file format and selects sample files as are included by the scan_inputs macro
+
+      User must also supply their own manifest file, a TSV file of the format:
+      name, experiment, bioreplicate, data type
+    -->
+    <xml name="samples">
+        <conditional name="scan">
```
bernt-matthias, line 23: I guess you want to forbid that users mix input with different formats. 
Maybe use a single input and just document that the inputs need to be of the same format. 

Or just support mzml .. people can use converters. 

Reason: with a conditional this becomes hard to use in a workflow.

tools/fragpipe/fragpipe.xml
```diff
@@ -0,0 +1,286 @@
...
+        <param name="workflow_only" type="boolean" truevalue="yes" falsevalue="no" checked="false" label="Only generate the workflow output"/>
+
+        <param name="output_options" label="Additional outputs" type="select" multiple="true" display="checkboxes" optional="true">
+            <option value="workflow">FragPipe Workflow</option>
+            <option value="log">FragPipe Log</option>
+            <option value="combined_outputs">FragPipe Combined Outputs</option>
+            <option value="concatenated_outputs">TSV Outputs Concatenated</option>
+        </param>
+
+        <param argument="--threads" type="integer" value="8" label="Number of threads"/>
```
bernt-matthias, line 87: This should not be controllable by the user. Instead use `\${GALAXY_SLOTS:-8}` in the command.

bernt-matthias (commented):
Some initial thoughts .... maybe we can sit together these days and see if we can simplify the configfile generation or just go through step by step... But not sure when.

tools/fragpipe/genericize_db.py
```diff
@@ -0,0 +1,18 @@
...
+input_db_file = sys.argv[1]
+output_db_file = sys.argv[2]
+
+with open(input_db_file) as f:
+    input_db = f.readlines()
+
+subbed_lines = [re.sub(r'^>(?!sp\||tr\||db\||AP_|NP_|YP_|XP_|WP_|ENSP|UniRef|nxp|generic)', '>generic|', l) for l in input_db]
+
+with open(output_db_file, 'w') as f:
+    f.writelines(subbed_lines)
```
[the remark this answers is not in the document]
reid-wagner: Good point. Updated.

tools/fragpipe/fragpipe.xml
```diff
@@ -0,0 +1,286 @@
...
+        - Modifies database path in workflow file
+        - Runs Philosopher to add decoys and contaminants
+        - Runs FragPipe, passing paths to MSFragger, Philosopher, Python, and IonQuant
+    -->
+    <command detect_errors="exit_code"><![CDATA[
+        mkdir .cache .config tmp &&
+        export XDG_CONFIG_HOME=\$(pwd)/.config &&
+        export XDG_CACHE_HOME=\$(pwd)/.cache &&
+        export JAVA_OPTS="\$JAVA_OPTS -Xmx\$[ \${GALAXY_MEMORY_MB:-8192} / 1024 ]g -Djava.io.tmpdir=\$(pwd)/tmp" &&
+        export LD_LIBRARY_PATH="\$LD_LIBRARY_PATH:/usr/local/lib" &&
```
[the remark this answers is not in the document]
reid-wagner: I actually needed to add this for the Github tests. The biocontainer had dependencies installed in that location, in this case `libz.so`, but it was not on `LD_LIBRARY_PATH` (`ldconfig` not run?).

Is a non-existent path in that variable transparent to `ld`? I kind of expected it to be, and I just ran `planemo test --biocontainers` with a garbage path and it worked.

tools/fragpipe/macros.xml
reid-wagner: Sure. I just removed the conditional.

tools/fragpipe/macros.xml
```diff
@@ -0,0 +1,1161 @@
...
+          combined_pepxml="false"
+          protp_minprob="0.5"/>
+        <expand macro="isobaric_quantification" yes="true" no="false" plex="@PLEX@" level="2" ref_tag="Bridge" mod_tag="none"/>
+        <expand macro="label_free_quantification" yes="false" no="true"/>
+        <expand macro="ptm_shepherd" yes="false" no="true" token_unimod="false" token_common="true"/>
+    </xml>
+
+    <!-- Generates the FragPipe workflow-->
+    <xml name="workflow_configfile">
+        <configfile name="workflow_configfile"><![CDATA[#slurp
```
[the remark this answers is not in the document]
reid-wagner: I can put that together. I'll convert the FragPipe-provided workflows into macro files.

tools/fragpipe/fragpipe.xml
```diff
@@ -0,0 +1,286 @@
...
+        - Links input files
+        - Modifies database path in workflow file
+        - Runs Philosopher to add decoys and contaminants
+        - Runs FragPipe, passing paths to MSFragger, Philosopher, Python, and IonQuant
+    -->
+    <command detect_errors="exit_code"><![CDATA[
+        mkdir .cache .config tmp &&
+        export XDG_CONFIG_HOME=\$(pwd)/.config &&
+        export XDG_CACHE_HOME=\$(pwd)/.cache &&
+        export JAVA_OPTS="\$JAVA_OPTS -Xmx\$[ \${GALAXY_MEMORY_MB:-8192} / 1024 ]g -Djava.io.tmpdir=\$(pwd)/tmp" &&
```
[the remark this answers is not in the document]
reid-wagner: > Not sure if setting the -Xmx in the tool is a good idea.

Ah, yes. This should already be handled via via the `fragpipe --ram` flag. I should have removed this.

> Also overwriting JAVA_OPTS seems not not be a good idea.

This was the action suggested by the developer to specify a non-system tmp location: https://github.com/Nesvilab/FragPipe/issues/478. That said, it looks like there may be a bug in fragpipe such that it's not being correctly used.

Is there a specific reason not to supplement `JAVA_OPTS` in the tool xml?

tools/fragpipe/fragpipe.xml
reid-wagner: Basically this works to only `mv` the files if they exist but I agree it's too convoluted, I got tunnel-vision with the for-loop.

I think the line you suggested will still fail if there are no files, but something like `find . -name 'combined*tsv' | xargs -I {} mv {} combined_outputs` would work. Or just a bash conditional.

tools/fragpipe/fragpipe.xml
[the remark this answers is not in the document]
bernt-matthias: My mistake. Since you are not overwriting, but appending to the variable this  is fine.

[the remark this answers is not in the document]
reid-wagner: Cool. I removed the `-Xmx`.

tools/fragpipe/msfragger_macros.xml
```diff
@@ -0,0 +1,1009 @@
...
+#import re
+#def identifier_or_name($input1)
+    #if hasattr($input1, 'element_identifier')
+        #return $input1.element_identifier
+    #else
+        #return $input1.name
+    #end if
+#end def
+#def clean($name1)
+    #set $name_clean = $re.sub('[^\w\-_]', '_', $re.sub('(?i)[.](fa|fas|fasta|imzml|mzml|mzxml|raw)$','', $re.sub('.*/','', $name1.rstrip('.gz'))))
```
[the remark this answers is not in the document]
reid-wagner: Yes, we discard the path so we can create a symlink with the correct extension.

tools/fragpipe/macros.xml
```diff
@@ -0,0 +1,1161 @@
...
+#set $kv = $k + '=' + str($wfdict[$k])
+$kv
+#end for
+#slurp]]></configfile>
+    </xml>
+
+    <!-- Sets database name in workflow. Necessary because <command> section may generate new database file after configfile has been generated. -->
+    <token name="@UPDATE_WORKFLOW@"><![CDATA[
+DB=`pwd`/$db_name &&
+sed "s#^database.db-path.*#\database.db-path=\$DB#" $workflow_configfile > fp.workflow;
```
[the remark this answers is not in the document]
reid-wagner: Yes. I changed it to use a common symlink name whatever the path it takes for decoys and contaminants, and simply use this name in the configfile.

tools/fragpipe/fragpipe.xml
reid-wagner: I updated up the entire `<command>` tag to be more clear in this commit (https://github.com/galaxyproteomics/tools-galaxyp/pull/761/commits/5c2189e00f8aa1a451b567d51c16c4a252077f62), including indenting and commenting this. I broke `<command>` into further macros as well.

Copying the comment I added:
> For each category of results - psm, ion, peptide, protein - this loop concatenates results from each sample into a single result. For example, the entries sample1/psm.tsv and sample2/psm.tsv would be combined to concat_psm.tsv, with a single psm.tsv header as its first line.

tools/fragpipe/fragpipe.xml
```diff
@@ -0,0 +1,286 @@
+<tool id="fragpipe" name="FragPipe -  Academic Research and Education User License (Non-Commercial)" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" python_template_version="3.5">
```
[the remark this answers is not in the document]
reid-wagner: > Any chance to reduce the name to just FragPipe? Often tool devs add a boolean switch to check for academic use.

We do have that boolean as well, but unfortunately we can't change the name. This was specifically requested by the developer's legal team to make it more visible when re-running shared histories or workflows.

> Please ass a recent profile and remove python_template_version.

Having trouble running tests when I define profile, but working on it.

tools/fragpipe/fragpipe.xml
reid-wagner: I updated this to use the `xargs` command I quoted.

tools/fragpipe/fragpipe.xml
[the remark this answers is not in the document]
bernt-matthias: > we can't change the name

Maybe move to the description? As a compromise... or move the boolean to the very top.

tools/fragpipe/macros.xml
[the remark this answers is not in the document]
reid-wagner: Per discussion, we will leave this as-is due to the complexity of storing all workflow parameters as defaults within the tool definition. In the future we will explore ways we might achieve this.

tools/fragpipe/fragpipe.xml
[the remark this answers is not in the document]
reid-wagner: > Maybe move to the description

It was a pretty direct requirement from their team. We could go back to the drawing board on any of the licensing  in the future, but for the time being I think further deliberation might not be worth.

> Please add a recent profile

`20.05` was the most recent version that works with the biocontainer tests, so I added that!

tools/fragpipe/fragpipe.xml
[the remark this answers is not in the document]
reid-wagner: I also addressed the local cache issue I mentioned with this commit (https://github.com/galaxyproteomics/tools-galaxyp/pull/761/commits/3031e8f28e182197f68151537bee32f81616c61b).

You can see here (https://github.com/Nesvilab/FragPipe/blob/develop/MSFragger-GUI/src/com/dmtavt/fragpipe/FragpipeLocations.java#L91) that FragPipe always resolves its cache location relative to its jar location, so I copy the jar to the tool directory so that it can create the cache.

The alternative is that on a system where the user running the Galaxy job has read/write access to the conda installation, there could be multiple FragPipe processes using the same cache.

tools/fragpipe/fragpipe.xml
reid-wagner: Fixed to use GALAXY_SLOTS.

tools/fragpipe/macros.xml
```diff
@@ -0,0 +1,1206 @@
...
+    <!--
+      User specifies input file format and selects sample files as are included by the scan_inputs macro
+
+      User must also supply their own manifest file, a TSV file of the format:
+      name, experiment, bioreplicate, data type
+    -->
+    <xml name="samples">
+        <param name="inputs" type="data" format="mzML,mzXML" multiple="true" label="Proteomics Spectrum files"  help="All input scan files must of a matching format: mzML, mzXML"/>
+        <param name="input_prefix" type="text" value="" optional="true" label="File name prefix" help="Names inputs: prefix_rep#.mzXML Leave blank to use History names of inputs">
+              <validator type="regex" message="">[a-zA-Z][a-zA-Z0-9_-]*</validator>
```
bernt-matthias, line 25: Interesting that the validator does not give an error when the value is `""`. But makes sense for optional parameter.

tools/fragpipe/macros.xml
```diff
@@ -0,0 +1,1206 @@
...
+    ]]></token>
+
+    <!--
+      Prepares outputs.
+
+      Creates log file symlink, makes a new directory for "combined" outputs and creates "concatenated" results.
+    -->
+    <token name="@PREPARE_OUTPUTS@"><![CDATA[
+        cd outputs &&
+        ln -s log_*.txt log.txt &&
```
bernt-matthias, line 112: There is only one `log_*.txt` file?

tools/fragpipe/fragpipe.xml
```diff
@@ -0,0 +1,276 @@
...
+            <filter>(not workflow_only) and ('TMT' in wf['workflow_name'])</filter>
+        </collection>
+    </outputs>
+
+    <tests>
+      <test expect_num_outputs="11">
+        <param name="inputs" value="default/test1.mzML,default/test2.mzML" ftype="mzml"/>
+        <param name="database_name" value="default/test.fasta" ftype="fasta"/>
+        <param name="manifest" value="default/test.manifest" ftype="tabular"/>
+        <param name="workflow_name" value="Default"/>
```
bernt-matthias, line 156: Any chance to add a test for other workflow(s)? Or better later?

tools/fragpipe/macros.xml
```diff
@@ -0,0 +1,1206 @@
...
+            <conditional name="isobaric_quantification">
+                <param name="isobaric_quantification_step" type="select" label="Perform Isobaric Quantification">
+                    <option value="default">Use workflow defaults</option>
+                    <option value="yes" selected="@YES@">yes</option>
+                    <option value="no" selected="@NO@">no</option>
+                </param>
+                <when value="default"/>
+                <when value="no"/>
+                <when value="yes">
+                    <section name="options" expanded="true" title="Isobaric Quantification">
```
bernt-matthias, line 952: Does having a single section make sense here?

tools/fragpipe/macros.xml
```diff
@@ -0,0 +1,1206 @@
...
+                    #set $wfdict['tmtintegrator.mod_tag'] = $cxt.glycan_conditional.mod_tag
+                #end if
+                #if $cxt.glycan_conditional.min_site_prob != ''
+                    #set $wfdict['tmtintegrator.min_site_prob'] = $cxt.glycan_conditional.min_site_prob
+                #end if
+                #if $cxt.glycan_conditional.glyco_qval != ''
+                    #set $wfdict['tmtintegrator.glyco_qval'] = $cxt.glycan_conditional.glyco_qval
+                #end if
+            #end if
+            #if $cxt.unique_gene != 'None'
```
bernt-matthias, line 1098: I'm pretty sure that unset optional selects evaluate to `None`. Needs to be checked  everywhere.

tools/fragpipe/macros.xml
```diff
@@ -0,0 +1,1206 @@
...
+                #set $wfdict['tmtintegrator.unique_gene'] = $cxt.unique_gene
+            #end if
+            #if $cxt.unique_pep != 'None'
+                #set $wfdict['tmtintegrator.unique_pep'] = $cxt.unique_pep
+            #end if
+            #if $cxt.min_pep_prob != ''
```
bernt-matthias, line 1104: Not 100% sure, but I think you should test for `str($cxt.min_pep_prob) != ''` . Needs to be checked everywhere.

tools/fragpipe/macros.xml
```diff
@@ -0,0 +1,1206 @@
...
+    <token name="@PREPARE_ENVIRONMENT@"><![CDATA[
+        mkdir -p fragpipe_local/lib &&
+        for dir in bin philosopher tools workflows; do
+            ln -s \${CONDA_PREFIX}/share/fragpipe*/\${dir} fragpipe_local/\${dir};
+        done &&
+        for file in \${CONDA_PREFIX}/share/fragpipe*/lib/*; do
+            ln -s \${file} fragpipe_local/lib/\$(basename \${file});
+        done &&
+        cp \${CONDA_PREFIX}/share/fragpipe*/lib/fragpipe-*.jar fragpipe_local/lib/ &&
+        mkdir .cache .config tmp &&
```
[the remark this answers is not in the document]
reid-wagner: Yes, I am going to bring this up with the developers. It appears to be intentional so maybe there is a technical limitation that led to this choice.

I can definitely just copy everything - looks like its 250M. Updated.

tools/fragpipe/macros.xml
```diff
@@ -0,0 +1,1206 @@
...
+    <!--
+      Prepares input scan files.
+
+      Scans and moved to a scan directory and linked.
+
+      Annotation files are copied into the scan directory for TMT workflows.
+    -->
+    <token name="@LINK_SAMPLES@"><![CDATA[
+        #import os
+        #set CWD = os.getcwd()
```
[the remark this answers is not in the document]
reid-wagner: Good catch. Removed. Thanks!

tools/fragpipe/macros.xml
```diff
@@ -0,0 +1,1206 @@
...
+        CWD=`pwd | sed 's/\//\\\\\//g'` &&
```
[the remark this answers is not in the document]
reid-wagner: That looks much nicer. Updated.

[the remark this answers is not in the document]
reid-wagner: Interestingly, the tests fail with `sed: bad option in substitution expression`, despite this working on the command-line in the biocontainer. I've reverted temporarily but will come back to it.

tools/fragpipe/macros.xml
reid-wagner: I suppose the assumption is that the author will ensure the default value is acceptable. Sounds like this is fine as is.

tools/fragpipe/macros.xml
```diff
@@ -0,0 +1,1206 @@
...
+
+        mkdir -p scans &&
+        #if $wf.workflow_name == "TMT10" or $wf.workflow_name == "TMT11"
+            cp -p '$wf.annotation' 'scans/annotation.txt' &&
+        #end if
+
+        sed "s/^/\$CWD\/scans\//" $manifest > fp.manifest &&
```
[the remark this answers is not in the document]
reid-wagner: From my testing I think these files might be referenced from different directories within FragPipe execution. E.g. with the suggested change: 
```
Could not find MS file at /private/var/folders/hv/l_4hpsb56pj8nrc4pqczx5yc0000gn/T/tmp97qulb9c/job_working_directory/000/5/working/outputs/scans/test1.mzML
```

So the working directory is `working/outputs/`. But if I try to create a relative path `../scans`:

```
2024-07-05 22:07:10,669 ERROR - Some loaded files don't exist. Please double check.
2024-07-05 22:07:10,670 ERROR - /private/var/folders/hv/l_4hpsb56pj8nrc4pqczx5yc0000gn/T/tmpxk18xxrx/job_working_directory/000/5/working/../scans/test1.mzML
```

I could create a symlink, or keep the absolute path.

tools/fragpipe/macros.xml
reid-wagner: Yes, but it is dated. I've moved the wildcard to the `from_work_dir` attribute to be more concise and removed this link.

tools/fragpipe/macros.xml
reid-wagner: I actually put a bit of time into getting this and the below comment (`str($cxt.min_pep_prob) != ''`) right. Here's what my tests shows. I couldn't get builtin `type()` working in the cheetah so I'm just printing the evaluation:

```
#set $unique_gene_test_str_None = $cxt.unique_gene == 'None'
#set $unique_gene_test_bool_None = $cxt.unique_gene == None
#set $min_pep_prob_test_str = $cxt.min_pep_prob == ''
unique_gene_test_str_None $unique_gene_test_str_None
unique_gene_test_bool_None $unique_gene_test_bool_None
min_pep_prob_test_str $min_pep_prob_test_str
```

Result:
```
unique_gene_test_str_None True
unique_gene_test_bool_None False
min_pep_prob_test_str True
```

bernt-matthias (approved):
Looks very good to me. Any comments @bgruening ?

tools/fragpipe/fragpipe.xml
reid-wagner: I can definitely follow up with more tests. I picked these two because they had good coverage of the remaining workflows.

reid-wagner:
@bernt-matthias Thanks for the review!

tools/fragpipe/fragpipe.xml
```diff
@@ -0,0 +1,276 @@
...
+                --ionquant_key 47cb2953-b4bf7726-2c96f5fe-8f1013aa
+                --headless
+                --threads \${GALAXY_SLOTS:-8}
+                --ram \$[ \${GALAXY_MEMORY_MB:-8192} / 1024 ]
+                --workflow fp.workflow --manifest fp.manifest
+                --workdir outputs
+                --config-msfragger \${CONDA_PREFIX}/share/msfragger-*/MSFragger-*/MSFragger-*.jar
+                --config-philosopher `which philosopher`
+                --config-python `which python`
+                --config-ionquant \${CONDA_PREFIX}/share/ionquant-*/IonQuant.jar &&
```
bgruening, line 47: I'm surprised that CONDA_PREFIX works in a container setup. Does `which philosopher/../share/ionquant` also works and might be more stable?

tools/fragpipe/macros.xml
```diff
@@ -1074,7 +1075,7 @@ $kv
             #if $cxt.prot_norm != 'None'
                 #set $wfdict['tmtintegrator.prot_norm'] = $cxt.prot_norm
             #end if
-            #if cxt.glycan_conditional.use_glycan_composition == 'yes'
+            #if $cxt.glycan_conditional.use_glycan_composition == 'yes'
```
bgruening, line 1078: Good catch, is there a test missing that could spot this error before?

reid-wagner: Yes, this line is now covered by the tests - the bigger issue was that the parameter wasn't being evaluated: https://github.com/galaxyproteomics/tools-galaxyp/pull/761/commits/fd36cd446a61315ecedc75a9aea4a64b8ac6ea84#diff-11d9fe8e0e5baad2e54c5f8ca8c0e2c01a1b2d875f21f09ab5d8dc1d35aa7157R327.

I could in the tests set arbitrary non-default parameters from each token macro and check that the change is present in the workflow.

tools/fragpipe/fragpipe.xml
bgruening: ping ... The conda prefix is what makes me nervous in container envs

reid-wagner: Sorry for the delay - I just pushed a commit that removes all CONDA_PREFIX references.

tools/fragpipe/macros.xml
reid-wagner: A little confused on this point. Would it be to remove the "no" and "default" options? The "yes"/"no" distinction allows us to set the param that enables this in FragPipe. The "default" is usually a bit redundant but the argument could be that if a user is re-running a shared job, they might not know the workflow default and could use that option to revert to it.

tools/fragpipe/macros.xml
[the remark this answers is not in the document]
reid-wagner: Okay, this is working now.

tools/fragpipe/macros.xml
bernt-matthias: default / no / yes is fine for me. was just wondering if a single section inside a conditional is helpful. 

but fine with me.

bgruening (approved):

bgruening:
Awesome! Thanks @reid-wagner and @bernt-matthias

reid-wagner:
Thank you @bgruening and @bernt-matthias!

PR #764, opened 2024-06-29
Adding a tool to create Zeiss laser-microdissection annotations from tabular

foellmelanie:
Laser-microdissection extracts tissue areas of interest from thin tissue sections. These specific tissue areas can be prepared in the laboratory to extract DNA, RNA, proteins etc. and measure them for example with omics methods. 
Traditionally the annotation of tissue areas of interest are done manually directly in the software of  the laser-microdissection system. Here, we provide a tool that increases throughput and allows compatibility with machine learning based imaging software such as Qupath and HaloAI. This tool uses a tabular file with x and y coordinates and transforms it into a specifically formatted annotation text file that can be imported into Zeiss laser-microdissection systems.


foellmelanie:
New PR here: https://github.com/galaxyproteomics/tools-galaxyp/pull/765

PR #770, opened 2024-07-24
Adds option to run FragPipe job with higher memory

reid-wagner:
The Galaxy administrator may use this variable to run the job with increased resources.

### DIFF
--- a/tools/fragpipe/fragpipe.xml
+++ b/tools/fragpipe/fragpipe.xml
@@ -71,6 +71,9 @@
         <!-- Only generate a workflow, don't run FragPipe -->
         <param name="workflow_only" type="boolean" truevalue="yes" falsevalue="no" checked="false" label="Only generate the workflow output"/>
 
+        <!-- Galaxy admins may utilize this variable to for jobs requiring higher memory  -->
+        <param name="high_memory" type="boolean" checked="false" label="High memory job" help="This option is made available for Galaxy administrators to allocate more resources to FragPipe when selected. It may not be used by all systems - please contact your adminstrator to learn if your Galaxy instance does."/>
+
         <param name="output_options" label="Additional outputs" type="select" multiple="true" display="checkboxes" optional="true">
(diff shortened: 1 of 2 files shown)


tools/fragpipe/fragpipe.xml
```diff
@@ -71,6 +71,9 @@
         <!-- Only generate a workflow, don't run FragPipe -->
         <param name="workflow_only" type="boolean" truevalue="yes" falsevalue="no" checked="false" label="Only generate the workflow output"/>
 
+        <!-- Galaxy admins may utilize this variable to for jobs requiring higher memory  -->
+        <param name="high_memory" type="boolean" checked="false" label="High memory job" help="This option is made available for Galaxy administrators to allocate more resources to FragPipe when selected. It may not be used by all systems - please contact your adminstrator to learn if your Galaxy instance does."/>
```
bgruening, line 75: Sorry I missed some discussion I guess. How should a user know that? What is the use-case here?
Is the TPV rule not working?

reid-wagner: The TPV rule checks for this parameter. So this parameter will not be used by all installations, which I noted to the user in the help section.

reid-wagner: I'm not sure there's an exact precedent for this. I see that hifiasm also passes the parameter it uses for TPV to the program's command-line args, whereas this parameter is only used by TPV. But I thought this was a straightforward way to allow a user to request more memory, if the request can be fulfilled.

bgruening:
I guess this can be cloned?

reid-wagner:
> I guess this can be cloned?

Yes, thanks for the reminder. To document the discussion here - we are leaving the memory configuration choices to the admins alone, who can utilize input file sizes and other tool-related config. In the meantime, on usegalaxy.eu we simply increased the memory for all FragPipe jobs and it seems to be working well.

PR #772, opened 2024-07-30
Fix for QuPath ROI Splitter (v0.3.1)

npinter:
This version fixes the PNG output, changed the datatype of the coordinates to tabular and added a fix for a special case (LMD data).


bgruening (approved):

bgruening:
Merci!

PR #774, opened 2024-09-12
Fix IonQuant Options in FragPipe

reid-wagner:
Two options in IonQuant were incorrectly defined, causing the tool to fail immediately. I've fixed the automated test as well to cover this section.


bgruening (approved):

bgruening:
Thanks!

reid-wagner:
Thank you!

PR #779, opened 2024-11-06
add bio.tools id to xml

georgehe23:
Added uniprotxml_downloader bio.tools id


bernt-matthias:
I would prefer just linking https://bio.tools/uniprot

PR #781, opened 2024-11-22
Set default to "No" for conditional parameters in msconvert

reid-wagner:
Between Galaxy versions .2 and .3 (https://github.com/galaxyproteomics/tools-galaxyp/pull/748) the default setting for these conditional parameters changed from false to true when the parameter type was changed from `boolean` to `select`, because of the way these two types are interpreted.

I believe we want these to be off by default, which this PR implements. @bebatut - please let me know if this is incorrect. Thanks!


reid-wagner:
Seems to be failing due to disk space. Is there an existing solution to fix CI here?

bgruening:
@reid-wagner just want to make you aware that the errors might be in the lining.

planemo is way more stringent now. Especially, the syntax for tests is way more rigorous.  Please ping me if you need any help.

reid-wagner:
@bgruening - Yes, I noticed the linting has begun to fail due to that change.

First I'm attempting to address the tool tests, which fail due to limited runner space. I've pushed a number of commits just making sense for myself of where the job is running out of space, and where we can reduce storage.

That said, I think I've got a solution between reducing the number of tests (with the same coverage of features tested), and potentially just removing more items from the filesystem in  `pr.yaml` as we already do with `/usr/share/dotnet`. I'll address the new linting items, and then this should be ready soon! Thanks.

tools/msconvert/msconvert_macros.xml
```diff
@@ -551,7 +551,7 @@
   <xml name="msconvertOutput">
     <outputs>
       <data format="mzml" name="output" label="${($input.name[:-4] if $input.name.endswith('.tar') else $input.name).rsplit('.',1)[0]}.${output_type}" >
-        <filter>general_options['multi_run_output']['do_multi_run_output'] == False</filter>
+        <filter>general_options['multi_run_output']['do_multi_run_output'] == "false"</filter>
```
bgruening, line 554: This looks wrong if its a boolean.

tools/msconvert/msconvert.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="msconvert" name="msconvert" version="@VERSION@.4">
+<tool id="msconvert" name="msconvert" version="@VERSION@.5">
```
bgruening, line 1: please add a profile version

tools/msconvert/msconvert_macros.xml
reid-wagner: This was changed to a select option in version 3.0.20287.3 (https://github.com/galaxyproteomics/tools-galaxyp/pull/748).

tools/msconvert/msconvert.xml
reid-wagner: Added!

reid-wagner:
Thanks for the review @bgruening. Here are the changes I've made to get tests passing, and a few others:

- Removed a number of unnecessary files from the runner to free up space.
- Removed unused test data.
- Removed a number of individual tests, adding their filters to two composite tests. I've checked that if any of these filters fail, the test output will be invalid.
- Removed a couple of tests I felt could be considered redundant. Specifically, the SciexTest tests with `runIndexSet=0`, as this parameter as well as `indices` and `srmAsSpectra` were already accounted for in another test.
- Updated test parameter name attributes to pass current linting standards.
- Added profile=24.2

This PR is ready to be reviewed and merged from my end.

bgruening (approved):

bgruening:
Cool, thanks a lot!

bernt-matthias:
@reid-wagner was there a specific reason for using `profile=24.2`? Or would also a slightly older profile work? 
Galaxy's which do not have 24.2 can not use the tool at the moment.

reid-wagner:
@bernt-matthias - I did consider this, but I thought it would be sufficient to use the latest profile. Do you have a recommended method for ascertaining the oldest compatible profile version?

I don't have a specific reason for 24.2, so if we can lower the version to allow more admins to install it then that would be great.

bernt-matthias:
> @bernt-matthias - I did consider this, but I thought it would be sufficient to use the latest profile. Do you have a recommended method for ascertaining the oldest compatible profile version?
> 
> I don't have a specific reason for 24.2, so if we can lower the version to allow more admins to install it then that would be great.

Unless the tool does not use a new datatype or new tool feature, or one of the changes listed here https://docs.galaxyproject.org/en/master/dev/schema.html#tool there is no real need to a higher profile. 

My rule of thumb is to use a profile from 1 year back, e.g. 23.2

bernt-matthias:
https://github.com/galaxyproteomics/tools-galaxyp/pull/790

reid-wagner:
Got it - thanks. That's a good rule of thumb to know.

PR #782, opened 2024-11-24
Sync CI with IUC

bernt-matthias:


.github/workflows/ci.yaml
```diff
@@ -7,7 +7,7 @@ on:
     types: [run-all-tool-tests-command]
 env:
   GALAXY_FORK: galaxyproject
-  GALAXY_BRANCH: release_24.0
+  GALAXY_BRANCH: release_24.1
```
bgruening, line 10: Should we test against 24.2 now?

bgruening (approved):

.github/workflows/ci.yaml
bernt-matthias: I would say next time :) We do not need to be faster than IUC.

PR #783, opened 2024-11-24
OpenMS 3.2

bernt-matthias:


tools/openms/MSGFPlusAdapter.xml
bernt-matthias, line 1: MSGFPlus -tasks should be hardcoded

PR #787, opened 2025-01-22
FragPipe: Fix Digestion Options and Remove IonQuant

reid-wagner:
1. The only available version of IonQuant (due to licenses limitations) was found to be incompatible with this FragPipe version. An update to the wrapper for FragPipe 22.0 is in the works, which will allow use of IonQuant.
2. The digestion macro was found to be missing when setting the workflow values with user settings.


bernt-matthias (approved):
Tests run .. I'm happy :)

bgruening:
Thanks @reid-wagner

reid-wagner:
Thank you, both.

PR #788, opened 2025-03-20
Fix FragPipe Database Contaminants and Small Improvements

reid-wagner:
This PR primarily addresses a database contaminant issue, with a few smaller improvements and fixes:

1. An incorrect comparison meant contaminants were not added to database as expected. This is one of the pieces of the tool not covered by the workflow settings. This has been fixed, and an additional test added to verify contams.
2. MSBooster is enabled by default in multiple workflows, now that it is available in the tool suite.
3. A new test for the Nonspecific-HLA workflow has been added. This includes the test for contaminants in the database.
4. The workflow dictionary value for `run-peptide-prophet` is now correctly false, in the case of running msbooster and/or percolator. Most of the time this would have already been false by default.
5. The default value for `digest_mass_range_min` in the Nonspecific-HLA workflow has been updated to match the GUI default.
6. IonQuant parameters removed from tests, as IonQuant won't be available until the FragPipe 22.0 release.
7. Test parameters updated for latest linting standards.


bernt-matthias (approved):

bernt-matthias:
Good to go?

reid-wagner:
Great, thanks @bernt-matthias!

Issue #753, opened 2024-03-30
maxquant_mqpar attempts to read input file in cheetah templating


mvdbeek:
https://github.com/galaxyproteomics/tools-galaxyp/blob/2cce57c0b3173029ce8fb3e7b4607b3913581dfd/tools/maxquant/macros.xml#L155

causes the entire input file to be loaded into memory inside the job handler process. This patterns also doesn't work when using pulsar or where the head node can't read dataset contents. This breaks execution on usegalaxy.org currently.

bernt-matthias:
should be fixed in https://github.com/bernt-matthias/tools-galaxyp/commit/3fd0f018d423c779416a859b6b08b9005357a124

mvdbeek:
Thank you!

Issue #705, opened 2023-03-20
Should maxquant tools fail when PXTQC fails?


eschen42:
@npinter Whatcha think?

Both `maxquant` and `maxquant_mqpar` fail if PTXQC fails.  I wonder whether this is appropriate.

In both cases, the command (optionally) ends with running the PTXQC report
```
    #if $qc.do_it:
        &&
        Rscript '$qr' '$qr_yaml'
        #if 'log' in $output:
            >> '$log' 2>&1
        #end if
        &&
        cp ./combined/txt/report_v@VERSION_PTXQC@_combined.pdf '$ptxqc_report'
    #end if
```
The trouble is that, even when MaxQuant completes successfully, if PTXQC fails then the MaxQuant results (obtained at great expense of time and computational cost) are not copied back from the cluster.

I would prefer to see something like this (writing freely here, not presenting something tested):
```
    #if $qc.do_it:
        && (
         (
          Rscript '$qr' '$qr_yaml'
          #if 'log' in $output:
              >> '$log' 2>&1
          #end if
          &&
          cp ./combined/txt/report_v@VERSION_PTXQC@_combined.pdf '$ptxqc_report'
         ) || (echo 'PTXQC report failed: see log')
       )
    #end if
```

This would allow the user (e.g., me) to salvage PTXQC failures by running PTXQC (which takes a few minute or less rather than several days) separately, outside of Galaxy.

bernt-matthias:
My suggestion would be to put the R script in a separate tool.

PR #792, opened 2025-04-09
Maxquant fixes for modifications and enzymes parameters

bernt-matthias:
- many modifications contain `>` hence we must allow it as valid character.  otherwise it will be replaced by `__gt__` which leads to an error in MQ.
- fixed and valiable modifications are optional
- same for enzymes

Fixes

- [x] https://github.com/galaxyproteomics/tools-galaxyp/issues/753 (fixed it first, but was not really necessary anyway, since ptxqc should read it from the mqpar file anyway)
- [x] https://github.com/galaxyproteomics/tools-galaxyp/issues/705 (by making ptxqc a separate tool)


bgruening (approved):
That is a large change for a minor version bump :)

bernt-matthias:
> That is a large change for a minor version bump :)

Having a major update would be a larger project ..

PR #794, opened 2025-04-15
Specify equate_il=false instead of assuming default value from API

reid-wagner:
I found an issue documented here (https://github.com/unipept/unipept-api/issues/69) where the API seems to default to `equate_il=true`, whereas documentation states the default is false: https://unipept.ugent.be/apidocs/pept2lca

The Galaxy tool `unipept.py` doesn't specify the `equate_il` parameter if a user selects false, resulting in incorrect results from the API. In this PR the script is just explicit about this value so that results are as expected.

As an aside, I'm not sure how to handle versioning for this tool. I'm not sure if previously the API version was synced with web, but now it's a different system, the latest being 2.4.0. I can't think of a version to use here that feels correct, not to mention that as the API is updated this version string will become outdated. Are there other tools that handle this scenario?

Maybe I'm just confused about where the 4.5.1 version comes from, but input on this would be appreciated! Thanks


reid-wagner:
@pverscha - Thanks! That makes sense.

So I guess then that the Galaxy tool always used the web app version (`<tool id="unipept" name="Unipept" version="4.5.1" profile="22.05">`), despite technically calling the API, presumably because it's the same Unipept behind each interface generating results. I suppose we can continue with this scheme, meaning we just update it to the latest web version when we make a change?

This still has the pitfall that as Unipept has new releases, the Galaxy tool version will become outdated, despite actually using the latest Unipept. I'll see if I can find any examples of existing tools that handle this.

reid-wagner:
This is ready to be reviewed and merged. I added some changes to get tests working with the latest linting and Unipept results.

bernt-matthias (approved):
Out of interest: How do you define the version number, i.e. 6.2.4

reid-wagner:
That's the latest web version, which is what I think was the existing versioning scheme used.

PR #799, opened 2025-06-14
Add NetMHCPan and NetMHCiiPan Tools

reid-wagner:
Adds wrappers for these two tools via the latest IEDB API.

While there is already the `tools-iuc/iedb_api` tool, IEDB has released a new "Next-Generation" API (https://nextgen-tools.iedb.org/) with differences from the original. These tools are built to provide individual wrappers for the `mhci` and `mhcii` functions of the API, with a script adapted from the original to work with the new API. Additionally, a new mhcii-specific feature is added in the Python script to mimic the web version's inversion feature, which reverses input peptides and includes these in the analysis.

In the future, this could be broadened to include more available IEDB tools and replace the current `iedb_api` tool.


tools/iedb_netmhcpan/shared_macros.xml
```diff
@@ -0,0 +1,70 @@
+<macros>
+    <xml name="requirements">
+        <requirement type="package" version="3.7">python</requirement>
```
bgruening, line 3: I would use a more recent Python version, that one is already EOL

tools/iedb_netmhcpan/netmhcpan.xml
```diff
@@ -0,0 +1,88 @@
...
+    </macros>
+    <requirements>
+        <expand macro="requirements" />
+    </requirements>
+
+    <command detect_errors="exit_code"><![CDATA[
+        #import re
+        python '${__tool_directory__}/nextgen_iedb_api.py'
+        --tool-group=mhci
+        #for $method in str($methods).strip().split(","):
```
bgruening, line 15: same here?

bgruening (commented):
Just 2 tiny comments.

tools/iedb_netmhcpan/netmhciipan.xml
```diff
@@ -0,0 +1,104 @@
...
+    <requirements>
+        <expand macro="requirements" />
+    </requirements>
+
+    <command detect_errors="exit_code"><![CDATA[
+        #import re
+
+        python '${__tool_directory__}/nextgen_iedb_api.py'
+        --tool-group=mhcii
+        #for $method in str($methods).strip().split(","):
```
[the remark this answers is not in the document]
reid-wagner: Yep, that should. I've updated it.

PR #800, opened 2025-07-07
Update msconvert

chambm:
Update Docker container version
Add mzMLb output format

mzMLb is HDF5 based like mz5 but supports the ion mobility 3 array format (and any arbitrary extra arrays).


bgruening:
Add such a line: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/sample/datatypes_conf.xml.sample#L231

You are allowed to add this the last stable release, aka release_25.0

chambm:
> You are allowed to add this the last stable release, aka release_25.0

I don't really understand this sentence. But I'll make a PR to add the datatype. Does that mean this PR can't proceed until that PR is merged?

mbodova:
Hi, we would like to merge this branch since the msconvert on galaxy currently does not process some newer MS accessions. However the tests fail on launching wine64. Can I ask how galaxy uses it so we can maybe locate the bug?

mbodova:
Thank you, it seems that wine is no longer the problem, only failing tests. Are you working on it or would you like someone to have a look at it?:)

chambm:
Looking at those failures they are pretty varied. The corrupt RAW thing mirrors another report I got from another user outside Galaxy so I will be looking into that (will probably require an image version update). I don't think I have time to tackle the other at the moment though so if you've got bandwidth, I appreciate it!

hechth:
@chambm so it seems we have everything fixed except for the raw file thing which you successfully identified as being related to symlinks. We also tested the Galaxy tool wrapper with replacing all symlinks with copying, then it works.

Now my question is - when will your fix (https://github.com/ProteoWizard/pwiz/commit/dc73e9b017419978488a0fdac2c6023aacf5ac74) be deployed and when will there be a new dockerfile with this version of msconvert?

In case you currently don't have the bandwidth for this, do you think it would be feasible to replace the 'ln -s' with 'cp' for the time being, @bgruening ?

bgruening:
Yes, cp is ok.

chambm:
I thought I already fixed the Thermo thing. I updated to the latest Docker image which in my testing and someone else's testing works with symbolic links. It's a shame to have to switch to copying as that could bite quite a bit with large RAW files, but it's better than not working at all. We can revisit when I've finished porting pwiz to .NET so we can drop Wine.

hechth:
> I thought I already fixed the Thermo thing. I updated to the latest Docker image which in my testing and someone else's testing works with symbolic links. It's a shame to have to switch to copying as that could bite quite a bit with large RAW files, but it's better than not working at all. We can revisit when I've finished porting pwiz to .NET so we can drop Wine.

That is interesting - no idea. It still doesn't work in the wrapper and it works when using cp, as you can see. I can try also testing the desktop version with a symlink to see if that causes issues.

PR #820, opened 2026-04-27
Update siriusAdapter credentials

PaulineSGN:
Hello! I updated the way credentials are provided for the SiriusAdapter tool and added a unit test for it (API connection not included).

While working on this, I noticed the API doesn't behave as expected — this might be a Bright Giant server issue, or the tool may need to be updated for a newer version of OpenMS. I didn't have the time to investigate further, so feel free to contribute and dig deeper into this if you'd like!

I'm not a domain expert, so don't hesitate to suggest corrections or improvements to what I've done — all feedback is welcome. Thanks a lot!

ping : @TuturBaba @yvanlebras @melpetera


tools/openms/SiriusAdapter.xml
```diff
@@ -197,7 +191,16 @@ python3 '$__tool_directory__/fill_ctd.py' '@EXECUTABLE@.ctd' '$args_json' '$hard
       <filter>OPTIONAL_OUTPUTS is not None and "ctd_out_FLAG" in OPTIONAL_OUTPUTS</filter>
     </data>
   </outputs>
-  <tests/>
+  <tests>
```
bernt-matthias, line 194: What is the new test good for?

bernt-matthias (commented):
Hi @PaulineSGN . Thanks for the effort. The problem with the openms wrappers is that they are autogenerated. That is all your changes (except those for the macros file would be overwritten). You would need to blacklist the parameters in https://github.com/galaxyproteomics/tools-galaxyp/blob/master/tools/openms/aux/hardcoded_params.json and the regenerate with 
https://github.com/galaxyproteomics/tools-galaxyp/blob/master/tools/openms/aux/generate.sh .. let me know if you need help with the setup.

Then the current implementation would add the credentials to all openms tools ,but we should add only to Sirius adapter

tools/openms/SiriusAdapter.xml
PaulineSGN: In the previous version, there were no tests, so I thought it would be good if there was one to comply with Galaxy development best practices, even though it doesn’t test the part with credentials.

bernt-matthias: Certainly a good idea. Manual tests need to go into https://github.com/galaxyproteomics/tools-galaxyp/blob/master/tools/openms/aux/macros_test.xml, otherwise they will be overwritten when generating a new version.

PR #823, opened 2026-05-07
Update metaQuantome tools to version 2.0.4 

reid-wagner:
This includes a new option for the expand module allowing it to include obsolete go terms from the input database instead of failing the process.

In addition, parameter definitions were re-arranged to remove the unused `--slim_down` parameter from the filter and stat modules where it wasn't actually used in the command block.

Finally there are a few updates to pass linting:
- Profile added
- Test definitions updated
- Unnecessary "name" attribute removed from unique_value filter.
- Remove `value=""` default from integer parameter.


bgruening:
Ready for review?

bgruening (approved):

reid-wagner:
@bgruening Yes - thank you!

PR #835, opened 2026-09-08
Detect msfragger OOM

bernt-matthias:
Ideally fragpipe would not return an exit code of 0 in this case .. otherwise Galaxy would be able to detect the OOM without this change (but only for some runners I guess). 

@reid-wagner  you may decide if this should go in the next release


bgruening (approved):

reid-wagner (approved):
Sounds reasonable to release now. Thank you!

PR #838, opened 2026-09-16
Proteomiqon tool updates

paulineHans:
I have updated all existing ProteomIQon tools. To do this, I added more metadata (`<creator>`, `<citation>`), updated the tool versions and brought the tools into line with the IUC standard.


bgruening:
grafik

Those are still failing.

bgruening:
grafik

one tool is still resisting ...

paulineHans:
> grafik
> one tool is still resisting ...

That is confusing, because yesterday i fixed the errors and run planemo test locally and everything worked, but i will investigate this

PR #839, opened 2026-09-16
Fixes various tool tests

guerler:
Fixes failing tool tests across seven repositories: bumbershoot, cardinal, maxquant, moFF, morpheus, msstats and pyprophet.

Most were tests that had drifted from the tools they exercise: parameters naming inputs the tool no longer declares, conditional branches selected by the wrong value, and expected outputs that no longer matched what the corrected parameters produce. Where a test parameter sat outside the conditional that owns it, it is nested, and the profile is raised where the tool relies on validation that only newer profiles apply. `pyprophet`'s merged test output is regenerated, since the corrected template parameter changes it.

Also replaces documentation links whose hosts no longer resolve.

Version suffixes are bumped for the repositories whose tools changed. Two ProteomIQon tools are left at the versions master already carries: the only change here is a test parameter rename, which master had independently made to the declaration, so there is nothing to bump for.


tools/bumbershoot/idpqonvert.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<tool id="idpqonvert" name="idpQonvert" version="@VERSION@.0">
+<tool id="idpqonvert" name="idpQonvert" version="@VERSION@.1">
```
bgruening, line 2: @guerler can you please add profiles to all tools when we touch them? And maybe also make sure its a recent profile version?

